Issue #14, opened 2024-09-30
Create structure for entire AIR


mitschabaude:
Including vanishing polynomials

Maybe various types of specifying the rows: single row, every row etc

Apply structure to 8-bit addition

PR #16, opened 2024-10-10
AIR table structure defintion

gio54321:
Issue #14

Here are the things implemented so far
- Made `Expression` type safe by providing the row and column numbers
- Defined the equality gadget
- Defined a first iteration of the AIR constraints and table structure, including the definition of the equivalence theorem
- Implemented a proof-of-concept table in which every row is a valid 8-bit addition

The main problems with the current equivalence statement definition are:
- the specification has to be provided generic in the `row`, but probably it is better to express a global property about the trace,
- we need to find a way to exploit some kind of inductive structure over the rows, but currently it is hard to express it.

In the 8-bit Fibonacci example, I think an expressive description should look like this (currently not quite there)
```lean
def constraints : List (TableConstraint 3 T p) := [
  TableConstraint.boundary 0 (λ row => Equality.circuit 3 T (var row 0) (const 1)),
  TableConstraint.boundary 0 (λ row => Equality.circuit 3 T (var row 1) (const 1)),
  TableConstraint.everyRowExceptLast (λ row => Addition8.circuit 3 T (var row 0) (var row 1) (var row 2) (var (row+1) 1)),
  TableConstraint.everyRowExceptLast (λ row => Equality.circuit 3 T (var row 1) (var (row+1) 1))
]

def spec := fun trace => (∀ row : ZMod T, (trace row 0).val = (fib row.val) % 256)
```
The equivalence should then be proved by induction on `row`. The base case is satisfied by the boundary constraints, and the equivalence reasoning should proceed by induction up until the last row.


gio54321:
Made a bit of progress, the first issue is solved as now the specification is just a Prop over the trace.
```lean
spec : Inputs N M (F p) -> Prop
```
Also gadgets lookups and table lookups have been decoupled to allow more optimized lookup structures over the tables.

The second issue is however still there. Fundamentally, we have two opposing facts:
- we want to represent row indices as elements of `ZMod M` so that we can reference rows at an offset from the current row, and we want to express the fact that this operation could wrap around the table, 
- we want to prove things by induction on the row.

Those two are incompatible because `ZMod M` does not have an inductive structure. Note that proving things by induction on the row intuitively is correct, but actually it is *not* possible in the general case, because constraints could reference rows above (or equivalently wrap around), breaking well-foundedness.
Note also that the [Fibonacci table](https://github.com/zksecurity/clean/blob/f8ed44bb05a33c7f5eaaea0ff264946fcc8fbc6b/Clean/Tables/Fibonacci.lean#L31) equivalence theorem would be easy to prove (or at least doable) if the table were infinite i.e., if the row indices were element of `Nat`. 

Here are some ideas for solving this:
- Make constraints only reference rows below the current one (not sure how we could express this, while also retaining information about the possibility of wrapping).
- Prove equivalence by induction on `M` (the table length).

mitschabaude:
First feedback, I noticed that `Circuit.lean` got out of date with other stuff and no longer builds. Feel free to delete anything in there that you don't use, but I think we should always keep the project building with `lake build`

gio54321:
Right, I removed all the unused code, I think we can come back later to the AIR polynomials at a later stage.
Also moved the main build targets to the tables, so that everything is checked by `lake build`. At some point I think it would make sense also to add a CI to ensure that there are no errors

mitschabaude (approved):

PR #17, opened 2024-10-17
Experiment: inductive trace definition

gio54321:
I've added some (very quick and dirty) code to research whether an inductive trace definition could be used. For now only the "all rows" constraints are supported.
It includes a quick PoC with add8 on every row.

### DIFF
--- a/Clean/Expression.lean
+++ b/Clean/Expression.lean
@@ -14,7 +14,6 @@ import Mathlib.Algebra.Field.Basic
 import Mathlib.Data.ZMod.Basic
 import Mathlib.Data.Real.Basic
 
-def Row (M : ℕ+) (F: Type) := Fin M -> F
 
 inductive Expression (N : ℕ+) (M : ℕ) (F : Type) where
   | var : Fin M -> Fin N -> Expression N M F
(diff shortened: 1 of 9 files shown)


gio54321:
Here's a summary of the changes. First of all the [trace definition](https://github.com/zksecurity/clean/blob/6ee3655ca748d4e4558f56fd8a751dcef1fe1fbd/Clean/Expression.lean#L33-L83) has been changed to be inductive. This includes also some new syntax which I think is nice to have:
```lean
<+> -- to indicate an empty trace
trace +> row -- to indicate some trace plus a row at the end
trace +> curr_row +> next_row -- this is how two adjacent rows are represented
```
but can be removed if is too much confusing :sweat_smile:.

Then with this trace it is defined a [Table](https://github.com/zksecurity/clean/blob/6ee3655ca748d4e4558f56fd8a751dcef1fe1fbd/Clean/Table.lean#L141), which is composed of table lookup assumptions, table constraints, a specification and an equivalence theorem. The main function used in the equivalence theorem is the one that [applies the constraints to every row of the trace](https://github.com/zksecurity/clean/blob/6ee3655ca748d4e4558f56fd8a751dcef1fe1fbd/Clean/Table.lean#L87), which applies an induction first on the constraints and then on the rows.

Two PoC tables are provided to test the approach:
- [Addition8](https://github.com/zksecurity/clean/blob/6ee3655ca748d4e4558f56fd8a751dcef1fe1fbd/Clean/Tables/Addition8.lean#L18) which simply constraints every row of the table to be a valid add8 instance, this do not actually require any inductive reasoning
- [Fibonacci](https://github.com/zksecurity/clean/blob/6ee3655ca748d4e4558f56fd8a751dcef1fe1fbd/Clean/Tables/Fibonacci.lean#L38) which do require inductive reasoning. The soundness proof is almost finished, and makes use of this nice utility [induction function](https://github.com/zksecurity/clean/blob/6ee3655ca748d4e4558f56fd8a751dcef1fe1fbd/Clean/Expression.lean#L53) to express induction over pairs of rows.

A few things to improve:
- the table constraints are defined using the `const` variant of expressions and are generic over one row or two rows, maybe the `Row` structure could be refactored to ensure that serialization is possible, e.g.,
```lean
def Row (M : ℕ+) (F: Type) := Fin M -> Expression N M F
-- instead of
def Row (M : ℕ+) (F: Type) := Fin M -> F
```
- It's not really clear from the current definition the assumptions on the trace length (e.g., in the Fibonacci one, `M` should be at least 2, in practice it should be always satisfied but it is not "checked" in the formalization, see also [this](https://github.com/zksecurity/clean/blob/6ee3655ca748d4e4558f56fd8a751dcef1fe1fbd/Clean/Table.lean#L121))

PR #19, opened 2024-10-25
Add CI

gio54321:
Add CI that automatically builds the project, with default config from [lean-action](https://github.com/leanprover/lean-action)


mitschabaude (approved):
nice! where is the `leanprover` action from?

gio54321:
It's from this repo https://github.com/leanprover/lean-action
I think it is used also in a few large lean projects

PR #23, opened 2024-12-03
Add 8 bit full adder gadget

gio54321:
Add a gadget that enforces the relation of an 8-bit full adder, with two boolean flags: an input carry and an output carry. Maybe it will not be used in Add32, but IMO is a nice gadget to have in any case.


Clean/Gadgets/Addition8Full.lean
```diff
@@ -0,0 +1,244 @@
...
+    [
+      carry_in + x + y - out - carry_out * (const 256)
+    ],
+    [
+      ByteLookup.lookup N M x,
+      ByteLookup.lookup N M y,
+      ByteLookup.lookup N M out,
+    ],
+    [
+      Boolean.circuit N M carry_in,
```
mitschabaude, line 30: I guess when passing carries from one to another instance of this gadget, they will be constrained twice -- once as `carry_in`, once as `carry_out`

PR #24, opened 2024-12-05
Add 32-bit addition

gio54321:
This PR addresses #6

The implementation is done by composing one 8-bits half adder and three 8-bits full adders, so no additional lookups and constraints are required. The nice thing is that the statement trivially translates to a statement with limbs over the naturals (thanks to the sub equivalence proofs), so the majority of the proof will prove the interesting statement, without much boilerplate.


gio54321:
Closing this, as add32 has already been addressed in other PRs

PR #31, opened 2024-12-16
[Circuit monad] Proper soundness using env function

mitschabaude:
* remove the possibility of passing in witness value inputs when running a circuit, and instead use the more general and sound notion that the prover can use any `env: Nat -> F` to determine assignment of values to variables
* modify `SubCircuit` structure to a form which I think will stand the test of time
* prove a larger part of the underlying justification for replacing subcircuits with their specs
  * reduce the theorem that allows us to use a `FormalCircuit` as a `SubCircuit` to the equivalence of flattened and nested constraints
* remove lots of (now) obsolete code

### DIFF
--- a/Clean/Circuit/Circuit.lean
+++ b/Clean/Circuit/Circuit.lean
@@ -700,10 +700,10 @@ def add8_full (input : Vector (Expression (F p)) 3) (locals: Option (Vector (F p
   let z := do let ⟨ [z, _], _ ⟩ ← locals; return z
   let carry_out := do let ⟨ [_, carry], _ ⟩ ← locals; return carry
 
-  let z ← witness_or_value z (fun () => mod_256 (x + y + carry_in))
+  let z ← witness (fun () => mod_256 (x + y + carry_in))
   byte_lookup z
 
-  let carry_out ← witness_or_value carry_out (fun () => floordiv (x + y + carry_in) 256)
+  let carry_out ← witness (fun () => floordiv (x + y + carry_in) 256)
   assert_bool carry_out
 
   assert_zero (x + y + carry_in - z - carry_out * (const 256))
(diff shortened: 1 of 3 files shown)


Clean/Circuit/Circuit.lean
```diff
@@ -714,34 +605,39 @@ def spec (input : Vector (F p) 3) (z: F p) :=
...
+    injection h_inputs with hx h_inputs
+    injection h_inputs with hy h_inputs
+    injection h_inputs with hcarry_in h_inputs
+    dsimp at hx hy hcarry_in
+    guard_hyp hx : x_var.eval_env env = x
+    guard_hyp hy : y_var.eval_env env = y
+    guard_hyp hcarry_in : carry_in_var.eval_env env = carry_in
+
+    -- we use a trick to get Lean to display the actual parts of `h_holds` for us!
+    have h_holds: _ ∧ _ ∧ _  := h_holds
```
mitschabaude: this is the trick I was missing before this PR

Clean/Circuit/Circuit.lean
```diff
@@ -808,57 +705,47 @@ def assumptions (input : Vector (F p) 2) :=
...
+    -- characterize inputs
+    injection h_inputs with h_inputs
+    injection h_inputs with hx h_inputs
+    injection h_inputs with hy h_inputs
+    dsimp at hx hy
+
+    -- simplify constraints hypothesis
+    -- we know it must be just the `subcircuit_soundness` of `Add8Full.circuit`
+    -- so it has an implication form
+    have h_holds : _ → _ := h_holds
```
mitschabaude: another example of the same trick

Clean/Circuit/Circuit.lean
gio54321: Actually, just using
```
dsimp [Adversarial.constraints_hold] at h_holds
```
simplifies things nicely and works as intended

Clean/Circuit/Circuit.lean
gio54321: Same thing, just doing
```
dsimp [Adversarial.constraints_hold] at h_holds
```
works :slightly_smiling_face:

mitschabaude: wait what, am I crazy

mitschabaude:
thanks @gio54321 :D I pushed an update

gio54321 (approved):
lgtm, nice!

I wonder if with the new `env` we can remove the `Provable.eval_env env b_var = b` assumption. We could quantify only over all possible `beta.var` and then introduce in the proofs the values that are the evaluation of those vars, since the env maps every var to a field element. Something like this (not really deeply tested)
```lean
  soundness:
    -- for all environments that determine witness generation
    ∀ env: ℕ → F,
    -- for all inputs that satisfy the assumptions
    ∀ b_var : β.var,
    let b := Provable.eval_env env b_var
    assumptions b →
    -- if the constraints hold
    Adversarial.constraints_hold env (main b_var) →
    -- the spec holds on the input and output
    let a := Provable.eval_env env (output (main b_var))
    spec b a
```

mitschabaude:
> I wonder if with the new `env` we can remove the `Provable.eval_env env b_var = b` assumption

I thought it's a good idea and tried it. So far it made the introduction of variables a bit more complicated though, because I still need the same equality assumption and need to introduce it manually (see screenshot). Maybe you can come up with a better way to handle this later

image

PR #32, opened 2024-12-17
[Circuit monad] Shared context

mitschabaude:
Instead of instantiating subcircuits with a new `Context F` -- which determines the numbering of variables --, we pass the current context down to subcircuits

Thanks to this, in the final list of constraints we now have distinct indices for distinct vars, which seems important


gio54321 (approved):
lgtm!
I agree, this is a cleaner way to structure things and does not change much the proofs

PR #33, opened 2024-12-17
[Circuit monad] Begin integration of the new circuit definition

gio54321:
- Begin refactoring and integrating the new circuit definition into the existing codebase, forking from the `eval-env-soundness` branch.
- Eventually when we will integrate also the trace definition, the `GadgetsNew` folder will be renamed into `Gadgets`.
- This PR only moves code around, it does not change anything else.
- Changes the target of the CI to the new formulation, which is incompatible with the old one (for example because of the clashing `Expression` definition)


Clean/Circuit/Basic.lean
```diff
@@ -448,336 +448,3 @@ def subcircuit (circuit: FormalCircuit F β α) (b: β.var) := as_stateful (F:=F
...
-  let p := 1009
-  let p_prime := Fact.mk prime_1009
-  let p_non_zero := Fact.mk (by norm_num : p ≠ 0)
-  let p_large_enough := Fact.mk (by norm_num : p > 512)
-  let main := do
-    let x ← witness (fun _ => 10)
-    let y ← witness (fun _ => 20)
-    add8_wrapped (p:=p) (vec [x, y])
-  main.operations
-end
```
mitschabaude: maybe we could move that to some test file? during development it was pretty useful to see how the printed constraints changed (or when they couldn't be printed)

gio54321: Yes! I think it got lost into the different files.

Also, we should at some point define a custom command which nicely displays some information about a circuit, ideally without having to resort to concretize a specific prime or specific witness values. This may be non-trivial to do though.

PR #39, opened 2025-01-09
Formalize assertion-like subcircuits

mitschabaude:
I wanted to embed the `assert_bool` circuit in our framework but realized it's not a `FormalCircuit`: by design, `assert_bool` is not complete on all inputs - it will of course fail on non-boolean inputs!

This lead me to formulate an alternative called `FormalAssertion` to capture circuits like this. The low-level representation is the same -- `Operation.SubCircuit` - but the structure of hypotheses is slightly different.

To test this addition I also made `assert_bool` a subcircuit inside `add8`. This is done in the last commit and is not strictly necessary -- because assert_bool is so simple there's not much gained from abstracting it. But I think it doesn't hurt either

### DIFF
--- a/Clean/Circuit/Basic.lean
+++ b/Clean/Circuit/Basic.lean
@@ -281,6 +281,52 @@ def subcircuit_soundness (circuit: FormalCircuit F β α) (b_var : β.var) (a_va
 def subcircuit_completeness (circuit: FormalCircuit F β α) (b_var : β.var) :=
   let b := Provable.eval F b_var
   circuit.assumptions b
+
+/--
+`FormalAssertion` models a subcircuit that is "assertion-like":
+- it doesn't return anything
+- by design, it is not complete: it further constrains its inputs
+
+The notion of _soundness_ is the same as for `FormalCircuit`: some `assumptions` + constraints imply a `spec`.
+
+However, the _completeness_ statement is weaker:
+If both the assumptions AND the spec are true, then the constraints hold.
+
+In other words, for `FormalAssertion`s the spec must be an equivalent reformulation of the constraints.
+(In the case of `FormalCircuit`, the spec can be strictly weaker than the constraints.)
+-/
+structure FormalAssertion (F: Type) (β: TypePair) [Field F] [ProvableType F β] where
+  main: β.var → Circuit F Unit
+
+  assumptions: β.value → Prop
+  spec: β.value → Prop
+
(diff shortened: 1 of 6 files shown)


Clean/Circuit/Basic.lean
```diff
@@ -281,6 +281,52 @@ def subcircuit_soundness (circuit: FormalCircuit F β α) (b_var : β.var) (a_va
...
+
+/--
+`FormalAssertion` models a subcircuit that is "assertion-like":
+- it doesn't return anything
+- by design, it is not complete: it further constrains its inputs
+
+The notion of _soundness_ is the same as for `FormalCircuit`: some `assumptions` + constraints imply a `spec`.
+
+However, the _completeness_ statement is weaker:
+If both the assumptions AND the spec are true, then the constraints hold.
```
gio54321, line 293: > the _completeness_ statement is weaker

Is it? I'm not sure, this seems to me actually a stronger formulation (very similar to the one we had before), where given that `assumptions` are true, then there is an iff between the constraints and the specification.

gio54321 (approved):
LGTM! I encountered the same exact problem when trying to implement a simple gadget that asserted that two input elements were equal.

I wonder if the two distinct formalizations could be merged by treating the inputs and outputs as just variables and asserting a relation between them. To me it seems a bit weird to have a very strong distinction between inputs and outputs of a gadget. I guess it is very natural when thinking about computation flow, but what a gadget ultimately does is enforcing a set of constraints between variables.

Clean/Circuit/Basic.lean
mitschabaude: yeah I agree with what you said so depends on what exactly you call weaker 😅 

to me, `assumptions AND spec => constraints` is a weaker statement than `assumptions => constraints`, because to use the first, you have to prove more yourself (`assumption AND spec`)

mitschabaude:
> I wonder if the two distinct formalizations could be merged by treating the inputs and outputs as just variables and asserting a relation between them.

Yeah I guess there could be a middle ground as well -- `FormalAssertion` with outputs. It wasn't really necessary to remove outputs from `FormalAssertion` (you can always make `unit` the output if you don't need one), I just liked the simpler interface for `assert_bool`.

For me personally, I like that there is a common low-level representation as `SubCircuit` but we are able to provide different interfaces to that depending on what is most natural for the gadget.

PR #43, opened 2025-01-10
Add32 proofs in new formulation

### LINKED ISSUE #6: 32-bit addition circuit

mitschabaude:
closes #6

* proofs for the new add32 gadget (with carry)
* new tactic script `field_to_nat_u32`


Clean/Types/U32.lean
```diff
@@ -44,6 +47,12 @@ def is_normalized (x: U32 (F p)) :=
 def value (x: U32 (F p)) :=
   x.x0.val + x.x1.val * 256 + x.x2.val * 256^2 + x.x3.val * 256^3
 
+/--
+Return the value of a 32-bit unsigned integer as a field
```
marcobesier: There is a missing word in this comment:

`Return the value of a 32-bit unsigned integer as a field **element.**`

Admittedly, it is a non-issue, but we might still want to fix it for the sake of clarity.

marcobesier (approved):
This is very nice and clean! There's a minor typo in one comment, but this surely doesn't block approval. :)

Clean/Types/U32.lean
mitschabaude: ah, yep I'm actually used to just saying "field" instead of "field element", because this was the widely used shorthand in the ecosystem I worked in (Mina / o1js).
(in Starkware/Cairo they say "felt" :D)

marcobesier: ah, got it! felt is next level, tho 😂

PR #44, opened 2025-01-14
Decouple witness generation and variables

gio54321:
- [ ] Add a default environment which represents the assignment of variables to field elements given by the default witness generator
- [ ] Prove new foundational theorems
- [ ] Fix all the gadget proofs


gio54321:
Closing this as this PR is not relevant anymore

PR #47, opened 2025-01-21
Add 32-bit Fibonacci trace

gio54321:
- Define the `StructuredElements` typeclass, which is a way to characterize structures that hold some elements, and can be flattened and unflattened into a vector
- Change trace definition so that every row is a structure implementing `StructuredElements`
- Add equality gadget for U32
- Update add8 and fib8 tables
- Add fib32 table


mitschabaude (approved):
great work!

PR #48, opened 2025-01-22
Foundation: Unique variables indices and reformulated completeness

mitschabaude:
massive refactor (sorry)
* remove witness generator attached to variables
* add offset as type argument to subcircuits to that we can formulate the completeness condition
* redefine circuit operations as inductive datatype so that subcircuit offset must be consistent
* reformulate "constraints hold" in terms of inductive operations instead of a list
* refactor foundational theorems
* a bunch of other changes


Clean/Gadgets/Addition8/Addition8FullCarry.lean
```diff
@@ -130,41 +128,52 @@ def circuit : FormalCircuit (F p) (Inputs p) (Outputs p) where
...
+      rwa [hx, hy, hcarry_in, hcout'] at henv1
 
     -- now it's just mathematics!
     guard_hyp as : x.val < 256 ∧ y.val < 256 ∧ (carry_in = 0 ∨ carry_in = 1)
 
     let goal_byte := ByteTable.contains (vec [z])
     let goal_bool := carry_out = 0 ∨ carry_out = 1
     let goal_add := x + y + carry_in + -1 * z + -1 * (carry_out * 256) = 0
-    show goal_byte ∧ (True ∧ goal_bool) ∧ goal_add
+    show ((True ∧ goal_byte) ∧ True ∧ goal_bool) ∧ goal_add
```
mitschabaude, line 169: TODO: where does the extra `True` come from?
(the change in parenthesis comes from iteration order for Lists vs our inductive type)

marcobesier (approved):
Looks good to me. 👍 CC: @gio54321

PR #57, opened 2025-02-18
Simplify `ProvableType`

mitschabaude:
* redefine `ProvableType` to be like `StructuredElements`: derived from a generic type `Type -> Type` instead of a "type pair". i.e. the var type is always `M (Expression F)` while the value type is `M F`
* replace `StructuredElements` with `ProvableType` (or, sometimes, with `NonEmptyProvableType` which has an extra size > 0 condition)
* make the new `ProvableType` _not_ depend on the field. instead, the individual methods `to_elements`, `from_elements` are generic over the input type

IMO, these changes make instantiating and using `ProvableType`s much simpler. type inference is also better, e.g. we can now define `Provable.witness` without an explicit type argument.


Clean/Gadgets/Addition8/Addition8.lean
```diff
@@ -4,47 +4,35 @@ namespace Gadgets.Addition8
...
   x: F
   y: F
 
-def Inputs (p : ℕ) : TypePair := ⟨
-  InputStruct (Expression (F p)),
-  InputStruct (F p)
-⟩
-
-instance : ProvableType (F p) (Inputs p) where
+instance : ProvableType Inputs where
```
mitschabaude, line 11: not having `ProvableType` depend on a field makes these definitions much nicer

Clean/Circuit/Provable.lean
```diff
@@ -3,116 +3,97 @@ import Clean.Utils.Vector
 import Clean.Circuit.Expression
 import Clean.Circuit.SimpGadget
 
+/--
+'Provable types' are structured collections of field elements.
+
+ We represent them as types generic over a single type argument (the field element),
+ i.e. `Type → Type`.
+-/
+def TypeMap := Type → Type
```
gio54321, line 13: nit: make this `@[reducible]`?

Clean/Circuit/Provable.lean
```diff
@@ -3,116 +3,97 @@ import Clean.Utils.Vector
...
+
+def Var (M: Type → Type) (F : Type) := M (Expression F)
```
gio54321: nit: maybe it is nicer to have `(M: TypeMap)`?
(same below in `ProvableType` and others)

gio54321 (approved):
Very nice!
Indeed improves a lot readability

PR #58, opened 2025-02-20
Improve simp sanity

gio54321:
- Add two custom simp sets for gadgets and traces, so that specific theorems are not included in the general-purpose simp set
- Replace most of `simp` with `simp only` in the core theorems

### DIFF
--- a/Clean/Circuit/Basic.lean
+++ b/Clean/Circuit/Basic.lean
@@ -76,7 +76,8 @@ def witnesses (env: Environment F) : (l: List (FlatOperation F)) → Vector F (w
     let ws := witnesses env ops
     match op with
     | witness m compute =>
-      ⟨ (compute env).val ++ ws.val, by simp [ws.prop, gadget_norm]; ac_rfl ⟩
+      ⟨ (compute env).val ++ ws.val, by simp only [ws.prop, gadget_norm,
+        List.length_append, Mathlib.Vector.length_val]; ac_rfl ⟩
     | assert _ | lookup _ =>
       ⟨ ws.val, by simp_all only [witness_length, ws.prop]⟩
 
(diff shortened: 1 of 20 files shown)


Clean/Circuit/Basic.lean
```diff
@@ -117,7 +120,7 @@ structure SubCircuit (F: Type) [Field F] (offset: ℕ) where
   implied_by_completeness : ∀ env, env.extends_vector (FlatOperation.witnesses env ops) offset →
     completeness env → FlatOperation.constraints_hold_flat env ops
 
-@[reducible, simp]
+@[reducible, simp, gadget_norm]
```
mitschabaude: any reason to keep the simp attribute here?

Clean/Circuit/Basic.lean
```diff
@@ -147,23 +150,23 @@ def initial_offset {n: ℕ} : Operations F n → ℕ
   | .lookup ops _ => initial_offset ops
   | .subcircuit ops s => initial_offset ops
 
-@[simp]
+@[gadget_norm]
```
mitschabaude: function above this one still has a simp

Clean/Circuit/SubCircuit.lean
```diff
@@ -58,7 +58,6 @@ theorem can_replace_subcircuits {n: ℕ} : ∀ {ops : Operations F n}, ∀ {env
 /--
 Theorem and implementation that allows us to take a formal circuit and use it as a subcircuit.
 -/
-@[simp]
 def formal_circuit_to_subcircuit (n: ℕ)
```
mitschabaude, line 61: it seems you need to pass this to simp in a proof below, so should be added to `gadget_norm`

Clean/Circuit/SubCircuit.lean
```diff
@@ -117,7 +116,6 @@ def formal_circuit_to_subcircuit (n: ℕ)
 /--
 Theorem and implementation that allows us to take a formal assertion and use it as a subcircuit.
 -/
-@[simp]
 def formal_assertion_to_subcircuit (n: ℕ)
```
mitschabaude, line 119: add to `gadget_norm`?

mitschabaude (approved):
nice! just a few comments

also I think `circuit_norm` might be more consistent with our naming than `gadget_norm`

Clean/Circuit/SubCircuit.lean
gio54321: In my mind, in the simp set there should only be things that in most cases we want to expand.
Turns out that this definitions cause a lot of growth of the term size when rewriting big terms, so I wanted to experiment removing them from the simp set and expanding them manually as needed.
We can always add them later to the set

mitschabaude: Ok makes sense!

PR #61, opened 2025-03-10
Modify the Circuit monad to be a state monad

gio54321:
- Modify the `Circuit` monad to be a `StateM` for better scalability
- Change the `FormalCircuit` and `FormalAssertion` definitions to include a proof of the main circuit being well-formed


mitschabaude (approved):
this is a real breakthrough!

PR #62, opened 2025-03-11
Make TableConstraint a state monad

mitschabaude:
- changes `TableConstraint` to be just `StateM (TableContext ...)`
- ensure all proofs still work
- misc. cleanup, esp. in `Table/Basic.clean`

PR #63, opened 2025-03-13
Remove formal table assumptions

mitschabaude:
This PR removes the ability to require assumptions on the entire table in the `FormalTable` soundness proof. Only an assumption on the number of rows is allowed.

It turns out that this works well for all our currently defined tables. In the `Addition8` table, extra byte lookups have to be added on the inputs. In the fibonacci tables, what used to be an assumption naturally follows from the induction hypothesis, if added to the spec.

No assumptions leads to stronger statements for individual tables, but decreases flexibility somewhat when combining tables. Scenarios could arise where constraints have to be duplicated just because we can't have assumptions on the inputs. If this turns out to be the case it will be easy to add assumptions back in.

### DIFF
--- a/Clean/Circuit/Basic.lean
+++ b/Clean/Circuit/Basic.lean
@@ -404,8 +404,9 @@ where
   initial_offset_eq: ∀ var, ∀ n, (main var |>.operations n).initial_offset = n := by intros; rfl
 
 @[circuit_norm]
-def subcircuit_soundness (circuit: FormalCircuit F β α) (b_var : Var β F) (a_var : Var α F) (env : Environment F) :=
+def subcircuit_soundness (circuit: FormalCircuit F β α) (b_var : Var β F) (offset: ℕ) (env : Environment F) :=
   let b := eval env b_var
+  let a_var := circuit.main b_var offset |>.fst
   let a := eval env a_var
   circuit.assumptions b → circuit.spec b a
 
--- a/Clean/Tables/Addition8.lean
+++ b/Clean/Tables/Addition8.lean
@@ -19,12 +19,26 @@ instance : NonEmptyProvableType RowType where
     let ⟨ [x, y, z], _ ⟩ := v
     ⟨ x, y, z ⟩
 
+def byte_lookup_circuit : FormalAssertion (F p) Provable.field where
+  main x := byte_lookup x
+  assumptions _ := true
+  spec x := x.val < 256
+  soundness := by
+    intro _ env x_var x hx _ h_holds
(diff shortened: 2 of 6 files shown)


Clean/Circuit/Basic.lean
```diff
@@ -404,8 +404,9 @@ where
   initial_offset_eq: ∀ var, ∀ n, (main var |>.operations n).initial_offset = n := by intros; rfl
 
 @[circuit_norm]
-def subcircuit_soundness (circuit: FormalCircuit F β α) (b_var : Var β F) (a_var : Var α F) (env : Environment F) :=
+def subcircuit_soundness (circuit: FormalCircuit F β α) (b_var : Var β F) (offset: ℕ) (env : Environment F) :=
```
mitschabaude, line 407: unrelated change, it felt more natural to not make the output variable a parameter here

Clean/Tables/Addition8.lean
```diff
@@ -19,36 +19,44 @@ instance : NonEmptyProvableType RowType where
     let ⟨ [x, y, z], _ ⟩ := v
     ⟨ x, y, z ⟩
 
+def byte_lookup_circuit : FormalAssertion (F p) Provable.field where
```
mitschabaude, line 22: our simplest formal circuit so far :)

Clean/Tables/Fibonacci32.lean
```diff
@@ -323,17 +313,15 @@ def formal_fib32_table : FormalTable (F p) RowType := {
...
     -- inductive step
-    · intro lookup_h
-      simp only [TraceOfLength.forAllRowsOfTrace.inner, Fin.isValue] at lookup_h
-
-      -- first of all, we prove the inductive part of the spec
+    · -- first of all, we prove the inductive part of the spec
       unfold TraceOfLength.forAllRowsOfTraceWithIndex.inner
       intros constraints_hold
-      specialize ih2 lookup_h.right
+      -- specialize ih2 lookup_h.right
```
marcobesier: `specialize ih2 lookup_h.right` can be removed instead of commented out, right?

marcobesier (approved):
Ready to merge. 🙂 You might want to remove the

` -- specialize ih2 lookup_h.right`

 comment before merging, though.

Clean/Tables/Fibonacci32.lean
mitschabaude: right, appreciate the diligence!

mitschabaude:
> Ready to merge. 🙂

awesome, will merge after parent PR #62

PR #64, opened 2025-03-20
Update Lean and add `ElaboratedCircuit`

mitschabaude:
- update lean and mathlib to v4.17
- expose more info about subcircuits to avoid reasoning about their internals in parent circuits

steps to udpate lean taken from here: https://malv.in/posts/2024-12-09-howto-update-lean-dependencies.html
tl;dr: change the versions manually, delete the manifest, and run `lake update -R`


Clean/Circuit/Basic.lean
```diff
@@ -364,8 +363,39 @@ def constraints_hold.completeness {n : ℕ} (eval : Environment F) : Operations
...
+Common base type for circuits that are to be used in formal proofs.
+
+It contains the main circuit plus some of its properties in elaborated form, to make it
+faster to reason about them in proofs.
+-/
+class ElaboratedCircuit (F: Type) [Field F] (β: TypeMap) [ProvableType β] (Output: Type) where
+  main: Var β F → Circuit F Output
+
+  /-- how many local witnesses this circuit introduces -/
+  local_length: Var β F → ℕ
```
gio54321, line 376: Why does `local_length` depends on the input variables?
Are there cases where it is not just a constant?

Clean/Utils/Field.lean
```diff
@@ -18,115 +19,79 @@ theorem ext {x y : F p} (h : x.val = y.val) : x = y := by
...
+- the equation to prove as the goal
+- size assumptions on variables and a sufficient `p > ...` instance
+
+if no sufficient inequalities are in the context, then the tactic will leave an equation of the form `expr : Nat < p` unsolved.
+
+note: this version is optimized specifically for byte arithmetic:
+- specifically handles field constant 256
+- expects `[Fact (p > 512)]` in the context
+-/
+syntax "field_to_nat" : tactic
```
gio54321, line 49: very nice!

Clean/Gadgets/Addition32/Addition32Full.lean
```diff
@@ -54,7 +54,19 @@ def spec (input : Inputs (F p)) (out: Outputs (F p)) :=
   ∧ carry_out.val = (x.value + y.value + carry_in.val) / 2^32
   ∧ z.is_normalized ∧ (carry_out = 0 ∨ carry_out = 1)
 
-theorem soundness : Soundness (F p) Inputs Outputs add32_full assumptions spec := by
+def circuit32 := Gadgets.Addition32Full.add32_full (p:=p_babybear) default
+#eval circuit32.operations.local_length
+#eval circuit32.output
```
gio54321: maybe move those evals to the examples folder?

Clean/Circuit/Basic.lean
```diff
@@ -374,39 +404,34 @@ def Soundness (F: Type) (β α: TypeMap) [Field F] [ProvableType α] [ProvableTy
...
   -- β = inputs, α = outputs
-  main: Var β F → Circuit F (Var α F)
   assumptions: β F → Prop
   spec: β F → α F → Prop
-  soundness: Soundness F β α main assumptions spec
-  completeness: Completeness F β α main assumptions
-  /-- technical condition needed for subcircuit consistency. usually automatically proved by `rfl`. -/
-  initial_offset_eq: ∀ var, ∀ n, (main var |>.operations n).initial_offset = n := by intros; rfl
+  soundness: Soundness F assumptions spec
+  completeness: Completeness F α assumptions
```
gio54321, line 429: To me it is somewhat strange that `soundness` and `completeness` do not explicitly take in input the circuit `main`. I assume that it gets bound automatically to the `main` from the `extends ElaboratedCircuit F β (Var α F)` but maybe we should make this more clear.

In general, should the `FormalCircuit` be parametrized explicitly an `ElaboratedCircuit` instead of extending it? Imo it would clarify the relation between them, which right now is somewhat opaque.

gio54321 (approved):
Great PR!
Left a few comments, but in general LGTM

Clean/Circuit/Basic.lean
mitschabaude: > To me it is somewhat strange that soundness and completeness do not explicitly take in input the circuit main. 

yeah it works simply because type class inference can find the `ElaboratedCircuit` argument that `Soundness` expects. but it makes sense to pass in `main` explicitly to make sure we are talking about the correct circuit :D

in general the relationship was intuitive to me: `FormalCircuit` has the same fields as `ElaboratedCircuit`, plus a few more (that can also use those from `ElaboratedCircuit`).

but I also found that the `extends` makes things sometimes more opaque in the Lean infoview, where you'll see something like `ElaboratedCircuit.local_length` and you have to hover over it to see what circuit it comes from.. I'll change it!

Clean/Gadgets/Addition32/Addition32Full.lean
mitschabaude: yeah makes sense! I wanted to document how to quickly get the info for `ElaboratedCircuit`

Clean/Circuit/Basic.lean
mitschabaude: yes, in my experience such a dependency is quite common:
* sometimes you need a variable but you have any expression, so depending on the input you might witness a new variable or not. e.g. here: https://github.com/Verified-zkEVM/clean/blob/5219b86b3740e45ee1d93bd33c65ab5aad7e2b81/Clean/Table/Basic.lean#L341-L347
* there might come situations where you want to behave differently given a constant than given another expression, to save constraints. for example, if all inputs are constant you only want to reduce them to a new constant instead of adding witnesses and constraints. in o1js we did this all the time

gio54321: Oh, yes this makes a lot of sense!

Clean/Circuit/Basic.lean
mitschabaude: @gio54321 I experimented a bit with changing the dependencies.
* elaborated circuit as a separate property: https://github.com/Verified-zkEVM/clean/compare/update-lean...experiment-elaborated-property
* elaborated circuit as an explicit parameter to formal circuit: https://github.com/Verified-zkEVM/clean/compare/update-lean...experiment-elaborated-parameter

the second one is better, but I don't like it better than the version in this PR. the explicit parameter makes reasoning about formal circuits more verbose, and defining them less intuitive IMO

mitschabaude: I'm gonna merge the PR as is for now!

PR #65, opened 2025-03-21
Refactor Table monad to extend Circuit

mitschabaude:
Stacked behind https://github.com/Verified-zkEVM/clean/pull/64

This refactors the `TableConstraint` monad in several ways:
* changes the assignment to be an explicit vector instead of a function
* allow variable assignment to cells outside the "main" input/output trace window. we don't really have to deal with those yet for soundness, but it will be possible to do so for completeness
* make the table context just be a circuit + assignments
* define a `MonadLift` from `Circuit`, which makes it possible to _directly_ use circuit operations in tables

All of the proofs are updated to work again. The new way of defining the environment causes a bit of extra work in proofs. Other than that, effort stays roughly the same. Overall this seems easily worth it for the power added to tables, and the more realistic outlook to prove completeness.


Clean/Table/Basic.lean
```diff
@@ -125,249 +125,249 @@ end TraceOfLength
...
-  | Assign : Variable F → CellOffset W S → TableConstraintOperation W S F
+The mapping must maintain the invariant that each variable is assigned to at most one cell.
+On the other hand, a cell can be assigned zero, one or more variables.
+-/
+structure CellAssignment (W: ℕ+) (S : Type → Type) [ProvableType S] where
+  offset : ℕ -- number of variables
+  aux_length : ℕ -- maximum number of auxiliary cells (i.e. those not part of the input/output layout)
+
+  /-- every variable is assigned to exactly one cell in the trace -/
+  vars : Vector (Cell W S) offset
```
mitschabaude, line 170: we redefine a `CellAssignment` to be a vector of variable assignments, rather than a function.

now, _all_ variables get assigned, and they can be either assigned to window cells or "auxiliary" cells, which hold any internal vars of our circuit. the idea is that some post processing could automatically add those aux vars at the end of the last row of each window

Clean/Table/Basic.lean
```diff
@@ -125,249 +125,249 @@ end TraceOfLength
...
+-/
+@[reducible, table_norm, table_assignment_norm]
+instance : MonadLift (Circuit F) (TableConstraint W S F) where
+  monadLift circuit ctx :=
+    let (a, ops) := circuit ctx.circuit
+    (a, {
+      circuit := ops,
+      -- the updated assignment is computed from a fresh starting circuit, independent of the circuit so far.
+      -- (if we would use `ops` instead of `circuit.operations 0`, we would be redoing previous assignments)
+      assignment := assignment_from_circuit ctx.assignment (circuit.operations 0)
```
mitschabaude, line 261: this is the main change! we can now use circuit operations directly inside a table

Clean/Table/Basic.lean
```diff
@@ -125,249 +125,249 @@ end TraceOfLength
...
   table .empty |>.snd.assignment
 
+@[table_assignment_norm]
+def offset_consistent (table : TableConstraint W S F α) : Prop :=
+  table.final_offset = table.final_assignment.offset
+
+-- construct an env by taking the result of the assignment function for input/output cells,
+-- and allowing arbitrary values for aux cells and invalid variables
+def window_env (table : TableConstraint W S F Unit)
+  (window: TraceOfLength F S W) (aux_env : Environment F) : Environment F :=
```
mitschabaude, line 284: the definition of environment is more complicated now, and this is a source of annoyance when unfolding proofs. usually it's necessary to first unfold constraints with an opaque environment, and separately unfold the environment later.

on the positive side, this definition is much closer to an actual implementation. we no longer assume all variables are assigned to an input/output cell (which was unrealistic and would have prevented completeness)

Clean/Table/Basic.lean
```diff
@@ -125,249 +125,249 @@ end TraceOfLength
...
-  Mapping from the index of a variable to a cell offset in the table.
--/
-@[reducible]
-def CellAssignment (W: ℕ+) (S : Type → Type) [ProvableType S] := ℕ → CellOffset W S
-
-/--
-  Atomic operations for constructing a table constraint, which is a constraint applied to a window
-  of rows in a table.
--/
-inductive TableConstraintOperation (W : ℕ+) (S : Type → Type) (F : Type) [Field F] [ProvableType S] where
```
mitschabaude, line 158: we no longer have a separate type for table operations! a table operation is either a circuit operation, and/or it directly changes the `ctx.assignment`

mitschabaude (commented):
some comments for reviewers!

gio54321 (approved):
LGTM, very nice!
I have a few ideas to simplify a bit the environment definition, but it requires some experimentation, so let's not wait to get this PR merged!

PR #67, opened 2025-03-31
Keccak blueprint

mitschabaude:


mitschabaude:
@gio54321 @marcobesier I think this is a very useful starting point for the keccak gadgets. I'm merging this PR, so that we can better parallelize further work on it

PR #68, opened 2025-03-31
Preserve Structure of `ProvableType`s

mitschabaude:
This PR adds several ways to preserve the structure of provable types inside proofs, so that we can reason at a high level about types like U32 or U64, and not about their individual field elements (bytes).

Using this, I managed to both significantly simplify the fib32 proof, and to prove soundness of the Keccak 'theta' gadget (with some lower level gaps). clean scales fairly easily to both of these examples now!

## Details

* add `ProvableStruct`, a way to specify a `ProvableType` from it's _higher-level components_
  * example from `Addition32Full`:
```lean
structure Inputs (F : Type) where
  x: U32 F
  y: U32 F
  carry_in: F

instance : ProvableStruct Inputs where
  components := [U32, U32, field]
  to_components := fun {x, y, carry_in} => .cons x ( .cons y ( .cons carry_in .nil))
  from_components := fun (.cons x ( .cons y ( .cons carry_in .nil))) => ⟨ x, y, carry_in ⟩
```
* add `ProvableVector` to talk about vectors of high-level types
  * example from `ThetaC`:
```lean
@[reducible] def State := ProvableVector U64 25
@[reducible] def Outputs := ProvableVector U64 5
```
* Remove `ProvableType.eval` and similar operations from `circuit_norm`, to prevent unfolding into individual field elements. Instead, add theorems like `ProvableType.eval env x = ProvableStruct.eval env x` to the simp set, which will only unfold the high-level `ProvableStruct` structure
* add a new `LawfulProvableType` class which adds the requirement that `to_elements` and `from_elements` are inverses. this enables to prove an additional theorem which was useful in simplification:
```lean
theorem eval_const {F : Type} [Field F] {α: TypeMap} [LawfulProvableType α]
  (env : Environment F) (x : α F) :
    eval env (const x) = x
```

### DIFF
--- a/Clean/Circuit/Provable.lean
+++ b/Clean/Circuit/Provable.lean
@@ -31,7 +31,7 @@ class NonEmptyProvableType (M : TypeMap) extends ProvableType M where
 export ProvableType (size to_elements from_elements)
 
 attribute [circuit_norm] size
--- tagged with low priority to prefer higher-level `Components` decompositions
+-- tagged with low priority to prefer higher-level `ProvableStruct` decompositions
 attribute [circuit_norm low] to_elements from_elements
 
 variable {M : TypeMap} [ProvableType M]
@@ -46,7 +46,7 @@ def from_vars (vars: Vector (Expression F) (size M)) := from_elements vars
 namespace Provable
 variable {α β γ: TypeMap} [ProvableType α] [ProvableType β] [ProvableType γ]
 
--- tagged with low priority to prefer higher-level `Components` decompositions
+-- tagged with low priority to prefer higher-level `ProvableStruct` decompositions
 @[circuit_norm low]
 def eval (env: Environment F) (x: Var α F) : α F :=
   let vars := to_vars x
@@ -120,7 +120,7 @@ end Provable
 
 export Provable (eval)
 
-namespace Components
+namespace ProvableStruct
 structure WithProvableType where
   type : TypeMap
   provable_type : ProvableType type := by infer_instance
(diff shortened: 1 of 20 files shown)


Clean/Circuit/Provable.lean
```diff
@@ -28,37 +28,82 @@ class ProvableType (M : TypeMap) where
...
 
 @[circuit_norm]
 def to_vars (var: M (Expression F)) := to_elements var
 
 @[circuit_norm]
-def from_vars (vars: Vector (Expression F) (ProvableType.size M)) := from_elements vars
+def from_vars (vars: Vector (Expression F) (size M)) := from_elements vars
+
+class LawfulProvableType (M : TypeMap) extends ProvableType M where
+  to_elements_from_elements {F: Type} : ∀ v: Vector F size, to_elements (from_elements v) = v
```
gio54321, line 46: Now that I think about it, is there a reason why there are separate `ProvableType` and `LawfulProvableType`? Why not to include this theorem directly in the `ProvableType` structure?

Clean/Circuit/Provable.lean
```diff
@@ -67,53 +112,214 @@ instance : ProvableType unit where
...
+where
+  eval_eq_eval_struct_aux (env : Environment F) : (cs : List WithProvableType) → (as : ProvableTypeList (Expression F) cs) →
+    eval.go env cs as = (components_to_elements cs as |> Vector.map (Expression.eval env) |> components_from_elements cs)
+  | [], .nil => rfl
+  | c :: cs, .cons a as => by
+    simp only [components_to_elements, components_from_elements, eval.go, ProvableType.eval, to_vars]
+    rw [Vector.map_append, Vector.cast_take_append_of_eq_length, Vector.cast_drop_append_of_eq_length]
+    congr
+    -- recursively use this lemma!
+    apply eval_eq_eval_struct_aux
```
gio54321, line 247: Nice!

Clean/Gadgets/Xor/ByteXorTable.lean
```diff
@@ -20,24 +20,20 @@ def ByteXorTable: Table (F p) where
...
-    (hx : x.val < 256)
-    (hy : y.val < 256) :
+    (x y z: F p) :
     ByteXorTable.contains (#v[x, y, z]) → z.val = Nat.xor x.val y.val := by
   sorry
 
 def ByteXorTable.completeness
-    (x y z: F p)
-    (hx : x.val < 256)
-    (hy : y.val < 256) :
```
gio54321, line 32: For `ByteXorTable.soundness` I agree `hx` and `hy` can be removed, as we can prove that
```
ByteXorTable.contains (#v[x, y, z]) -> x.val < 256 ∧ y.val < 256 ∧ z.val < 256
```
However, for `ByteXorTable.completeness` those assumptions I think are needed, because it is not true in general for every field element `x, y, z` that if `Nat.xor x.val y.val = z.val` then the triple is in the table.

Actually, this xor table serves also as a lookup, so we could write a stronger equiv theorem statement
```
ByteXorTable.contains (#v[x, y, z]) ↔ z.val = Nat.xor x.val y.val ∧ x.val < 256 ∧ y.val < 256 ∧ z.val < 256
```

gio54321 (approved):
LGTM, this is a huge step for scalability!
Left a few minor comments

Clean/Gadgets/Xor/ByteXorTable.lean
mitschabaude: Right, I forgot that! The second suggestion seems good to me

Clean/Circuit/Provable.lean
mitschabaude: I was afraid that it increases the burden of defining new provable types, since you have to prove those theorems every time. And the benefit of having them is currently pretty small.

I did find a default tactic that so far handles all "directly defined" provable types automatically (but not structs etc). Structs should also be very automatic. Let's see, we can still expand the lawfulness requirement later if that becomes useful

PR #71, opened 2025-04-08
`LawfulCircuit` and generic `assert_equals` Gadget

mitschabaude:
in this PR I was searching for simpler ways of reasoning about circuit constraints, for example, inductively starting from the beginning. the result is a successful treatment of simple loops, and some hints at future use cases.

* define what it means to "append" two instances of circuit `Operations`
* define a class `LawfulCircuit` which states that
  * the circuit is "append-only", it acts on the input operations only by appending more
  * the circuit's output and length do not depend on the input operations
* develop a tactic that can infer `LawfulCircuit`, and a few theorems about them
* show that simple for loops, using `forM` applied to circuits, are lawful circuits as well, and characterize the constraints they create
* implement a generic equality gadget that applies to every provable type, using a `forM` loop that is equivalent to
```lean
for xi in to_elements x, yi in to_elements y do
  assert_zero (xi - yi)
```
* make sure that `circuit_norm` cleanly unfolds the gadget to an equality statement without further user intervention
* simplify fib32 proofs by using the new equality gadget; remove the old, type-specific equality gadgets


Clean/Circuit/Theorems.lean
```diff
@@ -195,4 +195,28 @@ theorem can_replace_completeness  {n: ℕ} {ops : Operations F n} {env} : env.us
     · use ih h.left
       exact circuit.implied_by_completeness env (env_extends_subcircuit_inner h_env) h.right
 
+/--
+Generic version of `constraints_hold`, to reason about soundness and completeness at the same time
+-/
+def constraints_hold.generic (from_subcircuit : {n : ℕ} → Environment F → SubCircuit F n → Prop)
```
gio54321, line 201: nit: I feel like this could be better named as something like `Operations.propFold` instead of `generic`

gio54321 (approved):
LGTM!
I don't really have any substantial comments, this is great!

PR #72, opened 2025-04-14
Infer constant lawful circuit

mitschabaude:
Small follow-up to #71

Adds a tactic to automatically infer `ConstantLawfulCircuits`, which could be useful for loops where this stronger property is needed

### DIFF
--- a/Clean/Circuit/Lawful.lean
+++ b/Clean/Circuit/Lawful.lean
@@ -185,38 +185,6 @@ example :
   ConstantLawfulCircuits add := by infer_constant_lawful_circuits
 end
 
--- constant version of `bind`
-open LawfulCircuit in
-instance ConstantLawfulCircuit.from_bind {f: Circuit F α} {g : α → Circuit F β}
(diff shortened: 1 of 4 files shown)


Clean/Circuit/Lawful.lean
```diff
@@ -144,41 +144,51 @@ example :
   LawfulCircuit add := by infer_lawful_circuit
 end
 
--- constant version of `bind`
+-- `ConstantLawfulCircuit(s)` can be proved from `LawfulCircuit` by adding the requirement that `final_offset` is `n` plus a constant.
+-- the latter can usually be proved by rfl!
 open LawfulCircuit in
-instance ConstantLawfulCircuit.from_bind {f: Circuit F α} {g : α → Circuit F β}
-    (f_lawful : ConstantLawfulCircuit f) (g_lawful : ConstantLawfulCircuits g) : ConstantLawfulCircuit (f >>= g) where
```
mitschabaude, line 150: I removed `ConstantLawfulCircuit.from_bind` since I don't think it is useful in practice. The more straightforward way to achieve the same is to use first use `LawfulCircuit.from_bind` and then prove the "constant local length" property afterwards. This is what the new tactic does.

gio54321 (approved):
LGTM

PR #75, opened 2025-04-16
Byte decomposition proofs

gio54321:
- Fix byte decomposition gadget
- Provide soundness and completeness proofs


Clean/Gadgets/ByteDecomposition.lean
```diff
@@ -4,7 +4,12 @@ import Clean.Utils.Field
 
 namespace Gadgets.ByteDecomposition
 variable {p : ℕ} [Fact p.Prime]
-variable [p_large_enough: Fact (p > 512)]
+variable [p_large_enough: Fact (p > 2^16 + 2^8)]
+
+-- TODO: is there a better way?
+instance : Fact (p > 512) := by
```
gio54321, line 10: We should find a better method for deriving a lower bound on the prime from a stronger bound.

mitschabaude (approved):

PR #77, opened 2025-04-18
Abstract away subcircuit witnesses

mitschabaude:
This PR finishes a project that has been core to our library from the beginning: **"Intransparency" of subcircuits**.

In other words, when doing a proof about circuit A, we should not have to reason about what exactly is going on in A's subcircuits.

The final missing piece in this project was the `Environment.uses_local_witnesses` statement. So far, this statement was directly referencing the _entire_ vector of witnesses created inside a circuit.

This PR instead defines `uses_local_witnesses` inductively, and for subcircuits replaces it with a custom statement, in the same way we already do for `constraints_hold`.

* for **FormalCircuit**, we use the statement `assumptions -> spec`.
  * Note that this complements the `constraint_holds.completeness` statement, which is just `assumptions`. In the proof of a parent circuit, after you have proven a subcircuit's `assumptions`, you can use the `assumptions -> spec` implication to also use the `spec` in the rest of your proof.
* for **FormalAssertion**, we just ignore the statement and replace it with `True`
  * In the completeness proof for assertions, you already have to prove `assumptions AND spec`, and there is no way to strengthen that even further.

With that change, completeness proofs become much easier. To demonstrate this, we add a completeness proof for the Keccak `theta` gadget which is just a single `simp_all` call.


Clean/Circuit/Justification.lean
gio54321, line 1: I would call this file and its content something like `Foundations` instead of `Justification`

gio54321 (approved):
LGTM, much needed!

Clean/Circuit/Justification.lean
mitschabaude: Ok :D

PR #83, opened 2025-05-08
Loop simplification

### LINKED ISSUE #82: Loop circuit constructs that simplify under `circuit_norm`
Right now when proving something about loops we need to rw using a specific theorem and provide an instance of `ConstaintLawfulCircuits`. Both of this is hard to discover. Instead, we want `simp only [circuit_norm]` to do the work.

Getting loops to simplify under `circuit_norm` is not possible using lemmas that need a type class instance, because `simp` won't come up with that instance and so won't apply the lemma.

The solution is to already derive the necessary type class instance _when adding the loop as part of the circuit_. So we need something like `Circuit.mapM` that already requires the instance as argument -- so that we can have lemmas about this construct that `simp` can apply.

mitschabaude:
closes #82 

* add loop construct `Circuit.mapFinRange`, i.e. a `for i=0...n` loop usable in a circuit
* make `mapFinRange` simplify smoothly with `circuit_norm`
* rewrite Chi, ThetaC and ThetaXor to use it

### DIFF
--- a/Clean/Circuit/Basic.lean
+++ b/Clean/Circuit/Basic.lean
@@ -594,7 +594,7 @@ attribute [circuit_norm] Vector.map_mk List.map_toArray List.map_cons List.map_n
 attribute [circuit_norm] Vector.append_singleton Vector.mk_append_mk Vector.push_mk
   Array.append_singleton Array.append_empty List.push_toArray
   List.nil_append List.cons_append List.append_toArray
-  Vector.mapFinRange Vector.mapRange Vector.toArray_push Array.push_toList List.append_assoc
+  Vector.mapRange_zero Vector.mapRange_succ Vector.toArray_push Array.push_toList List.append_assoc
   Vector.eq_mk Vector.mk_eq
 
 -- simplify `vector.get 0` (which occurs in ProvableType definitions)
--- a/Clean/Circuit/Basic.lean
+++ b/Clean/Circuit/Basic.lean
@@ -597,8 +597,14 @@ attribute [circuit_norm] Vector.append_singleton Vector.mk_append_mk Vector.push
   Vector.mapRange_zero Vector.mapRange_succ Vector.toArray_push Array.push_toList List.append_assoc
   Vector.eq_mk Vector.mk_eq
 
--- simplify `vector.get 0` (which occurs in ProvableType definitions)
--- TODO handle other small indices as well
+-- simplify Vector.mapFinRange
+attribute [circuit_norm] Vector.mapFinRange_succ Vector.mapFinRange_zero
+    Nat.cast_zero Nat.cast_one Nat.cast_ofNat Fin.coe_eq_castSucc Fin.reduceCastSucc
+
+-- simplify stuff like (3 : Fin 8).val = 3 % 8
+attribute [circuit_norm] Fin.coe_ofNat_eq_mod
+
+-- simplify `vector.get i` (which occurs in ProvableType definitions) and similar
 attribute [circuit_norm] Vector.get Fin.val_eq_zero List.getElem_toArray
   List.getElem_cons_zero Fin.cast_eq_self List.getElem_cons_succ
   Vector.getElem_mk Vector.getElem_toArray Vector.getElem_map Fin.getElem_fin
--- a/Clean/Circuit/Basic.lean
+++ b/Clean/Circuit/Basic.lean
@@ -556,14 +556,14 @@ def FlatOperation.witness_generators : (l: List (FlatOperation F)) → Vector (E
     let ws := witness_generators ops
     match op with
     | witness m compute =>
-      ⟨ (Vector.mapFinRange (fun i env => (compute env).get i)).toArray ++ ws.toArray, by
+      ⟨ (Vector.mapFinRange m (fun i env => (compute env).get i)).toArray ++ ws.toArray, by
         simp only [Array.size_append, Vector.size_toArray, witness_length]; ac_rfl⟩
     | assert _ | lookup _ =>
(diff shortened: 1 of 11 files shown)


Clean/Gadgets/Addition32/Addition32Full.lean
mitschabaude, line 1: simplifications here are unrelated to the PR

Clean/Circuit/Loops.lean
mitschabaude, line 1: additions here are the main content of the PR

Clean/Gadgets/Keccak/Chi.lean
mitschabaude, line 1: soundness proof of chi is now very natural and doesn't require special knowledge of our loop theorems. it also no longer requires an explicit `LawfulCircuit` instantiation

Clean/Utils/Vector.lean
```diff
@@ -140,18 +140,39 @@ theorem induct_push_push {motive : {n: ℕ} → Vector α n → Sort u}
...
+
+theorem getElem_finRange {n} (i : ℕ) (hi : i < n) : (finRange n)[i] = ⟨ i, hi ⟩ := by
+  simp [finRange, List.finRange]
+
+theorem finRange_succ {n} : finRange (n + 1) = ((finRange n).map Fin.castSucc).push n := by
+  ext i hi
+  simp only [getElem_finRange, getElem_push, getElem_map, Fin.castSucc_mk]
+  by_cases hi' : i < n <;> simp [hi']; linarith
+
+def mapFinRange (n: ℕ) (create: Fin n → α) : Vector α n := finRange n |>.map create
```
mitschabaude, line 156: `Vector.mapFinRange` is changed to take the length argument explicitly, for consistency with similar methods

Clean/Circuit/Basic.lean
```diff
@@ -594,11 +594,17 @@ attribute [circuit_norm] Vector.map_mk List.map_toArray List.map_cons List.map_n
...
   Vector.eq_mk Vector.mk_eq
 
--- simplify `vector.get 0` (which occurs in ProvableType definitions)
--- TODO handle other small indices as well
+-- simplify Vector.mapFinRange
+attribute [circuit_norm] Vector.mapFinRange_succ Vector.mapFinRange_zero
+    Nat.cast_zero Nat.cast_one Nat.cast_ofNat Fin.coe_eq_castSucc Fin.reduceCastSucc
+
+-- simplify stuff like (3 : Fin 8).val = 3 % 8
+attribute [circuit_norm] Fin.coe_ofNat_eq_mod
```
mitschabaude, line 605: unrelated to other stuff in this PR, I played with `circuit_norm` attributes a bit to improve simplification of commonly occurring expressions

Clean/Circuit/Loops.lean
```diff
@@ -346,5 +346,59 @@ theorem mapFinRangeM_completeness {n : ℕ} :
   simp only [completeness_iff_generic, mapFinRangeM_generic]
 end
 end constraints_hold
+
+-- Loop constructs designed to simplify under `circuit_norm`
+
+def mapFinRange (m : ℕ) [NeZero m] [Nonempty β] (body : Fin m → Circuit F β)
+    (_lawful : ConstantLawfulCircuits body := by infer_constant_lawful_circuits) : Circuit F (Vector β m) :=
```
gio54321, line 353: Nice, this is fine as long as `infer_constant_lawful_circuits` works well enough. I'm curious about strange situations like nested loops, in which this tactic might not work. In that case, we can just provide an instance alongside the loop's definition.
But the tactic works well in practice, so it should not be needed.

Clean/Gadgets/Keccak/Chi.lean
```diff
@@ -24,75 +24,44 @@ def spec (state : KeccakState (F p)) (out_state : KeccakState (F p)) :=
...
-    rw [←h_input, Not.eval_not]
+  -- simplify constraints
+  simp only [circuit_norm, eval_vector, Vector.ext_iff] at h_input
+  simp only [assumptions, KeccakState.is_normalized, Fin.getElem_fin] at state_norm
+  simp only [main, circuit_norm, subcircuit_norm, Xor.circuit, And.And64.circuit, Not.circuit,
+    Xor.assumptions, Xor.spec, And.And64.assumptions, And.And64.spec, Nat.reduceAdd] at h_holds
 
-  have ⟨ state_not_value, state_not_norm ⟩ := Not.not_bytewise_value_spec (state_norm (i + 5))
-  simp_all [KeccakState.value]
+  simp_all
```
gio54321, line 58: Very nice!

Clean/Gadgets/Keccak/ThetaC.lean
```diff
@@ -46,66 +27,68 @@ def spec (state : KeccakState (F p)) (out: KeccakRow (F p)) :=
...
+  intro i
+  specialize h_holds i
+  obtain ⟨ h0, h1, h2, h3 ⟩ := h_holds
+  obtain ⟨ xor0, norm0 ⟩ := h0
+  have ⟨ xor1, norm1 ⟩ := h1 norm0
+  have ⟨ xor2, norm2 ⟩ := h2 norm1
+  have ⟨ xor4, norm ⟩ := h3 norm2
+  clear h1 h2 h3 norm0 norm1 norm2
+  use norm
+  simp only [xor0, xor1, xor2, xor4, KeccakState.value, Vector.getElem_map]
```
gio54321: Nit simplification
```suggestion
  simp only [KeccakState.value, Vector.getElem_map]
  intro i
  specialize h_holds i
  aesop
```

gio54321 (approved):
LGTM

PR #84, opened 2025-05-09
Support exporting air table to json

katat:
- [x] figure out how to serialize the [expression constant value](https://github.com/Verified-zkEVM/clean/compare/serialize-tables...json?expand=1#diff-36e9fb1219ad44dc1bd50d8183dbf26fcf47d1595ca15442144a07b370150470R24-R27)

### DIFF
--- a/Clean/Circuit/Basic.lean
+++ b/Clean/Circuit/Basic.lean
@@ -2,8 +2,6 @@ import Clean.Circuit.Expression
 import Clean.Circuit.Provable
 import Clean.Circuit.SimpGadget
 
-open Lean
-
 variable {F: Type} [Field F]
 
 structure Table (F : Type) where
@@ -533,7 +531,6 @@ instance [Repr F] : Repr (Operation F) where
     | assert e => "(Assert " ++ reprStr e ++ " == 0)"
     | lookup l => reprStr l
     | subcircuit { ops, .. } => "(SubCircuit " ++ reprStr ops ++ ")"
-
 end Operation
 
 def Operations.toList {n: ℕ} : Operations F n → List (Operation F)
--- a/Clean/Circuit/Basic.lean
+++ b/Clean/Circuit/Basic.lean
@@ -2,6 +2,8 @@ import Clean.Circuit.Expression
 import Clean.Circuit.Provable
 import Clean.Circuit.SimpGadget
 
+open Lean
+
 variable {F: Type} [Field F]
 
 structure Table (F : Type) where
@@ -19,6 +21,12 @@ structure Lookup (F : Type) where
 
 instance [Repr F] : Repr (Lookup F) where
   reprPrec l _ := "(Lookup " ++ l.table.name ++ " " ++ repr l.entry ++ ")"
+
+instance : ToJson (Lookup F) where
+  toJson l := Json.mkObj [
+    ("table", toJson l.table.name),
+    ("entry", toJson l.entry.toArray),
+  ]
 
 variable {α : Type} {n : ℕ}
 
@@ -46,6 +54,12 @@ def toString [Repr F] : FlatOperation F → String
 
 instance [Repr F] : Repr (FlatOperation F) where
   reprPrec op _ := toString op
+
+instance : ToJson (FlatOperation F) where
+  toJson op := match op with
+    | witness m _ => Json.mkObj [("witness", toJson m)]
+    | assert e => Json.mkObj [("assert", toJson e)]
+    | lookup l => Json.mkObj [("lookup", toJson l)]
 
 /--
 What it means that "constraints hold" on a list of flat operations:
@@ -531,6 +545,14 @@ instance [Repr F] : Repr (Operation F) where
     | assert e => "(Assert " ++ reprStr e ++ " == 0)"
     | lookup l => reprStr l
     | subcircuit { ops, .. } => "(SubCircuit " ++ reprStr ops ++ ")"
+
+instance : ToJson (Operation F) where
+  toJson op := match op with
+    | witness m _ => Json.mkObj [("witness", toJson m)]
+    | assert e => Json.mkObj [("assert", toJson e)]
+    | lookup l => Json.mkObj [("lookup", toJson l)]
+    | subcircuit { ops, .. } => Json.mkObj [("subcircuit", toJson ops)]
+
 end Operation
 
 def Operations.toList {n: ℕ} : Operations F n → List (Operation F)
@@ -548,6 +570,9 @@ def Circuit.operation_list (circuit: Circuit F α) (offset := 0) : List (Operati
 
 instance [Repr F] : Repr (OperationsList F) where
   reprPrec ops _ := reprStr ops.toList
+
+instance : ToJson (OperationsList F) where
+  toJson ops := ToJson.toJson ops.toList
 
 -- witness generation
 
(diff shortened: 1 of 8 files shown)


Clean/Circuit/Expression.lean
```diff
@@ -15,6 +15,30 @@ inductive Expression (F : Type) where
   | add : Expression F -> Expression F -> Expression F
   | mul : Expression F -> Expression F -> Expression F
 
+
+partial def exprToJson : Expression F → Lean.Json
```
mitschabaude: why is this a `partial def`?

Clean/Circuit/Expression.lean
```diff
@@ -15,6 +15,30 @@ inductive Expression (F : Type) where
...
+  | .var v   => Lean.Json.mkObj [
+                  ("type", Lean.Json.str "var"),
+                  ("index",     Lean.Json.num v.index)   -- Nat’s instance
+                ]
+  | .const c => Lean.Json.mkObj [
+                  ("type", Lean.Json.str "const"),
+                  ("value", "cst")   -- F’s instance
```
mitschabaude: I think the correct way to do this is assume `[Lean.ToJson F]` as a parameter to this function, and then just serialize the constant `c` using `Lean.ToJson.toJson c`

Clean/Circuit/Expression.lean
```diff
@@ -15,6 +15,30 @@ inductive Expression (F : Type) where
   | add : Expression F -> Expression F -> Expression F
   | mul : Expression F -> Expression F -> Expression F
 
+
+partial def exprToJson : Expression F → Lean.Json
+  | .var v   => Lean.Json.mkObj [
+                  ("type", Lean.Json.str "var"),
+                  ("index",     Lean.Json.num v.index)   -- Nat’s instance
+                ]
```
mitschabaude: can you consistently use an indentation of 2 spaces?
also, there should be just one space in between symbols.

here for example, I would expect:
```suggestion
  | .var v => Lean.Json.mkObj [
    ("type", Lean.Json.str "var"),
    ("index", Lean.Json.num v.index) -- Nat’s instance
  ]
```

Clean/Table/Basic.lean
```diff
@@ -402,6 +434,22 @@ instance [Repr F] : Repr (TableOperation S F) where
     | .EveryRow c => "EveryRow " ++ reprStr c
     | .EveryRowExceptLast c => "EveryRowExceptLast " ++ reprStr c
 
+instance : ToJson (TableOperation S F) where
+  toJson op := match op with
+    | .Boundary i c => Json.mkObj [
+        ("Boundary", Json.mkObj [
+          ("row", toJson i),
+          ("context", toJson c)
+        ])
```
mitschabaude: I think using the type of table operation (e.g. "Boundary") as a key is not optimal, usually enums are easiest to deal with by introducing a `"type"` key for them.

```suggestion
    | .Boundary i c => Json.mkObj [
      ("type", Json.str "Boundary"),
      ("row", toJson i),
      ("context", toJson c)
    ]
```

Clean/Table/Basic.lean
```diff
@@ -402,6 +434,22 @@ instance [Repr F] : Repr (TableOperation S F) where
...
+      ]
+    | .EveryRow c => Json.mkObj [
+        ("EveryRow", toJson c)
+      ]
+    | .EveryRowExceptLast c => Json.mkObj [
+        ("EveryRowExceptLast", toJson c)
+      ]
```
mitschabaude: see [my suggestion](https://github.com/Verified-zkEVM/clean/pull/84#discussion_r2081425721) for `Boundary`

Clean/Table/Basic.lean
```diff
@@ -402,6 +434,22 @@ instance [Repr F] : Repr (TableOperation S F) where
     | .EveryRow c => "EveryRow " ++ reprStr c
     | .EveryRowExceptLast c => "EveryRowExceptLast " ++ reprStr c
 
+instance : ToJson (TableOperation S F) where
+  toJson op := match op with
+    | .Boundary i c => Json.mkObj [
```
mitschabaude: nit: the idiomatic way to define functions whose arguments are immediately matched on is this:
```suggestion
  toJson
    | .Boundary i c => Json.mkObj [
```

Clean/Table/Basic.lean
```diff
@@ -156,6 +164,13 @@ instance : Repr (Cell W S) where
     | .input off => ".input " ++ reprStr off
     | .aux i => ".aux " ++ reprStr i
 
+instance : ToJson (Cell W S) where
+  toJson cell := match cell with
+    | .input off => toJson off
```
mitschabaude: nit: the idiomatic way to define functions whose arguments are immediately matched on is this:
```suggestion
  toJson
    | .input off => toJson off
```

Clean/Circuit/Basic.lean
mitschabaude, line 1: could you move all the `ToJson` instances here and in `Expression.lean` into a new file `Circuit/Json.lean`?

Clean/Table/Basic.lean
mitschabaude, line 1: Could you move all the `ToJson` instances here to a new file `Table/Json.lean`?

Clean/Table/Basic.lean
```diff
@@ -156,6 +164,13 @@ instance : Repr (Cell W S) where
...
+    | .aux i => Json.mkObj [
+        ("aux", toJson i)
+      ]
```
mitschabaude: nit: indentation is unexpected, here you could do
```suggestion
    | .aux i => Json.mkObj [("aux", toJson i)]
```

mitschabaude (commented):
left a few comments!

Clean/Circuit/Expression.lean
katat: changed to def :D

Clean/Circuit/Expression.lean
katat: Got it. Fixed

Clean/Circuit/Expression.lean
katat: maybe we should have a standard for auto formatting?

Clean/Table/Basic.lean
katat: fixed

Clean/Circuit/Expression.lean
mitschabaude: that would be nice, I don't think an established standard exists for Lean yet. I remember this was in the priorities for Lean language development this year

Clean/Table/Json.lean
```diff
@@ -0,0 +1,81 @@
...
+
+instance : ToJson (TableOperation S F) where
+  toJson
+    | .Boundary i c => Json.mkObj [
+      ("type", Json.str "Boundary"),
+      ("row", toJson i),
+      ("context", toJson c)
+    ]
+    | .EveryRow c => Json.mkObj [("EveryRow", toJson c)]
+    | .EveryRowExceptLast c => Json.mkObj [("EveryRowExceptLast", toJson c)]
```
mitschabaude: not sure if you noticed but you have this twice now, the one below is better!

katat: oops. removed the duplicated.

katat:
thanks for the review!

PR #85, opened 2025-05-09
Miscellaneous simplifications

mitschabaude:
- **minor changes to keccak spec**
- **introduce separate definition for circuit.local_length**
- **remove unnecessary openings and namespacing**
- **simplify some loop proofs**


marcobesier (approved):
Looking good to me 🙌

PR #89, opened 2025-05-12
KeccakRound

mitschabaude:
- **change round constant type to be UInt64 so that we don't have to prove the size individially**
- **keccak round gadget + elaboration**
- **lemma about decomposing and summing U64**
- **keccak round soundness**

### DIFF
--- a/Clean/Specs/Keccak256.lean
+++ b/Clean/Specs/Keccak256.lean
@@ -5,7 +5,7 @@ namespace Specs.Keccak256
 
 open Bitwise (not64 rot_left64)
 
-def roundConstants : Vector ℕ 24 := #v[
+def roundConstants : Vector UInt64 24 := #v[
   0x0000000000000001, 0x0000000000008082,
   0x800000000000808a, 0x8000000080008000,
   0x000000000000808b, 0x0000000080000001,
@@ -134,11 +134,11 @@ def chi (b : Vector ℕ 25) : Vector ℕ 25 :=
     b[24] ^^^ ((not64 b[4]) &&& b[9])
   ]
 
-def iota (state : Vector ℕ 25) (rc : ℕ) : Vector ℕ 25 :=
-  state.set 0 ((state.get 0) ^^^ rc)
+def iota (state : Vector ℕ 25) (rc : UInt64) : Vector ℕ 25 :=
+  state.set 0 ((state.get 0) ^^^ rc.val)
 
 
-def keccak_round (state : Vector ℕ 25) (rc : ℕ) : Vector ℕ 25 :=
+def keccak_round (state : Vector ℕ 25) (rc : UInt64) : Vector ℕ 25 :=
   let theta_state := theta state
   let rho_pi_state := rho_pi theta_state
   let chi_state := chi rho_pi_state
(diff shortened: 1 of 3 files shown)


Clean/Circuit/Basic.lean
```diff
@@ -284,6 +284,12 @@ def lookup (l: Lookup F) : Circuit F Unit :=
 
 end Circuit
 
+@[circuit_norm]
+def ProvableType.witness {α: TypeMap} [ProvableType α] (compute : Environment F → α F) : Circuit F (α (Expression F)) :=
+  modifyGet fun ops =>
+    let var := var_from_offset α ops.offset
+    ⟨var, .witness ops (size α) (fun env => compute env |> to_elements)⟩
```
mitschabaude, line 291: I moved this here from `Extensions`, I think it's generally useful, and better if defined using `modifyGet` and returns `var_from_offset`, since we already have good simp lemmas about that

Clean/Circuit/Basic.lean
```diff
@@ -613,21 +619,32 @@ attribute [circuit_norm] Vector.map_mk List.map_toArray List.map_cons List.map_n
...
+  Vector.getElem_map Vector.getElem_mapFinRange Vector.getElem_mapRange Vector.getElem_finRange
+  Vector.getElem_push Vector.getElem_set Vector.getElem_cast
+  Vector.getElem_mk Vector.getElem_toArray
+  List.getElem_cons_zero List.getElem_cons_succ List.getElem_toArray
+
+/-
+lemmas that would expand `Vector.{mapRange, mapFinRange}` are not added to the simp set,
+because they would sometimes be applied too eagerly where using the corresponding `getElem` lemma is much better
+-/
+-- attribute [circuit_norm] Vector.mapRange_zero Vector.mapRange_succ Vector.mapFinRange_succ Vector.mapFinRange_zero
```
mitschabaude, line 636: these lemmas led to some really annoying terms in the keccak round proof.
I asked about it on Zulip: [#general > ✔ how to tweak simp set to not unfold the &#96;list&#96; in &#96;list...](https://leanprover.zulipchat.com/#narrow/channel/113488-general/topic/.E2.9C.94.20how.20to.20tweak.20simp.20set.20to.20not.20unfold.20the.20.60list.60.20in.20.60list.2E.2E.2E/with/517759673)

but the conclusion is that these annoying simplifications can't be avoided. I think it's better err on the side of simping too little than too much, so I removed them from the set

Clean/Specs/Keccak256.lean
```diff
@@ -5,7 +5,7 @@ namespace Specs.Keccak256
 
 open Bitwise (not64 rot_left64)
 
-def roundConstants : Vector ℕ 24 := #v[
+def roundConstants : Vector UInt64 24 := #v[
```
mitschabaude, line 8: I changed the round constants type to `UInt64`, because the size bound is needed in proofs, and it's easier to have it in the type than to prove every time

marcobesier: Good idea, makes a lot of sense 👍

marcobesier (approved):
Looking good to me 👍

PR #91, opened 2025-05-18
Witness generation for table

katat:
The implementation assumption are:
- the `EveryRowExceptLast` table constraint contains all the witness generators for values in the next trace row.
- given a row, these witness generators can yields values for the next trace row.
- by using the `CellAssignment.vars`, the generated witnesses can be mapped to the next row, including the auxiliary columns.


Clean/Table/Basic.lean
```diff
@@ -404,6 +414,110 @@ instance [Repr F] : Repr (TableOperation S F) where
...
+  let aux_map := build_aux_map assignment
+  let next_row := Array.mkArray cur_row.size 0
+
+  -- rules for fetching the values for expression variables
+  let env i :=
+    if h : i < assignment.offset then
+      match assignment.vars.get ⟨i, h⟩ with
+      | .input ⟨r, c⟩ =>
+        -- fetch input values
+          if r = 0 then cur_row.get! c else next_row.get! c
```
mitschabaude: initially I thought accessing `next_row` here doesn't work as intended, because `next_row.set!` below does not mutate `next_row`. Lean is a purely functional programming language, so in most places you wouldn't expect that it does mutation.

I thought it will always access the _initial_ `next_row` of all zeros.

however, it seems it works because, as the docstring of `Array.set!` says:
> This will perform the update destructively provided that `a` has a reference count of 1 when called.

so, nothing to be done here!

Clean/Table/Basic.lean
```diff
@@ -404,6 +414,110 @@ instance [Repr F] : Repr (TableOperation S F) where
...
+  Mapping for auxiliary columns:
+  - The auxiliary cells' mapping from the `CellAssignment` to the aux columns in the trace row
+  is done using the `build_aux_map` function.
+  - The generated witnesses are assigned to the corresponding aux columns in next trace row.
+
+  Mapping for input columns:
+  - According to `CellAssignment` for input cells, the input columns are assigned to
+  the corresponding columns in the trace row.
+-/
+def generate_next_row [Lean.ToJson F] (tc : TableConstraint W S F Unit) (cur_row: Array F) : Array F :=
```
mitschabaude: you don't need `[Lean.ToJson F]` here

Clean/Table/Basic.lean
```diff
@@ -404,6 +414,110 @@ instance [Repr F] : Repr (TableOperation S F) where
...
+
+  next_row
+
+/--
+  Generates a trace of length n, starting from the given row.
+
+  Returns an array of rows where each subsequent row is generated using the
+  table constraint's witness generators.
+-/
+def witnesses [Lean.ToJson F]
```
mitschabaude: you don't need `[Lean.ToJson F]` here

Clean/Table/Basic.lean
```diff
@@ -404,6 +414,110 @@ instance [Repr F] : Repr (TableOperation S F) where
...
+    -- initialize the trace with the first row
+    let trace := #[cur_row]
+
+    -- generate the remaining n-1 rows iteratively
+    let rec generate_rows (trace : Array (Array F)) (current : Array F) (remaining : Nat) : Array (Array F) :=
+      if remaining = 0 then
+        trace
+      else
+        let next := generate_next_row tc current
+        generate_rows (trace.push next) next (remaining - 1)
```
mitschabaude: the fact that this loop is written as a recursive function probably leads to the stack overflow for large input sizes. (weird, because I would have expected the compiler to do tail call optimization)

you don't need to change it in this PR but eventually we will have to write it with `do` notation instead

Clean/Table/Basic.lean
mitschabaude, line 1: can you move the witness generation logic in this file to a new file called `WitnessGeneration.lean`?

It's not general enough yet to be part of the official library API. Right now, it's specifically only designed for `EveryRowExceptLast` type constraints with a two-row window

mitschabaude (approved):
nice work! approving preemptively, but have a few comments to be addressed

Clean/Table/Basic.lean
katat: moved

Clean/Table/Basic.lean
katat: I updated it to use `do`. Indeed, the stack overflow issue resolved. 
Using `#eval` would still introduce that issue, but streaming the calculation result to a file is fine.

PR #94, opened 2025-05-23
Update toolchain and mathlib to 4.19.0

gio54321:
Updates Lean toolchain and mathlib to 4.19.0 stable
For future reference: it is easier to update incrementally, getting a building state at each stable version.


Clean/Utils/Vector.lean
```diff
@@ -137,30 +150,17 @@ theorem induct_push_push {motive : {n: ℕ} → Vector α n → Sort u}
...
-  simp [finRange, List.finRange]
-
-theorem getElem_finRange {n} (i : ℕ) (hi : i < n) : (finRange n)[i] = ⟨ i, hi ⟩ := by
-  simp [finRange, List.finRange]
-
-theorem finRange_succ {n} : finRange (n + 1) = ((finRange n).map Fin.castSucc).push n := by
-  ext i hi
-  simp only [getElem_finRange, getElem_push, getElem_map, Fin.castSucc_mk]
-  by_cases hi' : i < n <;> simp [hi']; linarith
+theorem getElemFin_finRange {n} (i : Fin n) : (Vector.finRange n)[i] = i := by
```
mitschabaude: nit: I think we don't need the `Vector.` namespace in this file since we're in it?

mitschabaude (approved):
amazing!

Clean/Utils/Vector.lean
gio54321: Not sure, it's now using `finRange` from mathlib, so without the `Vector.` it does not find it.
I will add an `open Vector (finRange)` at the top

PR #96, opened 2025-05-23
forEach loop

mitschabaude:
This finishes a line of work on loops in clean, by adding `Circuit.forEach` which is just `Vector.forM` but simplifies well under `circuit_norm`.

* Refactor the generic equality gadget to use `Circuit.forEach`, mainly to test that everything works as expected
* Remove several theorems in `Loops.lean` that aren't needed for our collection of `Circuit` loops


marcobesier (approved):
No comments. Looking good 👍

PR #97, opened 2025-05-27
Num2Bits gadget and generalization of loop theorems

mitschabaude:
- **(potentially) auxiliary file with to/from bits methods and consistency theorems**
- **add witness_vector**
- **missing theorems and better auto tactic in Provable**
- **add to bits gadget, it doesn't neatly simplify**
- **generalize forEach lemmas. we'll have to do this everywhere**
- **progress on output**
- **problem: offset doesn't simplify**
- **this solves output simp**
- **more general foreach soundness**
- **good progress on to_bits soundness**
- **full soundness proof**
- **general forEach completeness theorems**
- **extract most of toBits proof into lemmas**
- **finish to_bits theorems**
- **add range check version to tobits**
- **simplify proofs and add comments**
- **get fairly close to a sorry-free proof**

### DIFF
--- a/Clean/Gadgets/Bits.lean
+++ b/Clean/Gadgets/Bits.lean
@@ -41,9 +41,8 @@ def circuit (n : ℕ) : FormalCircuit (F p) field (BitVector n) where
   local_length_eq _ _ := by simp only [main, circuit_norm, Boolean.circuit]; ac_rfl
 
   output_eq _ _ := by
-    simp only [main, circuit_norm, Boolean.circuit]
-    sorry
-
+    simp only [main, circuit_norm, Boolean.circuit, var_from_offset_vector]
+    rfl
 
   assumptions (x : F p) := x.val < 2^n
 
--- a/Clean/Gadgets/Bits.lean
+++ b/Clean/Gadgets/Bits.lean
@@ -149,7 +149,7 @@ def circuit (n : ℕ) (hn : 2^n < p) : FormalCircuit (F p) field (BitVector n) w
   initial_offset_eq _ n := by simp only [main, circuit_norm]
   local_length_eq _ _ := by simp only [main, circuit_norm, Boolean.circuit]; ac_rfl
   output_eq _ _ := by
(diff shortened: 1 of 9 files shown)


Clean/Gadgets/Bits.lean
```diff
@@ -0,0 +1,235 @@
...
+variable {p : ℕ} [prime: Fact p.Prime] [p_large_enough: Fact (p > 2)]
+
+def to_bits (n : ℕ) (x : F p) : Vector (F p) n :=
+  .mapRange n fun i => if x.val.testBit i then 1 else 0
+
+def from_bits {n : ℕ} (bits : Vector (F p) n) : F p :=
+  Fin.foldl n (fun acc ⟨i, _⟩ => acc + bits[i] * 2^i) 0
+
+def from_bits_expr {n: ℕ} (bits : Vector (Expression (F p)) n) : Expression (F p) :=
+  Fin.foldl n (fun acc ⟨i, _⟩ => acc + bits[i] * (2^i : F p)) 0
```
gio54321: Probably it is better to move these definitions and theorems into their separate file under `Utils/ ` or in a separate `Theorems` file.
Same thing for the theorems below about `to_bits` and `from_bits`.

Also worth mentioning that `BitVec.ofBoolListLE` exists, so probably this iso can be extended also to `BitVec`, which should simplify a lot other proofs!

Clean/Gadgets/Bits.lean
```diff
@@ -0,0 +1,235 @@
...
+  spec (x : F p) (bits : Vector (F p) n) :=
+    bits = to_bits n x
+
+  soundness := by
+    intro k eval x_var x h_input h_assumptions h_holds
+    dsimp only [main] at *
+    simp only [main, circuit_norm, Boolean.circuit, true_and] at *
+    -- TODO: simp [circuit_norm] is not able to exclude the case that `forEach` results in empty operations
+    -- which leads to this case split and hard-to-discover proof that dicharges the empty case
+    split at h_holds
```
gio54321, line 44: I see, this was quite annoying

Clean/Gadgets/Bits.lean
mitschabaude: done!

PR #98, opened 2025-05-27
Remove conditionals in `constraints_hold` definition

mitschabaude:
This PR simplifies the definitions of `constraints_holds`, which had a condition to avoid an extra leading `True` in the statement. It turns out that the condition can be hard to simplify, while the leading `True` is trivial to get rid of by adding `and_true` to the simp set.

* add `and_true` and a few similar lemmas to `circuit_norm`
* remove `and_true` & co from simp invocations where it's no longer needed
* simplify the `Bits` gadget proof
* remove unproved theorem that previously was needed for `Bits`


gio54321 (approved):
LGTM

PR #99, opened 2025-05-28
Add BLAKE3 blueprint

marcobesier:
This PR adds a blueprint of the core components of the BLAKE3 hash function. It includes:

- Core BLAKE3 constants and parameters (output length, key length, block length, chunk length)
- Flag definitions for the compression function (chunk start/end, parent, root, etc.)
- Initialization vectors (IV)
- Message permutation schedule
- Core cryptographic functions:
  - Mixing function G
  - Round function
  - Message permutation
  - Compression function

The implementation follows the [BLAKE3 paper](https://github.com/BLAKE3-team/BLAKE3-specs/blob/master/blake3.pdf) and the reference implementations in [Rust](https://github.com/BLAKE3-team/BLAKE3/blob/master/reference_impl/reference_impl.rs) and [Python](https://github.com/oconnor663/pure_python_blake3).

The PR further includes some minimalistic test cases verified against the reference Python implementation. 

The purpose of this blueprint is to serve as a reference specification for implementing BLAKE3 gadgets in `clean`.


gio54321 (approved):
LGTM!

PR #100, opened 2025-05-29
Rebuild Circuit monad on simpler types

mitschabaude:
This drastically simplifies the core types of clean:
* `Operations F n` becomes `List (Operation F)`, a simple list without length dependency
* `Circuit` becomes
```lean
def Circuit (F : Type) [Field F] (α : Type) := ℕ → α × List (Operation F)
```
with the monad bind defined as
```lean
theorem bind_def {α β} (f : Circuit F α) (g : α → Circuit F β) :
  f >>= g = fun n =>
    let (a, ops) := f n
    let (b, ops') := g a (n + Operations.local_length ops)
    (b, ops ++ ops') := rfl
```

In particular, the new definition satisfies the properties previously encoded in `LawfulCircuit`:
* the circuit output only depends on the input offset, not on previous operations
* the circuit operations are just appended to previous operations, not arbitrarily changed
* the new offset is derived from the operations (this used to be the case before as well)

The only thing we lose is "subcircuit consistency by definition". This means that subcircuit consistency is a new property we now have to prove for every formal circuit. This proof just amounts to showing an equality of two offsets for every subcircuit, which is usually trivial.

This PR fully transitions clean to the new types, porting all circuits and proofs without new sorrys.


Clean/Circuit/Basic.lean
```diff
@@ -1,223 +1,17 @@
-import Clean.Circuit.Expression
-import Clean.Circuit.Provable
-import Clean.Circuit.SimpGadget
+import Clean.Circuit.Operations
+import Mathlib.Control.Monad.Writer
 
-variable {F: Type} [Field F]
-
-structure Table (F : Type) where
```
mitschabaude, line 7: code here moved to `Operations.lean`

Clean/Circuit/Basic.lean
```diff
@@ -231,95 +25,99 @@ def circuit : Circuit F Unit := do
   lookup { table := MyTable, entry := [x], ... }
 ```
 -/
-@[reducible]
-def Circuit (F : Type) [Field F] := StateM (OperationsList F)
+def Circuit (F : Type) [Field F] (α : Type) := ℕ → α × List (Operation F)
```
mitschabaude, line 28: this is the change in a nutshell:
* operations are now a plain list type
* a circuit no longer depends on all previous operations, but just on the current offset
* we define a custom monad instead of using off-the-shelf `StateM`
  * this lets us make the circuit append-only by definition

Clean/Circuit/Basic.lean
```diff
@@ -394,12 +217,12 @@ class ElaboratedCircuit (F: Type) [Field F] (β α: TypeMap) [ProvableType β] [
...
     := by intros; rfl
 
-  /--
-    technical condition needed for subcircuit consistency: the circuit must not change its initial offset.
-    usually automatically proved by `rfl`.
-  -/
-  initial_offset_eq: ∀ var, ∀ n, (main var |>.operations n).initial_offset = n
-    := by intros; rfl
+  /-- technical condition: all subcircuits must be consistent with the current offset -/
+  subcircuits_consistent : ∀ var offset, ((main var).operations offset).subcircuits_consistent offset
```
mitschabaude, line 221: the concept of "initial offset" is no longer necessary, but since we simplified the operations type, we need this new "subcircuit consistency" statement on all formal circuits

Clean/Circuit/Basic.lean
```diff
@@ -508,48 +342,6 @@ end
 
 export Circuit (witness_var witness witness_vars witness_vector assert_zero lookup)
 
-/-- move from inductive (nested) operations back to flat operations -/
-def to_flat_operations {n: ℕ} : Operations F n → List (FlatOperation F)
-  | .empty _ => []
```
mitschabaude, line 513: moved to `Operations.lean`

Clean/Circuit/Basic.lean
```diff
@@ -580,34 +372,16 @@ def Circuit.witnesses (circuit: Circuit F α) (offset := 0) : Array F :=
     acc.push (compute ⟨ env ⟩))
   #[]
 
--- generic folding over `Operations` resulting in a proposition
-
-structure Operations.Condition (F: Type) [Field F] where
-  witness (n: ℕ) : (m : ℕ) → (Environment F → Vector F m) → Prop
-  assert (n: ℕ) : Expression F → Prop
```
mitschabaude, line 587: moved to `Operations.lean`

Clean/Circuit/Constant.lean
mitschabaude, line 1: we no longer need `LawfulCircuit`, but for loop bodies we still need the property that the local length is independent of the input offset, i.e. `ConstantCircuit`

Clean/Circuit/Explicit.lean
mitschabaude, line 1: This file and the `ExplicitCircuit` tactic is a port of the automatic derivation of `LawfulCircuit` that we had before. The "lawful" aspect is not needed anymore (all circuits are lawful in the old sense), but I wanted to preserve the automatic derivation, as this is still a compelling alternative to `circuit_norm` simplification. See `Add32Explicit` with examples

Clean/Circuit/Append.lean
mitschabaude, line 1: "append" for circuit operations doesn't need its own theory anymore, it's just `List.append`

Clean/Circuit/Operations.lean
```diff
@@ -0,0 +1,281 @@
...
+  | .subcircuit s :: ops => condition.subcircuit offset s ∧ forAll (s.local_length + offset) condition ops
+
+/--
+Subcircuits start at the same variable offset that the circuit currently is.
+In practice, this is always true since subcircuits are instantiated using `subcircuit` or `assertion`.
+ -/
+ @[circuit_norm]
+def subcircuits_consistent (offset : ℕ) (ops : Operations F) := ops.forAll offset {
+  subcircuit offset {n} _ := n = offset
+}
```
mitschabaude, line 246: new definition of subcircuit consistency

Clean/Circuit/Operations.lean
```diff
@@ -0,0 +1,281 @@
...
+  | .lookup _ => 0
+  | .subcircuit s => s.local_length
+end Operation
+
+/--
+`Operations F` is an alias for `List (Operation F)`, so that we can define
+methods on operations that take a self argument.
+-/
+@[reducible, circuit_norm]
+def Operations (F : Type) [Field F] := List (Operation F)
```
mitschabaude, line 159: operations are no longer type-dependent on the length! this simplifies so much

Clean/Circuit/Loops.lean
mitschabaude, line 1: loops was by far the hardest part to port over, but the end result is equivalent to before, with simpler proofs in many places and even slightly better simplification (see e.g. `Permutation.lean`)

Clean/Circuit/Theorems.lean
mitschabaude, line 1: this file now has a bunch of lemmas that previously needed the `LawfulCircuit` assumption

Clean/Circuit/Theorems.lean
```diff
@@ -8,7 +8,112 @@ import Clean.Circuit.Basic
...
+@[ext]
+theorem ext {f g : Circuit F α}
+  (h_output : ∀ n, f.output n = g.output n)
+  (h_operations : ∀ n, f.operations n = g.operations n) :
+    f = g :=
+  ext_iff.mpr fun n => ⟨ h_output n, h_operations n ⟩
+
+-- lawful monad
+
+instance : LawfulMonad (Circuit F) where
```
mitschabaude, line 94: I had to add this since `LawfulMonad` could no longer be derived automatically

Clean/Examples/Add32LawfulCircuit.lean
mitschabaude, line 1: this file is now `Add32Explicit.lean`

Clean/Gadgets/Addition32/Addition32Full.lean
```diff
@@ -82,9 +84,9 @@ theorem soundness : Soundness (F p) elaborated assumptions spec := by
   obtain ⟨ y0_byte, y1_byte, y2_byte, y3_byte ⟩ := y_norm
 
   -- simplify circuit
-  simp only [circuit_norm, subcircuit_norm, add32_full, add8_full_carry, Boolean.circuit, ByteLookup] at h
-  simp only [Boolean.spec, and_assoc, add_zero, h_inputs] at h
-  rw [ByteTable.equiv, ByteTable.equiv, ByteTable.equiv, ByteTable.equiv] at h
+  dsimp only [circuit_norm, subcircuit_norm, add32_full, add8_full_carry, Boolean.circuit, ByteLookup] at h
+  simp only [circuit_norm, subcircuit_norm] at h
+  simp only [h_inputs, ByteTable.equiv] at h
```
mitschabaude, line 89: add32 proof is the only case where the previous proof timed out, and I had to switch to `dsimp` first + then `simp` to make it work.
I think this is acceptable, as the add32 circuit is unusually large, and `Add32Explicit` demonstrates an alternative proof using `ExplicitCircuit` that is fast

Clean/Gadgets/Keccak/Chi.lean
```diff
@@ -31,8 +31,11 @@ instance elaborated : ElaboratedCircuit (F p) KeccakState KeccakState where
   output _ i0 := Vector.mapRange 25 fun i => var_from_offset U64 (i0 + i*16 + 8)
 
   local_length_eq state i0 := by simp only [main, circuit_norm, Xor.circuit, And.And64.circuit, Not.circuit]
-  initial_offset_eq state i := by simp only [main, circuit_norm]
-  output_eq state i := by simp only [main, circuit_norm, Xor.circuit, And.And64.circuit, Not.circuit,
+  subcircuits_consistent state i0 := by
+    simp only [main, circuit_norm]
+    intro i
+    and_intros <;> ac_rfl
```
mitschabaude, line 37: subcircuit consistency is usually easy to prove even if it doesn't auto-prove

Clean/Table/Basic.lean
```diff
@@ -243,27 +243,33 @@ variable [Field F] {α : Type}
...
   assignment := .empty W
 
 @[reducible, table_norm, table_assignment_norm]
-def offset (table : TableContext W S F) : ℕ := table.circuit.offset
+def offset (table : TableContext W S F) : ℕ := table.circuit.local_length
 end TableContext
 
 @[reducible, table_norm, table_assignment_norm]
 def TableConstraint (W: ℕ+) (S : Type → Type) (F : Type) [Field F] [ProvableType S] :=
   StateM (TableContext W S F)
```
mitschabaude, line 255: for now, I didn't change the table constraint monad and only made the minimal changes here necessary to keep everything working

Clean/Circuit/Basic.lean
```diff
@@ -231,95 +25,99 @@ def circuit : Circuit F Unit := do
...
 namespace Circuit
-@[reducible, circuit_norm]
-def final_offset (circuit: Circuit F α) (offset: ℕ) : ℕ :=
-  circuit ⟨ offset, .empty offset ⟩ |>.snd.offset
+-- definition of the circuit monad
+
+def bind {α β} (f : Circuit F α) (g : α → Circuit F β) : Circuit F β := fun (n : ℕ) =>
+  -- note: empirically, not unpacking the results of `f` here makes the monad scale to much more operations
+  let (b, ops') := g (f n).1 (n + Operations.local_length (f n).2)
+  (b, (f n).2 ++ ops')
```
gio54321, line 36: I am still pretty much clueless why this particular formulation is more scalable than the other one :)

Clean/Examples/Add32Explicit.lean
```diff
@@ -0,0 +1,60 @@
...
+  -- second version: using `ExplicitCircuit`
+  -- resolve explicit circuit operations
+  rw [ExplicitCircuit.operations_eq]
+  dsimp only [explicit_circuit_norm, explicit, Boolean.circuit]
+  -- simp `constraints_hold` expression
+  simp only [Circuit.constraints_hold.append_soundness, Circuit.constraints_hold.soundness, Gadgets.ByteLookup]
+  -- simp boolean subcircuit soundness and logical/arithmetic/vector expressions
+  simp only [subcircuit_norm, circuit_norm, Nat.reduceAdd]
+  -- finish
+  simp only [Gadgets.ByteTable.equiv]
```
gio54321, line 60: Personally I really like this new simplification approach, seems much simpler

Clean/Circuit/Basic.lean
gio54321: One minor observation is that circuits are append-only just because they are created with pure/bind.
Nothing in the type is actually constraining that the circuits are created that way: they could be arbitrary functions returning arbitrary lists of operations.

Not an issue of course, the assumption that circuits are "well constructed" was present also before (it was much worse actually, with arbitrary state changes possible), but maybe it is worth noting. If the user defines an arbitrary circuit returning an arbitrary list of operations it is on them to prove properties about it, and a lot of lemmas about pure/bind compositionality are not applicable.

gio54321 (approved):
Very nice!

PR #104, opened 2025-06-03
Add `prime_eq_prime'` theorem and definitional equality

oxarbitrage:
Hi, thank you for your work on this repository!

I'm not an expert, but a formal verification enthusiast, and I came across one of the TODOs suggesting a proof of equivalence between `prime` and `prime'`. I decided to give it a try.

This PR adds the following theorem:

```lean
theorem prime_eq_prime' : ∀ p : ℕ, prime p ↔ prime' p
```

It proves that, for every natural number `p`, the two definitions `prime p` and `prime' p` are logically equivalent.

Additionally, it defines:

```lean
noncomputable def prime_eq_prime'_fn : prime = prime'
```

Using function extensionality (`funext`) and propositional extensionality (`propext`), this lifts the equivalence to a definitional equality between the two predicates `prime` and `prime'`, both of type `ℕ → Prop`.

Happy to make any changes or feel free to discard the PR if you think it’s not worth including.

### DIFF
--- a/experiments/Primes.lean
+++ b/experiments/Primes.lean
@@ -2,14 +2,36 @@
 -- define prime numbers, prove enough basic theorems
 -- to show the numbers 0,...,p-1 for p prime are a field
 
+import Mathlib.Data.Nat.GCD.Basic
+
 def prime (p : Nat) :=
   (p ≠ 0) ∧ (p ≠ 1) ∧
   ∀ x : Nat, (x ∣ p) → (x = 1) ∨ (x = p)
 
--- TODO: show that this is equivalent to prime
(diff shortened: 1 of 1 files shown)


mitschabaude:
@oxarbitrage for example, the bits gadget has two sorries at the bottom that want to be filled :)
https://github.com/Verified-zkEVM/clean/blob/572e05c35b9dccdde63c4cc2f74b68c8afecf463/Clean/Gadgets/Bits.lean#L89-L93

oxarbitrage:
Thanks for putting me on track. Happy to test some of those suggestions.

PR #105, opened 2025-06-04
`InductiveTable`: simpler AIR interface

mitschabaude:


gio54321 (approved):
LGTM!
I really like this direction!

PR #107, opened 2025-06-06
add plonky3 backend

katat:
This PR adds a plonky3 backend to demonstrate how the clean DSL can be connected with the proving backends.

The plonky3 backend is based on the fibonacci example from the plonky3 repo. It supports multi-table permutation inspired from the sp1.

The overall workflow:
1. Export both [AST](https://github.com/zksecurity/clean/blob/fe33de5ef5548066f2acc66963dc08e1f5066f00/Clean/Examples/ToJson.lean#L6) of clean Circuit TableOperations and [trace](https://github.com/zksecurity/clean/blob/c8ab8a7d18fa1b056a70cf49ef2a3210b5737dc3/Clean/Examples/Trace.lean#L21) into json files
2. p3 backend [loads](https://github.com/zksecurity/clean/blob/9bcdbb78a09f92547a1b928bb12f263c68d898fd/backends/plonky3/tests/clean_air.rs#L64) the trace and [parse](https://github.com/zksecurity/clean/blob/9bcdbb78a09f92547a1b928bb12f263c68d898fd/backends/plonky3/tests/clean_air.rs#L79) the constraint AST as `MainAir`
3. Two types of constraints to convert: circuit constraints and lookups
	1. [circuit constraints](https://github.com/zksecurity/clean/blob/38a42fcac69fa8c9e84cdc166cae1d60b2ed914c/backends/plonky3/src/clean_air.rs#L56-L115): given a mapping between the assignment variables and the trace columns, the constraint AST can be converted.
	2. lookups: given a lookup type and the column to lookup, the lookup constraints can be achieved via logup permutation argument.
4. logup permutation argument for the lookup supports multi-tables permutation. Here is how it works:
	1. A [lookup](https://github.com/zksecurity/clean/blob/9bcdbb78a09f92547a1b928bb12f263c68d898fd/backends/plonky3/src/key.rs#L40) can be of direction `send` or `receive`, representing + or - respectively in logup calculation.
	2. `send` represents looking up a value (or could be a bunch of values together) in a lookup table.
	3. `receive` represents how many times a lookup table value has been queried, which is determined by the lookup table's multiplicity trace.
	4. At first, `MainAir` will [build the send lookups](https://github.com/zksecurity/clean/blob/38a42fcac69fa8c9e84cdc166cae1d60b2ed914c/backends/plonky3/src/clean_air.rs#L123-L128) if it contains any lookup operations. With these `send` lookups, it generate the multiplicity traces for the corresponding lookup tables. 
	5. With [both send and receive lookups are built](https://github.com/zksecurity/clean/blob/9bcdbb78a09f92547a1b928bb12f263c68d898fd/backends/plonky3/tests/clean_air.rs#L94) for both main table, lookup tables with multiplicity traces, it can generate the permutation trace for each table.
	6. Each permutation trace has the cumulative sum at the end. The sum of all these cumulative sums from the permutation traces is expected to be zero if values being looked up exist in the corresponding lookup tables. [This is checked](https://github.com/zksecurity/clean/blob/9bcdbb78a09f92547a1b928bb12f263c68d898fd/backends/plonky3/src/verifier.rs#L313) by the verifier.
	7. In addition to the air constraints, there are [permutation constraints](https://github.com/zksecurity/clean/blob/9bcdbb78a09f92547a1b928bb12f263c68d898fd/backends/plonky3/src/key.rs#L139) ensuring the permutation traces are generated properly.


backends/plonky3/readme.md
mitschabaude, line 1: as we said, would be good to update this README and explain that this is a not-production-ready PoC

backends/plonky3/tests/trace.json
mitschabaude, line 1: note for another PR: would be nice to get rid of this large JSON and generate it inline in the Rust test

backends/plonky3/readme.md
katat: Also added a placeholder for the upcoming blog post to introduce how it works.

backends/plonky3/tests/trace.json
katat: The test now generates the trace.

We would also do this for the json file of the exported constraints, if the generation process support specifying the public inputs. This could be done in rust, but I think it is more natural to do in lean directly.

katat:
maybe we should also update the ci to run the tests at the backends?

backends/plonky3/tests/trace.json
mitschabaude: nice! yes let's keep the constraints json as is 👍🏻

PR #108, opened 2025-06-06
Add Rotation32 gadget for BLAKE3 

marcobesier:
This PR adds a Rotation32 gadget. The main motivation for this is the goal to ultimately implement a BLAKE3 gadget, which will use 32-bit right rotation as one of its core operations.

Rotation32 is essentially a carbon copy of the existing Rotation64 gadget.

Besides the new rotation gadget, this PR also applies some restructuring to the U32.lean file to make its structure consistent with U64.lean. 

Notice that this PR carries over existing `sorry`s from Rotation64 - at least, for now. In particular, it does _not yet_ cover the latest changes @gio54321 applied in [PR 106](https://github.com/Verified-zkEVM/clean/pull/106).


Clean/Gadgets/Rotation32/ByteRotationTable.lean
mitschabaude, line 1: this table is not actually used right? maybe remove it

mitschabaude (approved):

Clean/Gadgets/Rotation32/ByteRotationTable.lean
marcobesier: Good point! Removed it. 👍

PR #109, opened 2025-06-06
Keccak table

mitschabaude:
* change `InductiveTable` to allow for an extra user input on every row. this essentially makes the table behave like a `foldl` loop, where the circuit defines how you map `(state, input) => state'`
* adds keccak gadget `AbsorbBlock` which XORs an input block with the state, and applies the permutation
* define an `InductiveTable` for keccak, that uses `AbsorbBlock` on every row to map the state + input block on the current row to the new state on the next row

### DIFF
--- a/Clean/Table/Inductive.lean
+++ b/Clean/Table/Inductive.lean
@@ -16,38 +16,43 @@ In the case of two-row windows, an `InductiveTable` is basically a `FormalCircui
 - with assumptions replaced by the spec on the previous row
 - with input offset hard-coded to `size Row`
 -/
-structure InductiveTable (F : Type) [Field F] (Row : Type → Type) [ProvableType Row] where
-  step : Var Row F → Circuit F (Var Row F)
-  spec : ℕ → Row F → Prop
+structure InductiveTable (F : Type) [Field F] (State Input : Type → Type) [ProvableType State] [ProvableType Input] where
+  step : Var State F → Var Input F → Circuit F (Var State F)
+  spec : ℕ → State F → Prop
+  input_spec : ℕ → Input F → Prop := fun _ _ => True
 
   soundness : ∀ (row_index : ℕ) (env : Environment F),
-    -- for all rows
-    ∀ (input_var : Var Row F) (input : Row F), eval env input_var = input →
+    -- for all rows and inputs
+    ∀ (acc_var : Var State F) (x_var : Var Input F)
+      (acc : State F) (x : Input F),
+      (eval env acc_var = acc) ∧ (eval env x_var = x) →
     -- if the constraints hold
-    Circuit.constraints_hold.soundness env (step input_var |>.operations (size Row)) →
-    -- and assuming the spec on the input row
-    spec row_index input →
-    -- we can conclude the spec on the output row
-    spec (row_index + 1) (eval env (step input_var |>.output (size Row)))
+    Circuit.constraints_hold.soundness env (step acc_var x_var |>.operations ((size State) + (size Input))) →
+    -- and assuming the spec on the current row
+    spec row_index acc →
+    -- we can conclude the spec on the next row
+    spec (row_index + 1) (eval env (step acc_var x_var |>.output ((size State) + (size Input))))
 
   completeness : ∀ (row_index : ℕ) (env : Environment F),
-    -- for all rows
-    ∀ (input_var : Var Row F) (input : Row F), eval env input_var = input →
+    -- for all rows and inputs
+    ∀ (acc_var : Var State F) (x_var : Var Input F)
+      (acc : State F) (x : Input F),
+      (eval env acc_var = acc) ∧ (eval env x_var = x) →
     -- when using honest-prover witnesses
-    env.uses_local_witnesses_completeness (size Row) (step input_var |>.operations (size Row)) →
-    -- and assuming the spec on the input row
-    spec row_index input →
+    env.uses_local_witnesses_completeness ((size State) + (size Input)) (step acc_var x_var |>.operations ((size State) + (size Input))) →
+    -- assuming the spec on the current row, and the input_spec on the input
+    spec row_index acc ∧ input_spec row_index x →
     -- the constraints hold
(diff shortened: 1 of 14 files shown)


Clean/Table/Inductive.lean
```diff
@@ -16,38 +16,43 @@ In the case of two-row windows, an `InductiveTable` is basically a `FormalCircui
 - with assumptions replaced by the spec on the previous row
 - with input offset hard-coded to `size Row`
 -/
-structure InductiveTable (F : Type) [Field F] (Row : Type → Type) [ProvableType Row] where
-  step : Var Row F → Circuit F (Var Row F)
-  spec : ℕ → Row F → Prop
+structure InductiveTable (F : Type) [Field F] (State Input : Type → Type) [ProvableType State] [ProvableType Input] where
+  step : Var State F → Var Input F → Circuit F (Var State F)
+  spec : ℕ → State F → Prop
+  input_spec : ℕ → Input F → Prop := fun _ _ => True
```
gio54321: My understanding is that this `input_spec` property is used to ensure some properties about the input witnesses, because we want completeness to hold for any witness generation function (actually, for every input value that is honestly satisfies `input_spec`).
I feel like it could be better named as something like `honest_input_assumptions`.

Also, we did not need this property in `FormalCircuit` because there we either specify assumptions that must hold also in the malicious prover case, or we specify directly the witness generation function.
Am I right? Maybe we could add a comment explaining this choice.

Clean/Table/Inductive.lean
```diff
@@ -16,38 +16,43 @@ In the case of two-row windows, an `InductiveTable` is basically a `FormalCircui
...
-    spec row_index input →
-    -- we can conclude the spec on the output row
-    spec (row_index + 1) (eval env (step input_var |>.output (size Row)))
+    Circuit.constraints_hold.soundness env (step acc_var x_var |>.operations ((size State) + (size Input))) →
+    -- and assuming the spec on the current row
+    spec row_index acc →
+    -- we can conclude the spec on the next row
+    spec (row_index + 1) (eval env (step acc_var x_var |>.output ((size State) + (size Input))))
 
   completeness : ∀ (row_index : ℕ) (env : Environment F),
```
gio54321, line 46: I actually did not fully realize in the previous PR that this is the first time talking about completeness in tables, nice :)

Probably we should find a general instantiation of `completeness` for `FormalTable` at some point, but for I think now it's ok to restrict to particular "shapes" of tables which are easy to reason about.

Clean/Tables/KeccakInductive.lean
```diff
@@ -0,0 +1,73 @@
...
+  simp only [initialState, KeccakState.value]
+  rw [Vector.getElem_map, Vector.getElem_fill, Vector.getElem_fill, U64.from_byte_value, Fin.val_zero]
+
+lemma initialState_normalized : (initialState (p:=p)).is_normalized := by
+  simp only [initialState, KeccakState.is_normalized, Vector.getElem_fill, U64.from_byte_is_normalized]
+  trivial
+
+def formalTable (output : KeccakState (F p)) := table.toFormal initialState output
+
+def domain {p q : Sort*} (_ : p → q) := p
```
gio54321: Is this used anywhere?

Clean/Tables/KeccakInductive.lean
```diff
@@ -0,0 +1,73 @@
...
+def table : InductiveTable (F p) KeccakState KeccakBlock where
+  step state block := do
+    assertion KeccakBlock.normalized block
+    subcircuit AbsorbBlock.circuit { state, block }
+
+  spec i state : Prop :=
+    state.is_normalized
+    ∧ ∃ blocks : List (KeccakBlock (F p)), blocks.length = i
+      ∧ state.value = absorb_blocks (blocks.map KeccakBlock.value)
+
```
gio54321: My understanding of this specification is that it is saying "for the i-th row, it holds that there exists a sequence of blocks such that if I absorb this sequence, then I get the output state matchig what is stored in this row".

Isn't this too weak? Take a silly example table: every row has a different boundary constraint:
- I constrain on row 0 equality with the constant `absorb_blocks ['AA..AA']`
- I constrain on row 1 equality with the constant `absorb_blocks ['BB..BB', 'BB..BB']`
- I constrain on row 2 equality with the constant `absorb_blocks ['CC..CC', 'CC..CC', 'CC..CC']`

and so on...

I can easily prove that for every row the spec is satisfied, because for every `i` I can provide a sequence of blocks that lead to that state.
But probably that's not the intended spec for this table

gio54321 (approved):
Nice!
Tentatively approving this, but left a few comments.
Happy to take second look if needed :slightly_smiling_face:

Clean/Tables/KeccakInductive.lean
mitschabaude: ah lol no, my mistake for leaving that in

Clean/Tables/KeccakInductive.lean
mitschabaude: hm yeah that _is_ a good point. the "exists blocks" should actually come _before_ the "forall rows", i.e. we want to talk about the same blocks on every row

Clean/Table/Inductive.lean
mitschabaude: exactly! I changed the name to `input_assumptions` and added an explanation

Clean/Table/Inductive.lean
mitschabaude: yes, I think in general it might be an easier strategy going forward to treat different "shapes" with separate theorems

Clean/Tables/KeccakInductive.lean
mitschabaude: I refactored everything so that we're able to have the correct statement here! the `spec` now receives the entire list of inputs up to the current row, and the table soundness theorem proves that this holds when passing in the "input" parts of the trace

mitschabaude (commented):
@gio54321 I made a number of changes following your review, ready for re-review!

mitschabaude (commented):
@gio54321 I made a number of changes following your review, ready for re-review!

PR #110, opened 2025-06-10
ThetaD completeness and remove some other sorries

mitschabaude:
* `ThetaD` completeness proof
* Proofs for the bit range check gadget
* Remove sorry in table theorem
* comment out noisy `#eval` calls


gio54321 (approved):
lgtm!

PR #111, opened 2025-06-10
Refactor Byte decomposition and finish proofs

mitschabaude:
This PR refactors the byte decomposition gadget, which splits a byte into a low part (the first `offset < 8` bits) and a high part.

* The new version no longer depends on a custom `TwoPowerLookup` for the specific offset, but only uses `ByteLookup`. This comes at no cost, since two byte lookups are enough to prove soundness.
* Add the missing completeness proofs for `ByteDecomposition`, `U64ByteDecomposition` and `U32ByteDecomposition`
* Simplify proofs in `ByteDecomposition.Theorems`


gio54321 (approved):
Very nice!

PR #112, opened 2025-06-10
Greatly simplify rotation64 bytes soundness proof

gio54321:
Apparently, `omega` suffices to prove soundness for the bytes case!
The proofs should now be trivial to adapt for the 32-bit case.


mitschabaude (approved):
lol, amazing :D

PR #114, opened 2025-06-11
Refactor Bit Rotation gadgets

mitschabaude:
This PR redefines the u64 and u32 bit rotation gadgets to use vector operations.

* Removes redundant lookup constraints on the output. Normalization of the output is now proved from the byte decomposition
* Adds missing completeness proofs
* Reuse existing core of soundness proofs in `Theorems` while simplifying them
* `U64ByteDecomposition` and `U32ByteDecomposition` are no longer needed, and removed
* add a few more theorems for reasoning about u32 / u64 at the vector level

### DIFF
--- a/Clean/Gadgets/Rotation64/Rotation64Bits.lean
+++ b/Clean/Gadgets/Rotation64/Rotation64Bits.lean
@@ -16,23 +16,18 @@ open Bitwise (rot_right64)
 open Rotation64.Theorems
 open ByteDecomposition (Outputs)
 open ByteDecomposition.Theorems (byte_decomposition_lt)
-
--- definitions to prevent `to_vars`/`from_vars` from being unfolded in the proof
--- TODO get rid of these
-def to_vars' (x : Var U64 (F p)) := to_vars x
-def from_vars' (x : Vector (Expression (F p)) (size U64)) := from_vars x
-
 /--
   Rotate the 64-bit integer by `offset` bits
 -/
 def rot64_bits (offset : Fin 8) (x : Var U64 (F p)) : Circuit (F p) (Var U64 (F p)) := do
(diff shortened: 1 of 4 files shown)


Clean/Gadgets/Rotation32/Rotation32Bits.lean
```diff
@@ -14,75 +14,127 @@ instance : Fact (p > 512) := by
...
 open Bitwise (rot_right32)
-open Gadgets.Rotation32.Theorems (rotation32_bits_soundness)
+open Gadgets.Rotation32.Theorems
+open ByteDecomposition (Outputs)
+open ByteDecomposition.Theorems (byte_decomposition_lt)
+
+-- definitions to prevent `to_vars`/`from_vars` from being unfolded in the proof
+-- TODO get rid of these
+def to_vars' (x : Var U32 (F p)) := to_vars x
+def from_vars' (x : Vector (Expression (F p)) (size U32)) := from_vars x
```
gio54321: Wouldn't those be better defined at `U32` level?
Probably defining two functions `U32.to_limbs {T : Type} : U32 T -> Vector T (size U32)` and `U32.from_limbs {T : Type} : Vector T (size U32) -> U32`.
It should also be trivial to prove that this is an isomorphism; this would avoid, by default, clashing with low-level `{from, to}_vars` (also applies to 64-bit variant).

Thinking about it, as a long-term refactor, it would also make sense to change the definition of `U{32, 64}` to a `ProvableVector`, and then provide generic rotation/addition/xor/... for any byte size!

gio54321 (approved):
Very nice!

Clean/Gadgets/Rotation32/Rotation32Bits.lean
mitschabaude: Good suggestion - done! I also agree with the longer-term refactor

PR #116, opened 2025-06-13
Get rid of copy constraint in bit rotation gadget

mitschabaude:
As the title says -- I didn't like that we copied variables just to make the output of bit rotation gadgets easier to deal with. In fact, the concrete value of a circuit's output rarely matters in parent circuits, and a reasonable strategy to reduce complexity is to wrap the output in a method that parent circuits never have to unfold.

* Abstract away output of bit rotation and general rotation gadgets
* Adapt many hard-coded offsets in Keccak elaborated circuits, since we reduced the number of variables
* Adapt and simplify Blake3 G proof

### DIFF
--- a/Clean/Gadgets/BLAKE3/BLAKE3G.lean
+++ b/Clean/Gadgets/BLAKE3/BLAKE3G.lean
@@ -56,8 +56,6 @@ def main (a b c d : Fin 16) (input : Var Inputs (F p)) : Circuit (F p) (Var BLAK
                     |>.set d state_d
   return state
 
--- #eval (main (p:=p_babybear) 0 1 2 3 default).local_length
--- set_option profiler true in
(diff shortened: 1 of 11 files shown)


gio54321 (approved):
Nice! A couple of comments, mainly not regarding the PR itself.

I wonder if we can avoid wrapping the output into an opaque definition by removing `ElaboratedCircuit.output` from `circuit_norm` ([src](https://github.com/Verified-zkEVM/clean/blob/28b57b21c078b65323d2bb2d0c1065d01453718a/Clean/Circuit/Basic.lean#L233)). As you said, in most cases, we do not want to unfold the output definition.

Second, is there anything we can do to avoid the leakage of implementation details through `local_length` and `output`? It is quite annoying to change every gadget down the dependency chain.

PR #117, opened 2025-06-13
Add BLAKE3 permutation

gio54321:
Add BLAKE3 permutation, with the corresponding soundness and completeness proofs ([reference](https://www.ietf.org/archive/id/draft-aumasson-blake3-00.html#name-message-word-permutation))

### DIFF
--- a/Clean.lean
+++ b/Clean.lean
@@ -8,3 +8,4 @@ import Clean.Tables.Fibonacci32Inductive
 import Clean.Tables.KeccakInductive
 import Clean.Gadgets.Bits
 import Clean.Gadgets.BLAKE3.BLAKE3G
+import Clean.Gadgets.BLAKE3.Permute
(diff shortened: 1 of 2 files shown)


mitschabaude (approved):
I wonder if it's even useful to define this operation as a circuit?
since it's actually a pure function without circuit operations.

mayke it'd be enough to create a theorem that says something like
```lean
def permute' state := Vector.ofFn (fun i => state[msgPermutation[i]])

theorem permute_spec {state} :
  state.is_normalized →
    (permute' state).value = permute state.value
    ∧ (permute' state).is_normalized
  := by sorry
```
..and use that in other gadget proofs where needed

gio54321:
Right, that's a good point @mitschabaude
I'm going to merge as is for now, but probably we can remove this gadget later on.

Another instance for a similar zero-constraints gadget is the [byte rotation](https://github.com/Verified-zkEVM/clean/blob/main/Clean/Gadgets/Rotation64/Rotation64Bytes.lean#L14-L32)

PR #118, opened 2025-06-13
 Add a new `assert_equals` notation of the form `===`

oxarbitrage:
Introduce a unified `===` notation for equality assertions in circuits. Under the hood, we keep the original `assert_equals` gadget and add:

- A `HasAssertEq` class and instances for `Expression F` and `α (Expression F)`.
- A `HasAssertEqM` class and MonadLift-based instance so `x === y` works in any monad built over `Circuit F`.
- Notation `infix:50 " === " => assert_eqM'`.
- `@[circuit_norm]/@[reducible]` on the defs/instances and a small lemma so that `dsimp only [circuit_norm]` (or `simp`) unfolds `===` back to the core `assert_equals` call.

It was tricky to make inference and normalization line up everywhere, so the added boilerplate is larger than expected, happy to refine further if requested.

### DIFF
--- a/Clean/Gadgets/Equality.lean
+++ b/Clean/Gadgets/Equality.lean
@@ -107,16 +107,63 @@ theorem completeness (α : TypeMap) [ProvableType α] (n : ℕ) (env : Environme
 end Equality
 end Gadgets
 
--- this is exported at the top level because it is a core builtin gadget
+/-
+Provides a unified `===` notation for asserting equality in circuits:
+- Core `assert_equals` for `α (Expression F)` and `Expression F`.
+- `HasAssertEq` / `HasAssertEqM` dispatch so `assert_equals_generic x y`
+  and `x === y` pick the right gadget and lift via `MonadLift`.
+- Marked `@[circuit_norm]`/`@[reducible]` so `dsimp only [circuit_norm]`
+  unfolds `===` to the original `assert_equals`.
+-/
 @[circuit_norm]
-def assert_equals {α : TypeMap} [ProvableType α] (x y : α (Expression F)) : Circuit F Unit :=
+def assert_equals {F : Type} [Field F] {α : TypeMap} [ProvableType α]
+  (x y : α (Expression F)) : Circuit F Unit :=
   assertion (Gadgets.Equality.circuit α) (x, y)
 
--- TODO unfortunately, if the inputs to `assert_equals` are just `Expression F`,
--- Lean doesn't come up with `α = id` -- even though `ProvableType` is inferred when `(α:=id)` is passed explicitly.
--- this definition is a somehwat natural alternative that can be used on `Expression F` directly
-@[reducible]
-def Expression.assert_equals (x y : Expression F) : Circuit F Unit :=
+@[circuit_norm, reducible]
+def Expression.assert_equals {F : Type} [Field F]
+  (x y : Expression F) : Circuit F Unit :=
   assertion (Gadgets.Equality.circuit id) (x, y)
 
--- TODO define a custom syntax e.g. `===` that figures out whether to use `assert_equals` or `Expression.assert_equals`
+class HasAssertEq (β : Type) (F : outParam Type) [Field F] where
+  assert_eq : β → β → Circuit F Unit
+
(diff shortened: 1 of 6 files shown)


Clean/Gadgets/Equality.lean
```diff
@@ -112,16 +112,63 @@ theorem uses_local_witnesses (α : TypeMap) [ProvableType α] (n : ℕ) (env : E
...
+  {m : Type → Type} [MonadLift (Circuit F) m] :
+  HasAssertEqM β m F where
+  assert_eqM x y := MonadLift.monadLift (assert_equals_generic x y)
+
+@[circuit_norm, reducible, simp]
+def assert_eqM' {β : Type} {m : Type → Type} {F : Type} [Field F] [HasAssertEqM β m F]
+  (x y : β) : m Unit :=
+  HasAssertEqM.assert_eqM x y
+
+infix:50 " === " => assert_eqM'
```
mitschabaude: is there a reason that this is not simply `HasAssertEq.assert_eq`? I tried and it builds with that change.
```suggestion
infix:50 " === " => HasAssertEq.assert_eq
```

Clean/Gadgets/Equality.lean
```diff
@@ -112,16 +112,63 @@ theorem uses_local_witnesses (α : TypeMap) [ProvableType α] (n : ℕ) (env : E
...
--- TODO define a custom syntax e.g. `===` that figures out whether to use `assert_equals` or `Expression.assert_equals`
+class HasAssertEq (β : Type) (F : outParam Type) [Field F] where
+  assert_eq : β → β → Circuit F Unit
+
+@[circuit_norm, reducible]
+instance {F : Type} [Field F] : HasAssertEq (Expression F) F where
+  assert_eq := Expression.assert_equals
+
+@[circuit_norm, reducible]
+instance {F : Type} [Field F] {α : TypeMap} [ProvableType α] :
```
mitschabaude: I don't think these have to be `reducible`, and we can replace the two `circuit_norm` attributes with just one on `HasAssertEq.assert_eq`:

```lean
attribute [circuit_norm] HasAssertEq.assert_eq
```

Clean/Gadgets/Equality.lean
```diff
@@ -112,16 +112,63 @@ theorem uses_local_witnesses (α : TypeMap) [ProvableType α] (n : ℕ) (env : E
...
+def assert_eqM' {β : Type} {m : Type → Type} {F : Type} [Field F] [HasAssertEqM β m F]
+  (x y : β) : m Unit :=
+  HasAssertEqM.assert_eqM x y
+
+infix:50 " === " => assert_eqM'
+
+@[circuit_norm]
+lemma HasAssertEq.assert_eq_eq {β : Type} {F : Type} [Field F] [inst : HasAssertEq β F]
+  (x y : β) :
+  inst.assert_eq x y = assert_equals_generic x y := rfl
```
mitschabaude: I think we don't need this lemma, do we? It introduces a tendency for simp to get stuck in a cycle

mitschabaude (commented):
This is an amazing contribution!! A few comments from me:

Clean/Gadgets/Equality.lean
mitschabaude: which would mean we can get rid of `HasAssertEqM` entirely

oxarbitrage: Thanks! That makes things a lot easier. I wish i had found out that before.

Changes made at https://github.com/Verified-zkEVM/clean/pull/118/commits/7243fe24de20768daa290fd50ff2e2c68f114a98

Clean/Gadgets/Equality.lean
oxarbitrage: Not needed now the monad class is gone.

Clean/Gadgets/Equality.lean
oxarbitrage: Thx, used it in https://github.com/Verified-zkEVM/clean/pull/118/commits/7243fe24de20768daa290fd50ff2e2c68f114a98

mitschabaude (approved):
🚀

PR #127, opened 2025-06-17
Cross-Table Lookups

### LINKED ISSUE #125: Cross-Table Lookups
there are two different kinds of lookup tables:

* _fixed_ tables. entries are hard-coded up front (and known to the verifier).
  * example: the "byte table" with the values 0, ..., 255.
* "dynamic" tables. entries are _defined_ by which lookups are done during the definition of other tables. this is used for cross-table interaction in zkvms like SP1.
  * example: a keccak precompile allows you to look up the triple `(state, block, absorb_block state block)` for any keccak state and block to be hashed.

right now, clean only supports the first kind. the goal is to change the `Lookup` type to also cover lookups of the second type.

in particular, it should then be possible to transform any `FormalCircuit` into a `Lookup`. the resulting interface would be similar to `subcircuit`: `Table.contains` should be rewritable to `SubCircuit.{soundness, completeness}`

mitschabaude:
closes #125

* Generalize `Table` to a definition that doesn't assume we have an actual table at hand
* Enable table constraints to be rewritten into custom `soundness` and `completeness` statements
* Add a new definition `FlatOperation.dynamicWitnesses` which computes the witnesses of a circuit _without_ having an environment. From that, we can define an environment, called `proverEnvironment`
* Add a new property of circuits called `computableWitnesses`
* Prove that `computableWitnesses` implies that the `proverEnvironment` satisfies `uses_local_witnesses` (the assumption we use for completeness)
  * this is the largest portion of the PR, see `Theorems.lean`
* Define a `LookupCircuit` which extends `FormalCircuit` by also satisfying `computableWitnesses`. Given a `LookupCircuit`, we can derive a `lookup` into a `Table` that has exactly the same interface as the original formal circuit (i.e., soundness is `assumptions -> spec` while completeness is just `assumptions`)

On another note, I discovered the [Mathlib naming conventions](https://leanprover-community.github.io/contribute/naming.html) and started to follow them (so far, we've mostly been using snake case which is "wrong", Mathlib uses camel case for definitions/methods and snake case only for theorem names)


Clean/Gadgets/Addition8/Addition8FullCarry.lean
```diff
@@ -132,4 +131,13 @@ def circuit : FormalCircuit (F p) Inputs Outputs where
 
     exact ⟨completeness1, completeness2, completeness3⟩
 
+def lookupCircuit : LookupCircuit (F p) Inputs Outputs := {
+  circuit with
+  name := "Addition8FullCarry"
+
+  computableWitnesses n input := by
+    simp_all only [circuit_norm, subcircuit_norm, circuit, add8_full_carry, Boolean.circuit,
+      Operations.forAllFlat, Operations.toFlat, FlatOperation.forAll, Inputs.mk.injEq]
```
mitschabaude, line 140: this proof ended up being quite easy after tweaking the `computableWitnesses` statement in formal circuits

Clean/Circuit/Basic.lean
```diff
@@ -370,14 +405,6 @@ def Operations.witness_generators : (ops: Operations F) → Vector (Environment
   | .lookup _ :: ops => witness_generators ops
   | .subcircuit s :: ops => (s.local_length_eq ▸ FlatOperation.witness_generators s.ops) ++ witness_generators ops
 
--- TODO this is inefficient, Array should be mutable and env should be defined once at the beginning
-def Circuit.witnesses (circuit: Circuit F α) (offset := 0) : Array F :=
-  let generators := (circuit.operations offset).witness_generators
-  generators.foldl (fun acc compute =>
-    let env i := acc.getD i 0
-    acc.push (compute ⟨ env ⟩))
-  #[]
```
mitschabaude, line 379: fun fact - this definition was wrong, because it doesn't recurse into subcircuits, so _all_ witnesses in subcircuits are created from an environment that is zero on the subcircuit's local variables.

`FlatOperation.dynamicWitnesses` is the corrected definition, it operates on the flattened operations.

Clean/Circuit/LookupCircuit.lean
```diff
@@ -0,0 +1,113 @@
...
+    simp only [h_output, LookupCircuit.constantOutput, and_true]
+    set env := circuit.proverEnvironment input
+    apply circuit.original_completeness 0 env (const input) input ProvableType.eval_const h_assumptions
+    exact circuit.proverEnvironment_uses_local_witnesses input
+
+-- we create another `FormalCircuit` that wraps a lookup into the table defined by the input circuit
+-- this gives `circuit.lookup input` _exactly_ the same interface as `subcircuit circuit input`.
+
+@[circuit_norm]
+def lookupCircuit (circuit : LookupCircuit F α β) : FormalCircuit F α β where
```
gio54321, line 64: Morally, here we are defining a new table where each row is `(input, output)`, and the circuit's constraints are applied to each row?
Therefore, a lookup into a table of that shape is equivalent to instantiating that circuit as a subcircuit.

Interestingly, the assumptions are needed to hold on the "calling" table inputs, and not on the sub-table.
In other words, in concrete instantiations, we could have a table of this shape where you witness some rows for which their inputs do not satisfy the assumptions.
However, to prove soundness and completeness of the caller circuit, we need to satisfy assumptions on the inputs to get soundness, so that's OK.

Not an issue afaiu, just something to keep in mind.
For now, I believe we have only dealt with full tables without making any assumptions (except for honest input assumptions for completeness).

Clean/Circuit/Lookup.lean
```diff
@@ -0,0 +1,81 @@
...
+  name: String
+  arity: ℕ
+  length: ℕ
+  row: Fin length → Vector F arity
+  -- TODO this would make sense if we had separate input and output types,
+  -- and the lookup would automatically witness the output given the input.
+  -- then we could weaken completeness to be `index input < length`!
+  index: Vector F arity → ℕ
+  soundness: Vector F arity → Prop
+  completeness: Vector F arity → Prop
```
gio54321, line 63: Why do we want to have two separate propositions for soundness and completeness for static tables?
In all the examples of static tables we have, they are equal.
It is probably helpful to define "partial" tables, which do not have a row for each possible case. Not sure

Clean/Circuit/Lookup.lean
```diff
@@ -0,0 +1,81 @@
...
+  there should be a concrete way of instantiating the table where `contains` is proved to hold on every row.
+  -/
+  contains : Vector F arity → Prop
+
+  /--
+  we allow to rewrite the `contains` property into two statements that are easier to work with
+  in the context of soundness and completeness proofs.
+  -/
+  soundness : Vector F arity → Prop
+  completeness : Vector F arity → Prop
```
gio54321, line 18: It seems to me that `soundness` is the semantic property we care about when looking up, while `completeness` is what we previously referred to as the "honest input assumptions."
I think calling it `completeness` is a bit confusing.

Clean/Circuit/Lookup.lean
mitschabaude: that's a good point actually, I think we won't ever need different soundness and completeness properties. I'll combine them in a follow-up PR!

PR #135, opened 2025-06-24
Lighter Mathlib dependency

yoichi-zksecurity:
I followed outputs of the min_imports command. The changes aim at importing only those Mathlib files that are needed.

I don't know whether we want to optimize imports this way. Caches work well enough?


mitschabaude:
excellent!

mitschabaude (approved):

PR #152, opened 2025-06-27
Some cleanup and renaming

mitschabaude:
various small things I wanted to clean up:
* rename a few more definitions according to lean conventions
* change the alpha and beta types in formal circuit definition to much clearer `Input` and `Output`
* get rid of `SubcircuitSoundness`, `SubcircuitCompleteness` definitions which were only used in one place and had to be unfolded in every proof
* consistently use `vector[i]` instead of `vector.get i`
* remove changes to the default `simp` set, I feel like we're not supposed to do that given that we maintain our own simp sets already

### DIFF
--- a/Clean/Circuit/Basic.lean
+++ b/Clean/Circuit/Basic.lean
@@ -198,7 +198,7 @@ def Environment.UsesLocalWitnessesFlat (env : Environment F) (n : ℕ) (ops : Li
 
 section
 open Circuit (ConstraintsHold)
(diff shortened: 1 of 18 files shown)


yoichi-zksecurity (approved):
Looks good to me.

I have a TODO to see `circuit_norm` in action to get a feeling of what it's supposed to do. That doesn't block me from approving this (seems to be getting more consistent).

PR #163, opened 2025-07-01
Kick off Circomlib in Clean

### LINKED ISSUE #160: Add stubs of 2-3 circomlib gadgets to act as examples for an LLM or human
there don't have to be proofs, just the circuit logic translated from circom to clean, and the scaffolding for formalcircuit, and the start of the proofs which just simplifies the statement

### LINKED ISSUE #161: Better primitive to represent most circom loops
circom loops
* are used a lot
* always go over a numeric range
* often depend on the last iteration, foldl style
* often fill out arrays by mutation

the last point is hard to represent in lean, in particular since we would want the output to be a vector and prove the length; but we can always fill arrays with witnesses up front, and then do a foldl like loop to add the constraints.

the current `foldl` primitive assumes constant outputs which might not be satisfied in general, so we should ad a foldl-like primitive that doesn't assume constant outputs, and loops over a numeric range

mitschabaude:
closes #160 
closes #161

I had two goals when writing the circomlib circuits:
* they should create exactly the same R1CS constraints as circom would
* they should look as similar to the source as possible (without changing the general look of clean)

For example, for `Num2Bits`, I could have just duplicated or reused our own `ToBits` gadget which works exactly the same. However, it looks quite different to the circomlib implementation, as it does everything in one loop. I added `Circuit.foldlRange` to match more closely what circom does, here and in other gadgets.

Note that I also added the full proof for `Num2Bits`, as a sanity check that `foldlRange` works well.


Clean/Circomlib/Comparators.lean
mitschabaude, line 1: this file is mostly scaffolding, generated with the help of AI. I think we can scaffold most circuits this way

Clean/Circuit/Loops.lean
```diff
@@ -194,22 +211,23 @@ theorem output_eq :
   case cons x xs ih =>
     rw [foldlM_cons, bind_output_eq, ih, constant.localLength_eq (init, x), finFoldl_cons_succ]
 
-variable [Inhabited β] [Inhabited α]
```
mitschabaude, line 197: there's a lot of work here to make the general foldl theory _not_ depend on the assumption that the input types are `Inhabited`.

why? because it would have meant that `foldlRange` doesn't work for length `n = 0`: because `Fin 0` is not inhabited! that would've been ugly, the > 0 assumptions would have propagated to many circom circuits, which actually work fine for the 0 case.

PR #172, opened 2025-07-03
Add FormalCircuitWithUniqueOutput class for deterministic circuits

yoichi-zksecurity:
## Summary
- Add new `FormalCircuitWithUniqueOutput` class that extends `FormalCircuit` with explicit uniqueness constraints
- Include U32 copy circuit example as a sanity check for the new class

## Test plan
- [x] Code builds successfully (`lake build` passes)
- [x] Uniqueness proof works for simple deterministic circuit (U32 copy)

### DIFF
--- a/Clean/Circuit/Basic.lean
+++ b/Clean/Circuit/Basic.lean
@@ -287,6 +287,17 @@ structure FormalCircuit (F: Type) [Field F] (Input Output: TypeMap) [ProvableTyp
   soundness : Soundness F elaborated Assumptions Spec
   completeness : Completeness F elaborated Assumptions
 
+/--
+`FormalCircuitWithUniqueOutput` extends `FormalCircuit` with an explicit uniqueness constraint.
+This ensures that for any input satisfying the assumptions, the specification uniquely determines the output.
+Use this class when you want to formally guarantee that constraints uniquely determine the output,
+preventing ambiguity in deterministic circuits.
+-/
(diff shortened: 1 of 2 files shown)


mitschabaude (commented):
I feel like this could be done differently and with a simpler interface: The `Spec` could just be a function from input to output, and thus uniquely determine the output by design. (can't extend `FormalCircuit` in that case though)

Clean/Circuit/Basic.lean
```diff
@@ -287,6 +287,17 @@ structure FormalCircuit (F: Type) [Field F] (Input Output: TypeMap) [ProvableTyp
   soundness : Soundness F elaborated Assumptions Spec
   completeness : Completeness F elaborated Assumptions
 
+/--
+`FormalCircuitWithUniqueOutput` extends `FormalCircuit` with an explicit uniqueness constraint.
+This ensures that for any input satisfying the assumptions, the specification uniquely determines the output.
+Use this class when you want to formally guarantee that constraints uniquely determine the output,
+preventing ambiguity in deterministic circuits.
+-/
+structure FormalCircuitWithUniqueOutput (F: Type) [Field F] (Input Output: TypeMap) [ProvableType Input] [ProvableType Output]
```
mitschabaude: How about renaming it to `DeterministicFormalCircuit`?

mitschabaude (approved):
anyway I think we can ship this and add the simpler intf later

PR #178, opened 2025-07-03
Use subcircuits like functions

mitschabaude:
inspired by circomlib, this lets us use subcircuits like functions

what I like most is that this unifies the usage of `FormalCircuit`, `FormalAssertion` and `GeneralFormalCircuit`

also: makes `assertBool` a core library primitive by exporting at the top level + adding to the simp set


Clean/Circuit/Subcircuit.lean
```diff
@@ -250,6 +250,17 @@ def subcircuitWithAssertion (circuit: GeneralFormalCircuit F β α) (b: Var β F
...
+-- we'd like to use subcircuits like functions
+
+instance : CoeFun (FormalCircuit F β α) (fun _ => Var β F → Circuit F (Var α F)) where
+  coe circuit input := subcircuit circuit input
+
+instance : CoeFun (FormalAssertion F β) (fun _ => Var β F → Circuit F Unit) where
+  coe circuit input := assertion circuit input
+
+instance : CoeFun (GeneralFormalCircuit F β α) (fun _ => Var β F → Circuit F (Var α F)) where
+  coe circuit input := subcircuitWithAssertion circuit input
```
mitschabaude, line 262: main change is here!

yoichi-zksecurity (approved):
Looks good to me.

Issue #164, opened 2025-07-02
Formalize circom gates.circom


yoichi-zksecurity:
This issue keeps track of formalizing [gates.circom](https://github.com/iden3/circomlib/blob/master/circuits/gates.circom) in the style of #163 .

yoichi-zksecurity:
Assigning myself at least to fill in the description.

PR #184, opened 2025-07-04
Add MultiAND gate implementation to complete gates.circom formalization

yoichi-zksecurity:
## Summary
- Implements the recursive multi-input AND gate from circomlib's gates.circom
- This is the last circuit in the gates.circom file  
- Follows the divide-and-conquer approach from the original circom implementation
- **Complete with soundness and completeness proofs**

## Implementation
- Recursive structure: splits n-input AND into two smaller ANDs for n≥3
- Base cases: n=0 returns 1, n=1 returns input, n=2 uses binary AND

## Test plan
lake build succeeds

Finishes #164


yoichi-zksecurity:
Draft because proofs are being added.

Clean/Utils/Vector.lean
```diff
@@ -11,6 +11,90 @@ def fromList (l: List α) : Vector α l.length := ⟨ .mk l, rfl ⟩
 namespace Vector
 def len (_: Vector α n) : ℕ := n
 
+/-- Membership in Vector.toList: if x ∈ v.toList, then x = v.get i for some i -/
+theorem mem_toList_iff_get {α : Type*} {n : ℕ} (v : Vector α n) (x : α) :
+    x ∈ v.toList ↔ ∃ i : Fin n, x = v.get i := by
```
mitschabaude: this theorem feels like it should not exist:

* using `GetElem` syntax `v[i]` or `v[i.val]` is the preferred way to express indexing into vectors, it looks better and is better supported by existing lemmas. `Vector.get` is a 'lower-level' method that probably shouldn't be used
* after replacing with `v.get i` with `v[i]`, this theorem is just the combination of two equivalences:
  * `x ∈ v.toList ↔ x ∈ v` (if this doesn't exist, you can add it)
  * `x ∈ v ↔ ∃ i : Fin n, x = v[i]` (I'm pretty sure this exists!)

it seems more natural to me to remove this and just use the two theorems above directly whereever you need it

Clean/Utils/Vector.lean
```diff
@@ -11,6 +11,90 @@ def fromList (l: List α) : Vector α l.length := ⟨ .mk l, rfl ⟩
...
+      rw [h_list_len, h_len]
+      exact i.isLt
+    use h_bound
+    -- Now show v.toList[i.val] = v.get i
+    simp only [Vector.get]
+    rfl
+
+-- Helper lemma: A vector of length 1 has toList = [v.get 0]
+theorem toList_length_one {α : Type} (v : Vector α 1) :
+    v.toList = [v.get 0] := by
```
mitschabaude:
```suggestion
theorem toList_length_one {α : Type} (v : Vector α 1) :
    v.toList = [v[0]] := by
```

Clean/Utils/Vector.lean
```diff
@@ -244,6 +328,35 @@ theorem append_take_drop {v : Vector α (n + m)} :
   simp only [hi', reduceDIte, getElem_cast, getElem_extract]
   congr
   omega
+
+/-- Splitting a vector via extract and then converting to lists gives concatenation -/
+theorem toList_extract_append {α : Type*} {n : ℕ} (v : Vector α n) (k : ℕ) (hk : k ≤ n) :
+    v.toList = (⟨v.toArray.extract 0 k, by simp [Array.size_extract, v.size_toArray]; exact hk⟩ : Vector α k).toList ++
+               (⟨v.toArray.extract k n, by simp [Array.size_extract, v.size_toArray]⟩ : Vector α (n - k)).toList := by
```
mitschabaude, line 275: I can somehow see the value of a `toList` version since it avoids `cast`, but still think it should be turned into a statement about the vector directly

also, it would be nicer to use `take` and `drop` as above, without casts, and just a final cast to the right length at the end

Clean/Utils/Vector.lean
mitschabaude: actually, why is this about `.toList`? it should just be
```lean
v = #v[v[0]]
```

Clean/Utils/Vector.lean
```diff
@@ -11,6 +11,90 @@ def fromList (l: List α) : Vector α l.length := ⟨ .mk l, rfl ⟩
...
+        | [x] =>
+          -- Goal: v.toList = [v.get 0]
+          -- v.toList = arr.toList = lst = [x]
+          -- v.get 0 = arr[0] = lst[0] = x
+          rfl
+        | _ :: _ :: _ => simp [List.length] at h
+
+-- Helper lemma: A vector of length 2 has toList = [v.get 0, v.get 1]
+theorem toList_length_two {α : Type} (v : Vector α 2) :
+    v.toList = [v.get 0, v.get 1] := by
```
mitschabaude:
```suggestion
theorem toList_length_two {α : Type} (v : Vector α 2) :
    v = #v[v[0], v[1]] := by
```

Clean/Utils/Vector.lean
```diff
@@ -11,6 +11,90 @@ def fromList (l: List α) : Vector α l.length := ⟨ .mk l, rfl ⟩
...
+    rfl
+
+-- Helper lemma: A vector of length 1 has toList = [v.get 0]
+theorem toList_length_one {α : Type} (v : Vector α 1) :
+    v.toList = [v.get 0] := by
+  -- Try using cases on the vector
+  cases v using Vector.casesOn with
+  | mk arr h =>
+      cases arr using Array.casesOn with
+      | mk lst =>
```
mitschabaude, line 22: you should be able to use a single `rcases` which unpacks the vector into a length-1 list, and then `rfl` should do the trick

Clean/Utils/Bitwise.lean
```diff
@@ -1,7 +1,28 @@
 import Mathlib.Analysis.Normed.Ring.Lemmas
 import Clean.Utils.Field
+import Mathlib.Data.Nat.Bitwise
 
 namespace Bitwise
+
+section BitwiseAND
```
mitschabaude: namespace feels overkill to me here, these fit well with the other theorems in one namespace. but I'd prefer if you put theorems below definitions, i.e. move this further down below

Clean/Utils/BinaryOps.lean
```diff
@@ -0,0 +1,106 @@
...
+  rw [Vector.mem_toList_iff_get] at h_mem
+  rcases h_mem with ⟨i, hi⟩
+  rw [hi]
+  exact h_vec i
+
+/-- Specialized version for F p where binary means = 0 or = 1 -/
+theorem Vector.toList_binary_field {n : ℕ} (v : Vector (F p) n) :
+    (∀ i : Fin n, v.get i = 0 ∨ v.get i = 1) →
+    (∀ x ∈ v.toList, x = 0 ∨ x = 1) := 
+  Vector.toList_binary v (fun x => x = 0 ∨ x = 1)
```
mitschabaude: these theorems feel overly specific, and shouldn't be here! use the `mem_iff_getElem`-type theorem directly

Clean/Utils/BinaryOps.lean
```diff
@@ -0,0 +1,106 @@
...
+      have h_one_one : (1 : ℕ) &&& 1 = 1 := by
+        simp only [HAnd.hAnd, AndOp.and]
+        rfl
+      rw [h_one_one]
+      exact h_tail_binary
+
+/-- For binary values and binary lists, a &&& foldl orig l = foldl (a &&& orig) l -/
+theorem List.and_foldl_eq_foldl_of_all_binary (a : ℕ) (orig : ℕ) (l : List ℕ)
+    (_ha : a = 0 ∨ a = 1) (hl : ∀ x ∈ l, x = 0 ∨ x = 1) :
+    a &&& List.foldl (· &&& ·) orig l = List.foldl (· &&& ·) (a &&& orig) l := by
```
mitschabaude: I feel like this must be true without the binary assumption?

Clean/Utils/BinaryOps.lean
```diff
@@ -0,0 +1,106 @@
...
+
+variable {p : ℕ} [Fact p.Prime]
+
+open Bitwise (and_zero_absorb and_one_id_binary)
+
+section ListOperations
+
+/-- Folding AND over a list of binary values gives a binary result -/
+theorem List.foldl_and_binary (l : List ℕ) :
+    (∀ x ∈ l, x = 0 ∨ x = 1) → (List.foldl (· &&& ·) 1 l = 0 ∨ List.foldl (· &&& ·) 1 l = 1) := by
```
mitschabaude: the binary assumption on `l` is actually not needed here! if you AND with something that is 1 bit it stays 1 bit.

I think we should introduce a property `IsBool` or `Binary` to make these more readable

```suggestion
theorem List.foldl_and_binary (l : List ℕ): Binary (List.foldl (· &&& ·) 1 l) := by
```

Clean/Circomlib/Gates.lean
```diff
@@ -9,6 +14,29 @@ https://github.com/iden3/circomlib/blob/master/circuits/gates.circom
...
+open Circuit (bind_output_eq bind_localLength_eq bind_forAll)
+open Operations (append_localLength)
+open BinaryOps (List.foldl_and_binary List.and_foldl_eq_foldl_of_all_binary Vector.toList_binary_field)
+
+section HelperLemmas
+
+/-- Extract individual field element from vector evaluation -/
+lemma eval_get_eq {n : ℕ} {env : Environment (F p)} {input_var : Var (fields n) (F p)}
+    {input : fields n (F p)} (h_env : eval env input_var = input) (i : Fin n) :
+    env (input_var.get i) = input.get i := by
```
mitschabaude: * shouldn't use `Vector.get`.
* you sure that `circuit_norm` can't auto-prove this?

mitschabaude (commented):
some comments, really great that you could prove this!

Clean/Utils/BinaryOps.lean
yoichi-zksecurity: I'll go with `IsBool`, and refactor the other gates and come back here.

yoichi-zksecurity: I realized that #190 introduces `IsBool` on field elements only. Here `IsBool` on `ℕ` is wanted. I'll cut another branch and generalize that to anything with `0` and `1` (and `< 2`).

yoichi-zksecurity: I've updated #190 to deal with natural numbers too. I'll rebase this branch on top.

yoichi-zksecurity: Anyway this slight generalization seems to be necessary for the induction to go through.

yoichi-zksecurity: done.

Clean/Utils/Vector.lean
yoichi-zksecurity: This has been done.

Clean/Utils/Vector.lean
yoichi-zksecurity: > v = #v[v[0]]

The proof script doesn't need any change; that strongly suggests the reasoning is mostly not about `toList`.

yoichi-zksecurity:
```
theorem length_one_form {α : Type} (v : Vector α 1) :
    v = #v[v[0]] := by
```
is provable, but it's very hard to use.

For instance,
```
theorem toList_of_length_one_form {α : Type} (v : Vector α 1) :
    v.toList = [v[0]] := by
  rw[length_one_form]
```
doesn't work.
```
  rw[length_one_form v]
```
rewrites the term into
```
{ toArray := #[v[0]], size_toArray := ⋯ }.toList
```

```
  simp[length_one_form, Array.toList]
```
doesn't work either (maximum recursion depth has been reached).

yoichi-zksecurity:
```
v.toList = [v[0]]
```
is a nicer statement because it gets you out of the dependent types cleanly.

Clean/Utils/Vector.lean
yoichi-zksecurity: This has gone.

yoichi-zksecurity:
The remaining points are around what we do with `Vector`.

I have a feeling that this PR is turning vectors into lists more aggressively than the rest of the project. I need to go and see what is happening in other proofs in the project.

PR #186, opened 2025-07-07
Add structural rules for FormalCircuit transformations

yoichi-zksecurity:
## Summary
Add structural transformation rules for FormalCircuit and related types to enable deriving new verified circuits from existing ones.

## Changes
- Add `strengthenAssumptions` method to create circuits with stronger preconditions
- Add `deriveSpec` method to create circuits with different specifications
- Implement similar rules for GeneralFormalCircuit and FormalAssertion

This is useful for creating specialized versions of general circuits or adapting existing circuits for different use cases without re-verifying from scratch.

### DIFF
--- a/Clean/Circuit/Theorems.lean
+++ b/Clean/Circuit/Theorems.lean
@@ -631,3 +631,120 @@ def FormalAssertion.isGeneralFormalCircuit (F : Type) (Input : TypeMap) [Field F
       rintro _ _ _ _ ⟨ _, _ ⟩
       apply orig.completeness <;> trivial
   }
+
+-- Structural rules for FormalCircuit
+
+namespace FormalCircuit
+
+/--
+Strengthen the assumptions of a FormalCircuit.
+Given a circuit with assumptions `Assumptions` and a stronger assumption `StrongerAssumptions`,
+this creates a new FormalCircuit with the stronger assumptions.
+-/
+def strengthenAssumptions {F : Type} [Field F] {Input Output : TypeMap} [ProvableType Input] [ProvableType Output]
+    (circuit : FormalCircuit F Input Output)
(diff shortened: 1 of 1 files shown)


Clean/Circuit/Theorems.lean
```diff
@@ -631,3 +631,120 @@ def FormalAssertion.isGeneralFormalCircuit (F : Type) (Input : TypeMap) [Field F
...
+    circuit.soundness offset env input_var input h_eval (h_stronger input h_stronger_assumptions) h_constraints
+  completeness := fun offset env input_var h_uses_local input h_eval h_stronger_assumptions =>
+    circuit.completeness offset env input_var h_uses_local input h_eval (h_stronger input h_stronger_assumptions)
+
+/--
+Derive a new specification from a FormalCircuit.
+Given a circuit and a proof that the original spec implies a new spec under the assumptions,
+this creates a new FormalCircuit with the derived specification.
+-/
+def deriveSpec {F : Type} [Field F] {Input Output : TypeMap} [ProvableType Input] [ProvableType Output]
```
mitschabaude, line 662: I think `weakenSpec` would be a more intuitive name here. Same for general formal circuit

mitschabaude (commented):
In principle, I don't think we should add tooling/theorems without using it anywhere. Adding more code always adds a maintenance burden and build times, so there should be something worth the trade-off.

I can see how methods to transform assumptions and spec _could_ be useful. However, to me it makes more sense to be able to transform both assumptions + spec at the same time. A concrete usage example would provide clarity on the ideal API.

I believe you could address this by using the new methods (or variants of them) to
* transform the more general version of `Num2Bits` to the 'default' version
* introduce a more general version of `LessThan`, use your methods to transform it to the current version, and use the more general version to prove `LessThanEq`

PR #190, opened 2025-07-08
feat: introduce IsBool type for boolean constraints

yoichi-zksecurity:
## Summary
- Introduces `IsBool` predicate to improve readability of boolean operations in circuits
- Refactors all gate implementations and addition gadgets to use the new predicate
- Removes duplicate theorem `boolean_lt_2` after migration

## Motivation
This PR addresses the comment in #184 (https://github.com/Verified-zkEVM/clean/pull/184#discussion_r2192260516) about improving code readability for binary operations.

## Test plan
- [x] All files build successfully with `lake build`


Clean/Utils/Bool.lean
```diff
@@ -0,0 +1,111 @@
...
+    · simp only [hy0, IsBool]; right; ring
+    · simp only [hy1, IsBool]; left; ring
+  · simp only [hx1, one_mul, mul_one]
+    rcases hy with hy0 | hy1
+    · simp only [hy0, mul_zero, add_zero, IsBool]; left; ring
+    · simp only [hy1, mul_one, IsBool]; left; ring
+
+end IsBool
+
+end Clean
```
yoichi-zksecurity, line 111: needs a newline at the end

Clean/Examples/Add32Explicit.lean
```diff
@@ -26,28 +27,15 @@ example : ((circuit32 default).operations 0).SubcircuitsConsistent 0 :=
...
-  -- for 90% of the unfolding; and doesn't even need to unfold names like `Addition32Full.main` and `Addition8FullCarry.main`
-
-  -- TODO on the whole, which is better?
-
-  -- first version: using `circuit_norm`
-  -- dsimp only [circuit_norm, circuit32, Addition32Full.main, Addition8FullCarry.main, Gadgets.ByteTable]
-  -- simp only [subcircuit_norm, circuit_norm, Nat.reduceAdd, and_assoc]
-  -- simp only [Gadgets.ByteTable]
-
-  -- second version: using `ExplicitCircuit`
```
yoichi-zksecurity, line 50: don't remove these comments.

mitschabaude:
Can you find a name that doesn't need to be namespaced? And add the definition to circuit_norm so to not make proofs more cumbersome?

yoichi-zksecurity:
> add the definition to circuit_norm so to not make proofs more cumbersome?

Unsure. I saw some goals containing `something_really_big_that_goes_on = 0 ∨ something_really_big_that_goes_on = 1` so maybe it's better to keep `IsBool something_really_big_that_goes_on`.

yoichi-zksecurity:
I'm still in the middle of renaming.

> I guess try to preserve the simplicity of old proofs

Yes. I'm trying one strategy to add two facts `IsBool 0` and `IsBool 1` in the default simpset. If that's not enough I'll be unfolding `IsBool` explicitly in `simp (only) [...]`. Agreed that the proofs shouldn't get complicated for this change.

> Boolean.lean just has the purpose of proving IsBool

I also had a sense of duplicates, so I'll do something.

Clean/Gadgets/Boolean.lean
```diff
@@ -1,6 +1,124 @@
 import Clean.Circuit.Basic
 import Clean.Utils.Field
 
+namespace Clean
```
mitschabaude: this namespace shouldn't be here.

actually, I don't think the name `IsBool` needs to be namespaced either. It doesn't appear in lean core and mathlib: https://leanprover-community.github.io/mathlib4_docs/search.html?q=IsBool

seems moving it into a `Clean` namespace was just an LLM mistake?
can you rename back to `IsBool`? :D

yoichi-zksecurity: done.

yoichi-zksecurity:
I updated this to work on natural numbers, too.

Clean/Gadgets/Boolean.lean
```diff
@@ -1,6 +1,134 @@
...
+/-- A predicate stating that an element is boolean (0 or 1) for any type with 0 and 1 -/
+def IsBool {α : Type*} [Zero α] [One α] (x : α) : Prop := x = 0 ∨ x = 1
+
+/-- IsBool is decidable for types with decidable equality -/
+instance {α : Type*} [Zero α] [One α] [DecidableEq α] {x : α} : Decidable (IsBool x) :=
+  inferInstanceAs (Decidable (x = 0 ∨ x = 1))
+
+namespace IsBool
+
+@[simp]
```
mitschabaude: we generally don't tag with `simp`, since we have our own dedicated set of simp lemmas (`circuit_norm`)

Clean/Gadgets/Boolean.lean
```diff
@@ -1,6 +1,134 @@
...
+  constructor
+  · exact mul_sub_one
+  · intro h
+    have : x = 0 ∨ x - 1 = 0 := by
+      rw [mul_eq_zero] at h
+      exact h
+    rcases this with h0 | h1
+    · left; exact h0
+    · right
+      exact sub_eq_zero.mp h1
```
mitschabaude: I think the old proof was better:
```suggestion
/-- x is boolean iff x * (x - 1) = 0 -/
theorem iff_mul_sub_one {α : Type*} [Ring α] [NoZeroDivisors α] {x : α} :
    IsBool x ↔ x * (x - 1) = 0 := by
  rw [mul_eq_zero, sub_eq_zero, IsBool]
```

Clean/Gadgets/Boolean.lean
```diff
@@ -1,6 +1,134 @@
...
+
+/-- For natural numbers, if x is boolean then x < 2 -/
+theorem nat_lt_two {x : ℕ} (h : IsBool x) : x < 2 := by
+  rcases h with h0 | h1
+  · rw [h0]; norm_num
+  · rw [h1]; norm_num
+
+section FieldSpecific
```
mitschabaude: would remove this namespace. clean is field-specific in its entirety 😅

Clean/Gadgets/Boolean.lean
```diff
@@ -1,6 +1,134 @@
...
+  · rcases hy with hy0 | hy1
+    · simp [hx1, hy0, one_mul, mul_zero, add_sub_cancel, sub_self]
+    · simp [hx1, hy1, one_mul]
+
+section FieldOperations
+variable {p : ℕ} [Fact p.Prime]
+
+-- Field-specific operations can go here if needed
+
+end FieldOperations
```
mitschabaude: I think you can remove this

mitschabaude (approved):
very nice PR, just a few more comments

Clean/Gadgets/Boolean.lean
yoichi-zksecurity: I see. You were removing the last of these recently.

Issue #155, opened 2025-06-27
Formalize all of circomlib


mitschabaude:
[circomlib](https://github.com/iden3/circomlib) is highly stable and it's an essential public good powering a large share of existing zk apps.

formalizing _all of it_ would:
- do a good service for the ecosystem
- demonstrate the power and usefulness of clean
- demonstrate how much easier and more efficient it is to write formally correct circuits than to audit them over and over
- be a fun project that also outside contributors can easily jump into

idea: we could use an LLM to scaffold all the circuits

PR #195, opened 2025-07-09
MultiAND gate

pirapira:
Implements, specifies and verifies the MultiAND gate of circomlib.

A part of #155 .

This solves #201.


Clean/Circuit/Theorems.lean
```diff
@@ -8,6 +8,7 @@ import Clean.Circuit.Basic
 
 variable {F: Type} [Field F] {α β : Type}
 
+
```
pirapira: No need to change this.

Clean/Circomlib/Gates.lean
```diff
@@ -1,9 +1,27 @@
...
+
+/-- Helper to prove evaluation equality at any index -/
+lemma eval_index_helper {env : Environment (F p)} {input_var : Var (fields n) (F p)} {input : fields n (F p)}
+    (h_eval : input = eval env input_var) (idx : ℕ) (hidx : idx < n) :
+    input[idx] = Expression.eval env input_var[idx] := by
+  have : input[idx] = (eval env input_var)[idx] := by rw [h_eval]
+  simp only [ProvableType.eval_fields, Vector.getElem_map] at this
+  exact this
+
+end VectorSplitHelpers
```
pirapira: This looks general enough. Move this somewhere?

Clean/Utils/Vector.lean
```diff
@@ -232,6 +268,35 @@ theorem append_take_drop {v : Vector α (n + m)} :
...
+    by_cases idx < k
+    · simp only [Array.length_toList, List.get_eq_getElem, Array.getElem_toList,
+      Vector.getElem_toArray, Array.toList_extract, List.extract_eq_drop_take, tsub_zero,
+      List.drop_zero]
+      aesop
+    · simp only [Array.length_toList, List.get_eq_getElem, Array.getElem_toList,
+      Vector.getElem_toArray, Array.toList_extract, List.extract_eq_drop_take, tsub_zero,
+      List.drop_zero]
+      aesop
+
```
pirapira: A previous comment by @mitschabaude: it's better to formulate this lemma in terms of Vector take and drop.

If we keep this lemma around, people (and LLMs) will try to use it, and introduce lots of `toList`, so it's better to state the Vector version and just use it.

pirapira:
Draft because still cleaning up.

Clean/Circuit/Theorems.lean
pirapira: removed.

Clean/Circomlib/Gates.lean
pirapira: done.

Clean/Utils/Vector.lean
pirapira: Removed this. This was no longer used.

Clean/Circomlib/Gates.lean
```diff
@@ -12,6 +17,11 @@ https://github.com/iden3/circomlib/blob/master/circuits/gates.circom
 namespace Circomlib
 variable {p : ℕ} [Fact p.Prime]
 
+-- Import the moved lemmas
```
mitschabaude: nit: weird comment

Clean/Circomlib/Gates.lean
```diff
@@ -247,16 +258,811 @@ def circuit : FormalCircuit (F p) fieldPair field where
...
+    let input2 : Vector (Expression (F p)) n2 := input.drop n1 |>.cast (by omega)
+
+    let out1 ← main input1
+    let out2 ← main input2
+
+    AND.circuit.main (out1, out2)
+
+
+
+
```
mitschabaude: nit: can we get rid of extra newlines? IMO, one is enough everywhere :D

Clean/Circomlib/Gates.lean
```diff
@@ -247,16 +258,811 @@ def circuit : FormalCircuit (F p) fieldPair field where
...
+    simp only [Array.size_extract]
+    have h1 : i < n1 := hi
+    have h2 : input.toArray.size = n := by simp [Vector.size_toArray]
+    rw [h2, min_eq_left h_n1_lt]
+    exact h1) =
+                  input.toArray[i]'(by
+                    have h1 : i < n1 := hi
+                    have h2 : input.toArray.size = n := by simp [Vector.size_toArray]
+                    rw [h2]
+                    omega) := by
```
mitschabaude: I'm not a fan of adding many spaces to align equality statements, I think this would be better formatting:
```suggestion
  have h_extract : (input.toArray.extract 0 n1)[i]'(by
      simp only [Array.size_extract]
      have h1 : i < n1 := hi
      have h2 : input.toArray.size = n := by simp [Vector.size_toArray]
      rw [h2, min_eq_left h_n1_lt]
      exact h1) =
    input.toArray[i]'(by
      have h1 : i < n1 := hi
      have h2 : input.toArray.size = n := by simp [Vector.size_toArray]
      rw [h2]
      omega) := by
```

Clean/Circomlib/Gates.lean
```diff
@@ -247,16 +258,811 @@ def circuit : FormalCircuit (F p) fieldPair field where
...
+  localLength _ := n - 1
+  localLength_eq := localLength_eq n
+  subcircuitsConsistent := subcircuitsConsistent n
+
+  Assumptions := Assumptions n
+  Spec := Spec n
+
+  -- TODO: Future work - update the FormalCircuit Soundness definition to have
+  -- h_env : input = eval env input_var instead of eval env input_var = input
+  -- for more natural theorem statements
```
mitschabaude: I disagree, because we always want to rewrite `eval env input_var` to `input`, instead of the other way round.
If the statement was reversed, it can no longer be directly passed to `simp`, and `simp_all` won't make effective use of it. the direction of a statement should be the direction that we want to rewrite in

Clean/Gadgets/Boolean.lean
```diff
@@ -103,6 +123,58 @@ theorem nor_is_bool {α : Type*} [Ring α] {x y : α} (hx : IsBool x) (hy : IsBo
...
+/-- If a is boolean (0 or 1), then a &&& 1 = a -/
+theorem land_one_of_IsBool (a : ℕ) (h : IsBool a) : a &&& 1 = a := by
+  rcases h with h0 | h1
+  · rw [h0]; norm_num
+  · rw [h1]; norm_num
+
+/-- If a is boolean (0 or 1), then 1 &&& a = a -/
+theorem one_land_of_IsBool (a : ℕ) (h : IsBool a) : 1 &&& a = a := by
+  rw [Nat.land_comm]
+  exact land_one_of_IsBool a h
```
mitschabaude, line 135: do we really need this as its own theorem?

Clean/Gadgets/Boolean.lean
```diff
@@ -103,6 +123,58 @@ theorem nor_is_bool {α : Type*} [Ring α] {x y : α} (hx : IsBool x) (hy : IsBo
...
+
+/-- If x is boolean in F p, then x.val is boolean as a natural number -/
+theorem val_of_IsBool {p : ℕ} [Fact p.Prime] {x : F p} (h : IsBool x) : IsBool x.val := by
+  rcases h with h0 | h1
+  · rw [h0]; simp only [ZMod.val_zero]; exact zero
+  · rw [h1]; simp only [ZMod.val_one]; exact one
```
mitschabaude, line 141: this is great, an `iff` version of this could have been useful for me a couple of times already

Clean/Gadgets/Rotation64/Rotation64Bytes.lean
```diff
@@ -5,7 +5,6 @@ import Clean.Utils.Primes
 namespace Gadgets.Rotation64Bytes
 variable {p : ℕ} [Fact p.Prime]
 
-open Bitwise (rotRight64)
```
mitschabaude, line 8: I appreciate all these clean up changes!

Clean/Utils/Vector.lean
```diff
@@ -12,6 +12,42 @@ def fromList (l: List α) : Vector α l.length := ⟨ .mk l, rfl ⟩
...
+        simp only [List.size_toArray] at h
+        -- Now we know lst has length 1, so it must be [x] for some x
+        match lst with
+        | [] => simp at h
+        | [x] =>
+          -- Goal: v.toList = [v.get 0]
+          -- v.toList = arr.toList = lst = [x]
+          -- v.get 0 = arr[0] = lst[0] = x
+          rfl
+        | _ :: _ :: _ => simp [List.length] at h
```
mitschabaude, line 34: recursive match lets you express the same proof succinctly:
```suggestion
  match v with | ⟨ ⟨ _ :: [] ⟩ , _ ⟩ => rfl
```

Clean/Utils/Vector.lean
```diff
@@ -12,6 +12,42 @@ def fromList (l: List α) : Vector α l.length := ⟨ .mk l, rfl ⟩
...
+  cases v using Vector.casesOn with
+  | mk arr h =>
+      cases arr using Array.casesOn with
+      | mk lst =>
+        simp only [List.size_toArray] at h
+        match lst with
+        | [] => simp at h
+        | [_] => simp [List.length] at h
+        | [x, y] => rfl
+        | _ :: _ :: _ :: _ => simp [List.length] at h
```
mitschabaude, line 49:
```suggestion
  match v with | ⟨ ⟨ _ :: _ :: [] ⟩ , _ ⟩ => rfl
```

Clean/Circomlib/Gates.lean
mitschabaude: I think the problem is just that you started using the opposite direction somewhere in this file? You should probably fix that

Clean/Circomlib/Gates.lean
```diff
@@ -247,16 +258,811 @@ def circuit : FormalCircuit (F p) fieldPair field where
...
+      let input1 : fields n1 (F p) := input.take n1 |>.cast (by simp only [Nat.min_def, n1]; split <;> omega)
+      let input2 : fields n2 (F p) := input.drop n1 |>.cast (by omega)
+      let input_var1 : Var (fields n1) (F p) := input_var.take n1 |>.cast (by simp only [Nat.min_def, n1]; split <;> omega)
+      let input_var2 : Var (fields n2) (F p) := input_var.drop n1 |>.cast (by omega)
+      have h_eval1 : input1 = eval env input_var1 := by
+        simp only [input_var1, input1]
+        apply Vector.ext
+        intro i hi
+        simp only [ProvableType.eval_fields, Vector.getElem_map, Vector.getElem_cast, Vector.getElem_take]
+        exact eval_index_helper h_env i (by omega)
```
mitschabaude: don't need `eval_index_helper`
```suggestion
        simp only [h_env, ProvableType.eval_fields, Vector.getElem_map, Vector.getElem_cast, Vector.getElem_take]
```

Clean/Circomlib/Gates.lean
```diff
@@ -247,16 +258,811 @@ def circuit : FormalCircuit (F p) fieldPair field where
...
+
+      have h_eval2 : input2 = eval env input_var2 := by
+        simp only [input_var2, input2]
+        apply Vector.ext
+        intro i hi
+        simp only [ProvableType.eval_fields, Vector.getElem_map, Vector.getElem_cast, Vector.getElem_drop]
+        exact eval_index_helper h_env (n1 + i) (by omega)
```
mitschabaude:
```suggestion
        simp only [h_env, ProvableType.eval_fields, Vector.getElem_map, Vector.getElem_cast, Vector.getElem_drop]
```

Clean/Circomlib/Gates.lean
```diff
@@ -247,16 +258,811 @@ def circuit : FormalCircuit (F p) fieldPair field where
...
+
+      have h_eval1 : input1 = eval env input_var1 := by
+        simp only [input_var1, input1]
+        apply Vector.ext
+        intro i hi
+        -- Need to show: input[i] = (eval env (Vector.cast _ (Vector.take input_var n1)))[i]
+        simp only [ProvableType.eval_fields, Vector.getElem_map, Vector.getElem_cast, Vector.getElem_take]
+        have hi' : i < n1 := hi
+        have hi'' : i < m + 3 := by omega
+        exact eval_index_helper h_env i (by omega)
```
mitschabaude:
```suggestion
        simp only [h_env, ProvableType.eval_fields, Vector.getElem_map, Vector.getElem_cast, Vector.getElem_take]
```

Clean/Circuit/Theorems.lean
```diff
@@ -631,3 +631,11 @@ def FormalAssertion.isGeneralFormalCircuit (F : Type) (Input : TypeMap) [Field F
...
       apply orig.completeness <;> trivial
   }
+
+/-- Helper to prove evaluation equality at any index -/
+lemma eval_index_helper {F : Type} [Field F] {n : ℕ} {env : Environment F} {input_var : Var (fields n) F} {input : fields n F}
+    (h_eval : input = eval env input_var) (idx : ℕ) (hidx : idx < n) :
+    input[idx] = Expression.eval env input_var[idx] := by
+  have : input[idx] = (eval env input_var)[idx] := by rw [h_eval]
+  simp only [ProvableType.eval_fields, Vector.getElem_map] at this
+  exact this
```
mitschabaude: I don't think this theorem is needed, since `simp only [h_eval, circuit_norm]` already proves it. see comments before for how to replace the existing invocations

Clean/Circomlib/Gates.lean
```diff
@@ -247,16 +258,811 @@ def circuit : FormalCircuit (F p) fieldPair field where
...
+        out <== and1.out;
+    } else {
+        and2 = AND();
+        var n1 = n\2;
+        var n2 = n-n\2;
+        ands[0] = MultiAND(n1);
+        ands[1] = MultiAND(n2);
+        var i;
+        for (i=0; i<n1; i++) ands[0].in[i] <== in[i];
+        for (i=0; i<n2; i++) ands[1].in[i] <== in[n1+i];
```
mitschabaude: unfortunately, these comments mess up code highlighting for me, and it can be fixed by adding spaces around `<`
```suggestion
        for (i=0; i < n1; i++) ands[0].in[i] <== in[i];
        for (i=0; i < n2; i++) ands[1].in[i] <== in[n1+i];
```

Clean/Circomlib/Gates.lean
pirapira: > because we always want to rewrite eval env input_var to input, instead of the other way round.

I noticed this a while ago. It feels odd. Usually `subst input` would be my first move for `input = something complicated`.

I will drop this comment.

pirapira: I frequently apply `rw [\l h_env]`. So probably I don't know the idiomatic style here. After knowing it, I might and might not like the idiomatic style. All that after knowing it.

pirapira: I usually wanted to replace `input` away and wanted to see `eval env ...` in the goal. That sounds more stable because we don't rely on the luck of seeing the exact `eval env ...` form.

pirapira: I will experiment.

Clean/Circomlib/Gates.lean
pirapira: I'll do a separate run of removing more of these.

Clean/Circomlib/Gates.lean
pirapira: But the experiment will be offline.

Clean/Circomlib/Gates.lean
pirapira: yes, it works.

I also see that we sometimes want to rewrite with

```
input = eval env input_var
```
from left to right.

Clean/Circuit/Theorems.lean
pirapira: Removed.

Clean/Circomlib/Gates.lean
mitschabaude: yeah I guess in principle, moving the other way works as well. the constraints are written in terms of `eval env input_var` (or variants of that), while the spec and assumption talk about `input`, and these have to be reconciled one way or another.

the thing is that `input` is the simpler of the two terms, and therefore leads to easier-to-understand intermediate goals. so we have optimized for rewriting everything into `input`. if you check existing proofs, this "input_var elimination" is typically achieved simply by `simp only [circuit_norm, h_input]`, or `simp_all only [circuit_norm]`, or similar

Clean/Gadgets/Boolean.lean
pirapira: Yes, also a good candidate to put in the circuit_norm simpset.

`land_one_of_IsBool.symm` doesn't give the same statement.

Clean/Circomlib/Gates.lean
mitschabaude: I can see your point for `subst input`, since it always works, and so gives a more mechanical proof strategy. This didn't occur to me back when I started writing proofs, I just wanted to see a "nice" mathematical statement which used intuitive variable names for each element participating in the constraint equations

Clean/Utils/Vector.lean
mitschabaude: @pirapira I noticed you didn't commit this suggestion, just making sure you saw it (no worries about making the change, it's cosmetic)

Clean/Circomlib/Gates.lean
mitschabaude: > I also see that we sometimes want to rewrite with
> 
> ```
> input = eval env input_var
> ```
> 
> from left to right.

I think that's just because of the way the helper lemmas in this file are formulated (in the reverse direction than usual)

Clean/Gadgets/Boolean.lean
mitschabaude: it doesn't, but the statement is a trivial combination of `land_one_of_IsBool` and `land_comm`

mitschabaude: the question is, do we want a separate theorem if it
* is not used often
* where used, it could be replaced by two other existing theorems

I think the question is analogous to: when do we extract program logic into a helper function?

My answer usually is:
* either when the logic is used sufficiently often and we can avoid duplicating it over and over
* or when the amount of complexity that is contained inside the new method, and thereby removed from other code, is sufficiently big (here, this is small, since it's just one lemma replacing an easy combination of two other lemmas)

pirapira: > the statement is a trivial combination of land_one_of_IsBool and land_comm

Yes, and we don't want to put `land_comm` in the simpset.

That's why we have both

@[simp]
theorem [Nat.zero_and](https://leanprover-community.github.io/mathlib4_docs/Init/Data/Nat/Bitwise/Lemmas.html#Nat.zero_and)(x : [Nat](https://leanprover-community.github.io/mathlib4_docs/Init/Prelude.html#Nat)) :
0 [&&&](https://leanprover-community.github.io/mathlib4_docs/Init/Prelude.html#HAnd.hAnd) x [=](https://leanprover-community.github.io/mathlib4_docs/Init/Prelude.html#Eq) 0

@[simp]
theorem [Nat.and_zero](https://leanprover-community.github.io/mathlib4_docs/Init/Data/Nat/Bitwise/Lemmas.html#Nat.and_zero)(x : [Nat](https://leanprover-community.github.io/mathlib4_docs/Init/Prelude.html#Nat)) :
x [&&&](https://leanprover-community.github.io/mathlib4_docs/Init/Prelude.html#HAnd.hAnd) 0 [=](https://leanprover-community.github.io/mathlib4_docs/Init/Prelude.html#Eq) 0

in `simp`.

And the `simp` doesn't contain

theorem [Nat.and_comm](https://leanprover-community.github.io/mathlib4_docs/Init/Data/Nat/Bitwise/Lemmas.html#Nat.and_comm)(x y : [Nat](https://leanprover-community.github.io/mathlib4_docs/Init/Prelude.html#Nat)) :
x [&&&](https://leanprover-community.github.io/mathlib4_docs/Init/Prelude.html#HAnd.hAnd) y [=](https://leanprover-community.github.io/mathlib4_docs/Init/Prelude.html#Eq) y [&&&](https://leanprover-community.github.io/mathlib4_docs/Init/Prelude.html#HAnd.hAnd) x

Something you can add to the simpset is worth stating.

pirapira: > My answer usually is:

I can play with some approaches:

* (a) always, given the statement looks simple enough ([example](https://www.isa-afp.org/sessions/aot/#AOT_PLM.html#AOT_PLM.cqt-further:1|fact))
* (b) whenever the statement in the simpset makes some progress in the simplification somewhere (some Isabelle / ACL2 people do this)
* (c) you design a struct type so that it aximatizes a concept, and ... (but I'm not that disciplined).

mitschabaude: > Something you can add to the simpset is worth stating.

this does make sense to me!

PR #196, opened 2025-07-10
refactor: remove redundant Vector instance and add ProvableVector utilities

pirapira:
## Summary
- Removed redundant `HasAssignEq` instance for `Vector` since it's already covered by `ProvableType`
- Updated `Inputs` structures in `Mux1.lean` to use `ProvableVector` directly instead of `Vector`
- Added `ProvableVector.provable_map` utility function to map over ProvableVectors while preserving the type

## Test plan
- [x] `lake build` succeeds

### DIFF
--- a/Clean/Circomlib/Mux1.lean
+++ b/Clean/Circomlib/Mux1.lean
@@ -38,7 +38,7 @@ def main (n: ℕ) (input : Var (Inputs n) (F p)) := do
   let { c, s } := input
 
   -- Witness and constrain output vector
-  let out : Var (ProvableVector field n) (F p) <== c.map fun (c0, c1) =>
+  let out <== c.provable_map field fun (c0, c1) =>
     (c1 - c0) * s + c0
   return out
 
--- a/Clean/Circomlib/Mux1.lean
+++ b/Clean/Circomlib/Mux1.lean
@@ -38,7 +38,7 @@ def main (n: ℕ) (input : Var (Inputs n) (F p)) := do
   let { c, s } := input
 
   -- Witness and constrain output vector
-  let out <== c.map fun (c0, c1) =>
+  let out : Var (ProvableVector field n) (F p) <== c.map fun (c0, c1) =>
     (c1 - c0) * s + c0
   return out
 
--- a/Clean/Circomlib/Mux1.lean
+++ b/Clean/Circomlib/Mux1.lean
@@ -16,8 +16,8 @@ https://github.com/iden3/circomlib/blob/master/circuits/mux1.circom
 namespace MultiMux1
 
 structure Inputs (n : ℕ) (F : Type) where
-  c : Vector (F × F) n  -- n pairs of constants
-  s : F                 -- selector
+  c : ProvableVector (ProvablePair field field) n F  -- n pairs of constants
+  s : F                                               -- selector
 
 instance {n : ℕ} : ProvableStruct (Inputs n) where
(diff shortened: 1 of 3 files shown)


Clean/Circomlib/Mux1.lean
```diff
@@ -16,8 +16,8 @@ https://github.com/iden3/circomlib/blob/master/circuits/mux1.circom
 namespace MultiMux1
 
 structure Inputs (n : ℕ) (F : Type) where
-  c : Vector (F × F) n  -- n pairs of constants
-  s : F                 -- selector
+  c : ProvableVector (ProvablePair field field) n F  -- n pairs of constants
+  s : F                                               -- selector
```
mitschabaude, line 20: to me this is no improvement. it'd be nicer for users of clean if they could just use standard types like `Vector` and `Prod` in their circuits, instead of having to learn about clean-specific versions of them (`ProvableVector` / `ProvablePair`)

Clean/Circomlib/Mux1.lean
```diff
@@ -38,7 +38,7 @@ def main (n: ℕ) (input : Var (Inputs n) (F p)) := do
   let { c, s } := input
 
   -- Witness and constrain output vector
-  let out <== c.map fun (c0, c1) =>
+  let out <== c.provable_map field fun (c0, c1) =>
```
mitschabaude, line 41: this isn't an improvement, it's more verbose and is using a custom method from clean for doing the standard `map` operation

Clean/Circuit/Provable.lean
```diff
@@ -169,6 +169,12 @@ def ProvableVector (α: TypeMap) (n: ℕ) := fun F => Vector (α F) n
 @[reducible]
 def fields (n: ℕ) := fun F => Vector F n
 
+-- beta is the ProvableType of the target, and it's explicit because omitting this has caused type inference failures
+namespace ProvableVector
+def provable_map {α : TypeMap} {n : ℕ} {F : Type} (x : ProvableVector α n F) (β : TypeMap) (f : α F → β F) : ProvableVector β n F :=
```
mitschabaude, line 174: do we really need this? if yes it should be `provableMap`

Clean/Gadgets/Equality.lean
```diff
@@ -156,13 +156,6 @@ instance {F : Type} [Field F] {α : TypeMap} [ProvableType α] :
     witness === rhs
     return witness
 
-instance {F : Type} [Field F] {n : ℕ} : HasAssignEq (Vector (Expression F) n) F where
-  assignEq vals := do
-    vals.mapM fun v => do
-      let witness ← witnessField fun env => v.eval env
-      witness === v
-      return witness
```
mitschabaude, line 164: from reviewing the PR, it seems to me that this instance was useful, because vectors are not automatically inferred to be `ProvableType`s everywhere

you can make it less duplicative by reusing the instance for `ProvableType`:
```lean
instance {F : Type} [Field F] {n : ℕ} : HasAssignEq (Vector (Expression F) n) F :=
  inferInstanceAs (HasAssignEq (fields n (Expression F)) F)
```

mitschabaude (commented):
not really convinced of the benefit of this PR, see comments

pirapira:
I had forgotten about this PR.

After using `Clean` for a while, the biggest pain point for me is something else. It's the need to add type-annotatations ` : ProvableVector _ _` frequently. Somehow this PR is making it worse.

PR #212, opened 2025-07-14
Complete BLAKE3 ApplyRounds circuit with FormalCircuit combinators

pirapira:
## Summary

This PR implements a complete BLAKE3 ApplyRounds circuit using new FormalCircuit combinators, with full soundness and completeness proofs.

### Key Contributions

**FormalCircuit Combinators (`Clean/Circuit/StructuralLemmas.lean`)**:
- `FormalCircuit.concat`: Sequential composition combinator for chaining circuits
- `FormalCircuit.weakenSpec`: Specification adaptation combinator  
- Complete soundness and completeness proofs for both combinators

**BLAKE3 ApplyRounds Circuit (`Clean/Gadgets/BLAKE3/ApplyRounds.lean`)**:
- Modular circuit construction using combinators:
  - `roundWithPermute`: Single round + permutation
  - `twoRoundsWithPermute`, `fourRoundsWithPermute`, `sixRoundsWithPermute`: Multi-round compositions
  - `sevenRoundsApplyStyle`: Complete 7-round BLAKE3 compression
- Full soundness & complete proofs

## Test plan

- `lake build` succeeds
- All proofs complete without sorry

🤖 Generated with [Claude Code](https://claude.ai/code)

Issue #123, opened 2025-06-16
[BLAKE3] Compression function: Initialisation and permutation loop


mitschabaude:
- [x] Implement and verify a circuit for `finalStateUpdate` (this has been long done)
- [x] Implement and verify a circuit for `compress`

PR #216, opened 2025-07-16
Implement BLAKE3 compress circuit

pirapira:
## Summary

This PR implements the BLAKE3 compress circuit, completing the implementation for issue #123. This is based on PR #212 which implements the ApplyRounds circuit.

- Implements `Clean.Gadgets.BLAKE3.Compress` that chains ApplyRounds and FinalStateUpdate circuits
- Complete soundness and completeness proofs
- Added to default lake build

## Test plan

- `lake build` succeeds
- All proofs complete without sorry


Clean/Gadgets/BLAKE3/Compress.lean
```diff
@@ -0,0 +1,77 @@
...
+  ApplyRounds.circuit.Assumptions (F := F p) = ApplyRounds.Assumptions := rfl
+
+lemma ApplyRounds.circuit_spec_is :
+  ApplyRounds.circuit.Spec (F := F p) = ApplyRounds.Spec := rfl
+
+lemma FinalStateUpdate.circuit_assumptions_is :
+  FinalStateUpdate.circuit.Assumptions (F := F p) = FinalStateUpdate.Assumptions := rfl
+
+lemma FinalStateUpdate.circuit_spec_is :
+  FinalStateUpdate.circuit.Spec (F := F p) = FinalStateUpdate.Spec := rfl
```
pirapira: Move these to their respective files.

Clean/Gadgets/BLAKE3/FinalStateUpdate.lean
```diff
@@ -140,4 +140,10 @@ def circuit : FormalCircuit (F p) Inputs BLAKE3State := {
   elaborated with Assumptions, Spec, soundness, completeness
 }
 
+lemma circuit_assumptions_is :
+  circuit.Assumptions (F := F p) = Assumptions := rfl
```
mitschabaude: instead of this you can usually just add `ApplyRounds.circuit` to the simp set. did you add it for efficiency?

pirapira: For efficiency and separating responsibilities. Callers care about `Assumptions` and `Spec` but not about the operation list.

pirapira: In the caller proof I saw `FinalStateUpdate.circuit.Assumptions`. I thought "sure you can see it" and added this lemma to simplify this term. Certain other things aren't supposed to appear in the goal state.

This is the first time I do this because I wrote the caller proof by hand from scratch, so I was more conscious.

pirapira: ^ is just one of my proof-writing modes.

mitschabaude: > For efficiency and separating responsibilities. Callers care about Assumptions and Spec but not about the operation list.

got it, makes sense!

fyi you can also do this inline with something like
```lean
rw [show FinalStateUpdate.circuit.Assumptions = FinalStateUpdate.Assumptions from rfl]
```

pirapira: I can go in the other direction and add `circuit_assumptions_is ` in `circuit_norm`. Maybe there's a way to make it happen everywhere for `FormalCircuit`s.

pirapira: There's @[simps] attribute that can be used for this.

pirapira: `@[simps]` didn't work. It's for the default simp set. To use it, I need to drop `only` from `simp only`. When I do that the `simp` timeouts.

pirapira: Instead, I added
```
@[circuit_norm]
lemma circuit_assumptions_is :
  circuit.Assumptions (F := F p) = Assumptions := rfl
```
in `circuit_norm`.

I think people don't need to unfold `circuit` in most cases, and they can often avoid seeing big struct literals.

pirapira:
```
@[simps! (config := {isSimp := false, attrs := [`circuit_norm]})]
```
did the trick. This is a common enough use case.

Clean/Gadgets/BLAKE3/ApplyRounds.lean
```diff
@@ -734,9 +734,18 @@ theorem completeness : Completeness (F p) elaborated Assumptions := by
...
   elaborated with Assumptions, Spec, soundness, completeness
 }
 
+@[circuit_norm]
+lemma circuit_assumptions_is :
+  circuit.Assumptions (F := F p) = Assumptions := rfl
+
+@[circuit_norm]
+lemma circuit_spec_is :
+  circuit.Spec (F := F p) = Spec := rfl
```
mitschabaude: these should definitely not go into `circuit_norm`, they are used in a single other proof and can just be added there directly

in general, anything about specific gadgets should be considered in user land and should not modify the core library. the only exception are core, very frequently used gadgets like `assertBool` or `Equality.circuit`

Clean/Gadgets/BLAKE3/FinalStateUpdate.lean
```diff
@@ -135,6 +135,12 @@ theorem completeness : Completeness (F p) elaborated Assumptions := by
...
   simp_all only [gt_iff_lt, forall_const, and_self]
 
+/-
+The `@[simps! ...] attributes add equations to `circuit_norm` so that `FinalStateUpdate.circuit.{Assumption, (or any other field)}`
+gets replaced with the definition. This leads to much less `simp only [FinalStateUpdate.circuit]` and we avoid seeing a big struct
+literal in the goal state.
+-/
+@[simps! (config := {isSimp := false, attrs := [`circuit_norm]})]
 def circuit : FormalCircuit (F p) Inputs BLAKE3State := {
   elaborated with Assumptions, Spec, soundness, completeness
```
mitschabaude: should not be added to `circuit_norm`

Clean/Gadgets/BLAKE3/FinalStateUpdate.lean
mitschabaude: interesting trick, nice to know! I don't think these should be added to `circuit_norm` for random gadgets though

Clean/Gadgets/BLAKE3/ApplyRounds.lean
pirapira: Here I'm making a strong statement that nobody ever needs to see `ApplyRounds.circuit.Spec` in the goal state, and they should see `ApplyRounds.Spec` instead.

Clean/Gadgets/BLAKE3/FinalStateUpdate.lean
pirapira: Here I'm also making a strong statement that there's no use seeing `FinalStateUpdate.circuit.Assumptions` in the goal state. Instead people should see `FinalStateUpdate.Assumptions`.

I also think all proofs should look like `simp only [circuit_norm]` without `FinalStateUpdate.circuit` etc.

mitschabaude:
fwiw, I don't think adding something like `ApplyRounds.circuit` to a simp invocation is too bad, even if you just want to unfold (say) `ApplyRounds.circuit.Assumptions`. `simp` will do the structure field reduction and, after first expanding `ApplyRounds.circuit`, find and get the `.Assumptions` attribute from the expanded circuit

Clean/Gadgets/BLAKE3/FinalStateUpdate.lean
pirapira: I believe this should be done for every gadget. Because there's no benefit in seeing `Gadget.circuit.Assumptions` ever in the goal state. Instead people should see `Gadget.Assumptions`. This makes the goal state shorter and simpler in all cases.

mitschabaude: but it has a cost, no? the theorem will be processed in every single gadget proof ever, even it only relates to a very specific gadget that's unrelated to almost all usages of the library. and the downside of unpacking the `.circuit` struct seems very small to me

pirapira:
> also will be accessed via a struct field and simp will reduce those as well.

It does not happen when `ApplyRounds.circuit` appears as an argument of a function (which is not a projection).

mitschabaude:
> > also, by analogy, we'd have to do that for every particular gadget
> 
> Yes, we need to. There's no merit in leaving the choice to the user of the gadgets.

I think there is merit: we improve efficiency of simp by keeping the simp set at a reasonable size

pirapira:
> an ever growing list of lemmas (one for each circuit) that all do the same thing

That's the way to go.

mitschabaude:
besides that, keeping subcircuit data opaque sometimes does have merit sometimes, it lets you stage a proof more incrementally. note that `FinalStateUpdate.circuit.Assumptions` is not necessarily defined in a separate method `FinalStateUpdate.Assumptions`. it could be defined directly on `FinalStateUpdate` and unfold to something large, even when you don't need it

pirapira:
simproc: the dynamically generated lemma needs to be type-checked. Is that really faster than finding matches in the simp set of already type-checked lemmas? I don't look at it right now.

mitschabaude:
maybe we can merge this PR without the simp set additions and launch a more thorough experiment towards simplifying proofs in a new PR?

pirapira:
Here's the version without the `simps`.

It seems like `simps` technique is different from the current intended usage of `circuit_norm`.

Clean/Gadgets/BLAKE3/ApplyRounds.lean
```diff
@@ -733,9 +733,8 @@ theorem completeness : Completeness (F p) elaborated Assumptions := by
...
   · apply h_message_normalized
 
-
+-- Unfortunately @[simps! (config := {isSimp := false, attrs := [`circuit_norm]})] timeouts.
+-- Therefore I had to add simplification rules `circuit_assumptions_is` and `circuit_spec_is` manually.
 def circuit : FormalCircuit (F p) Inputs BLAKE3State := {
   elaborated with Assumptions, Spec, soundness, completeness
 }
-
-end Gadgets.BLAKE3.ApplyRounds
```
mitschabaude: unintended change?

PR #220, opened 2025-07-17
Fix #217: Remove big struct literals from goal state by controlling subcircuit unfold

### LINKED ISSUE #217: Controlling subcircuit unfold
### Problem

At the end of [line 114 of Round.lean](https://github.com/Verified-zkEVM/clean/blob/7514d04092ca798185fd607c95035456769fa446/Clean/Gadgets/BLAKE3/Round.lean#L114), I see a big goal containing an unfolded struct literal for a `Subcircuit`:

```
henv : env.UsesLocalWitnessesCompleteness i0
  [Operation.subcircuit
      {
        ops :=
          Operations.toFlat (G.main 0 4 8 12 { state := state_var, x := message_var[0], y := message_var[1] } i0).2,
        Soundness := fun env ↦
          (eval env { state := state_var, x := message_var[0], y := message_var[1] }).state.Normalized ∧
              (eval env { state := state_var, x := message_var[0], y := message_var[1] }).x.Normalized ∧
                (eval env { state := state_var, x := message_var[0], y := message_var[1] }).y.Normalized →
```

This goal contains lots of unnecessary information because `env.UsesLocalWitnessesCompleteness` only uses `.UsesLocalWitnesses` of the subcircuit.

Even if this goal is never shown, this simplification path contributes to earlier simplification time out. This issue keeps track of tweaking `circuit_norm` so that the default path doesn't go through this state.

### Strategy

`circuit_norm` gets a new simplification

```
env.UsesLocalWitnessCompleteness _ (subcircuit (c : FormalCircuit) _ :: _)
```
into
```
c.UsesLocalWitness env /\ env.UsesLocalWitnessCompleteness _ 
```
This prevents most fields of the `Subcircuit` from appearing in the goal state at all in this scenario.

### Trade-off

* This complicates the source code a bit because of the three new simplification rules (for `GeneralFormalCircuit` `FormalCircuit` and `FormalAssertion`).
* This simplifies the source code whenever subcircuits are used.
* This simplifies the goal state during development. Better ergonomics, shorter build time, and more room till simplification timeouts.

This one is rather fortunate because only three additional equations cover all `toSubcircuit` variations.

pirapira:
This PR fixes #217 by removing the `subcircuit_norm` simp attribute and replacing it with specific simplification lemmas to prevent big struct literals from appearing in goal states.

## Motivation (from #217)

The problem was that during proofs, big unfolded struct literals for `Subcircuit` were appearing in goal states, containing lots of unnecessary information. For example, when `env.UsesLocalWitnessesCompleteness` only needs the `.UsesLocalWitnesses` field of a subcircuit, the entire subcircuit structure would be unfolded, showing all fields including `ops`, `Soundness`, etc.

This caused:
- Cluttered goal states that were hard to read during development
- Simplification timeouts due to processing unnecessary information
- Longer build times
- Poor ergonomics when working with subcircuits

## Solution

Instead of using `subcircuit_norm` which unfolds entire subcircuit structures, this PR implements the strategy outlined in #217: adding specific simplification lemmas that prevent unnecessary fields from appearing in goal states. This gives us fine-grained control over what gets simplified.

### 1. Added specific simplification lemmas for subcircuit fields:
- `toSubcircuit_usesLocalWitnesses` - simplifies only the UsesLocalWitnesses field
- `toSubcircuit_localLength` - simplifies only the localLength field
- `toSubcircuit_soundness` - simplifies only the Soundness field
- `toSubcircuit_completeness` - simplifies only the Completeness field  
- `toSubcircuit_ops` - simplifies only the ops field

These lemmas are added for all three circuit types as mentioned in #217: `FormalCircuit`, `GeneralFormalCircuit`, and `FormalAssertion`.

### 2. Removed `subcircuit_norm` attribute and all its uses:
- Removed the attribute definition from `SimpGadget.lean`
- Removed all attribute applications from `Subcircuit.lean`
- Systematically removed `subcircuit_norm` usage from ~30 files across the codebase

### 3. Fixed various build issues:
- Added `@[simps!]` to `Equality.circuit` to generate projection lemmas
- Fixed redundant simp calls in AliasCheck, CompConstant, Fibonacci32, and Bits
- Updated proofs that relied on the old subcircuit structure

## Performance Improvement

Benchmark by running `rm -rf .lake/build; time lake build --quiet` shows significant improvement:

**After this PR:**
```
lake build --quiet  361.43s user 41.40s system 455% cpu 1:28.41 total
```

**Before (on main branch):**
```
lake build --quiet  494.72s user 49.74s system 474% cpu 1:54.62 total
```

This represents a **~27% reduction in build time**.

## Benefits (as outlined in #217)

- Simplifies the goal state during development - no more big struct literals
- Better ergonomics when working with subcircuits
- Shorter build time and more room till simplification timeouts
- Simplifies source code whenever subcircuits are used
- Only the necessary fields appear in goal states

The trade-off mentioned in #217 (complicating source code with three new simplification rules) is outweighed by the benefits of cleaner goal states and better performance.


Clean/Circuit/Subcircuit.lean
```diff
@@ -345,11 +345,174 @@ theorem Circuit.subcircuit_computableWitnesses (circuit: FormalCircuit F β α)
...
+
+/--
+Simplifies ops for FormalAssertion.toSubcircuit to avoid unfolding the entire subcircuit structure.
+-/
+@[circuit_norm]
+theorem FormalAssertion.toSubcircuit_ops
+    {F : Type} [Field F] {Input : TypeMap} [ProvableType Input]
+    (circuit : FormalAssertion F Input) (n : ℕ) (input_var : Var Input F) :
+    (circuit.toSubcircuit n input_var).ops = (circuit.main input_var |>.operations n).toFlat := by
+  rfl
```
mitschabaude: These three should be removed from `circuit_norm`, we actually don't ever want to see subcircuit operations (or other internal details from subcircuits)

You can even remove the lemmas!

Clean/Gadgets/BLAKE3/Round.lean
```diff
@@ -105,13 +105,7 @@ theorem soundness : Soundness (F p) elaborated Assumptions Spec := by
...
   rintro i0 env ⟨state_var, message_var⟩ henv ⟨state, message⟩ h_input h_normalized
 
-  dsimp only [elaborated, main, Fin.isValue, G.circuit, Circuit.pure_def, Circuit.bind_def,
-    subcircuit.eq_1, ElaboratedCircuit.output, Fin.coe_ofNat_eq_mod, Nat.reduceMod,
-    FormalCircuit.toSubcircuit.eq_1, ElaboratedCircuit.main, Circuit.operations, G.Assumptions,
-    G.Spec, ElaboratedCircuit.localLength, List.cons_append, List.nil_append, Fin.val_two,
-    Operations.localLength.eq_5, Operations.localLength.eq_1, Nat.add_zero, Fin.val_one,
-    Fin.val_zero, Circuit.output, Circuit.ConstraintsHold.Completeness.eq_5,
-    Circuit.ConstraintsHold.Completeness.eq_1] at ⊢ henv h_input
+  simp only [main, circuit_norm, G.circuit, G.Assumptions, G.Spec] at ⊢ henv h_input
```
mitschabaude, line 108: awesome!

Clean/Gadgets/Equality.lean
```diff
@@ -38,6 +38,7 @@ instance elaborated (α : TypeMap) [ProvableType α] : ElaboratedCircuit F (Prov
   localLength_eq _ n := by simp only [main, circuit_norm, mul_zero]
   subcircuitsConsistent n := by simp only [main, circuit_norm]
 
+@[simps! (config := {isSimp := false, attrs := [`circuit_norm]})]
 def circuit (α : TypeMap) [ProvableType α] : FormalAssertion F (ProvablePair α α) where
```
mitschabaude, line 42: what's this doing and why?

mitschabaude (approved):
this is awesome! some comments (which might further improve build times)

Clean/Circuit/Subcircuit.lean
pirapira: When I remove these, the build fails at one place.

https://github.com/Verified-zkEVM/clean/blob/64fa247dda35945d6a6fe5a26e12c56ec5893681/Clean/Gadgets/Addition8/Addition8FullCarry.lean#L137

It looks like a legit usage because it's about `computableWitnesses`.

Clean/Gadgets/BLAKE3/Round.lean
pirapira: I want more and more proofs to look like this (even without `circuit` etc).

Clean/Circuit/Subcircuit.lean
mitschabaude: ok, that's because we don't "abstract away" the `computableWitnesses` property for subcircuits. the _intention_ is that we never need subcircuit operations, and the fact that we do need them in `computableWitnesses` shows that `computableWitnesses` proofs are not supported well yet :D

you can keep the lemmas, but remove them from `circuit_norm`; and add the necessary lemma to the `computableWitnesses` proof.

mitschabaude: (you could also entirely remove the lemmas and add `toSubcircuit` to the `computableWitnesses` proof)

pirapira: I'll remove these lemmas, and revisit `Addition8FullCarry` later.

Clean/Circomlib/Gates.lean
```diff
@@ -172,7 +172,7 @@ def circuit : FormalCircuit (F p) field field where
 
   soundness := by
     rintro _ _ _ _ h_env h_in h_hold
-    simp only [circuit_norm, main] at h_env h_hold ⊢
+    simp only [circuit_norm, main, Gadgets.Equality.circuit] at h_env h_hold ⊢
```
mitschabaude: btw why is `Gadgets.Equality.circuit` needed here? that worries me, the intention is that this circuit is already well unpacked by `circuit_norm`

Clean/Circuit/Subcircuit.lean
pirapira: removed these lemmas and repaired `Addition8FullCarry.

Clean/Gadgets/Addition8/Addition8FullCarry.lean
```diff
@@ -134,7 +134,7 @@ def lookupCircuit : LookupCircuit (F p) Inputs Outputs := {
   name := "Addition8FullCarry"
 
   computableWitnesses n input := by
-    simp_all only [circuit_norm, circuit, main,
+    simp_all only [circuit_norm, circuit, main, FormalAssertion.toSubcircuit,
       Operations.forAllFlat, Operations.toFlat, FlatOperation.forAll, Inputs.mk.injEq]
```
mitschabaude: this is still a very nice proof :D

Clean/Circomlib/Gates.lean
pirapira: I removed it and it's still working.

mitschabaude (approved):

Clean/Gadgets/Equality.lean
pirapira: The moment `circuit_norm` sees `Equality.circuit.anything`, it unfolds `Equality.circuit` and leaves `anything` field's definition.

pirapira: I think there's no opposition against having this for basic things like `Equality`. The thing is radical @pirapira is saying it should be done for all.

Clean/Gadgets/Addition8/Addition8FullCarry.lean
pirapira: The same thing is needed for the same reason in MultiAND.

https://github.com/Verified-zkEVM/clean/pull/220/commits/925042b8c1ff628534c957dc29f3354dcfd48a0d

The thing is not that complicated but it's there. Probably need a lemma for handling this.

Clean/Gadgets/Equality.lean
mitschabaude: alright, note that we didn't need that so far for `Equality.circuit` though by having dedicated rewrite rules. I'll have to see how this changed the look of goals

PR #221, opened 2025-07-17
Refactor ApplyRounds proof

pirapira:
Most lines in the `completeness` proof was spent for checking that the initial state vector is normalized. This PR factors these lines out to a new lemma `initial_state_and_messages_are_normalized`.

Also, the computation of the initial state vector was repeated. three times in the file, so this was factored out to `initializeStateVector` function.


pirapira:
Now I see the same thing in `soundness` too.

mitschabaude (approved):
nice deduplication

PR #226, opened 2025-07-19
Add decompose_provable_struct tactic

pirapira:
## Summary
- Added a new tactic `decompose_provable_struct` that automatically decomposes variables with `ProvableStruct` instances when they appear in field projections
- When the goal or hypotheses contain `x.field` for some variable `x` with a `ProvableStruct` instance, the tactic will decompose `x` into its components
- Only decomposes variables that actually appear in projections (not just any variable with the right type)

## Test plan
- Added comprehensive tests in `Clean/Utils/TestDecomposeProvableStruct.lean`
- Updated `Clean/Gadgets/BLAKE3/Compress.lean` to use the new tactic
- `lake build` succeeds


Clean/Utils/Tactics.lean
```diff
@@ -54,4 +56,192 @@ partial def SpecializeAuto (e_term : Lean.Term): Lean.Elab.Tactic.TacticM Unit :
...
+  1. Are in the context as variables with ProvableStruct instances
+  2. Appear in field projections in hypotheses (e.g., h : input.x = 5)
+  3. Appear in field projections in the goal
+  
+  Example:
+  ```lean
+  theorem example_theorem (input : Inputs (F p)) (h : input.x = 5) : 
+    input.y + input.z = someValue := by
+    decompose_provable_struct  -- This will destruct `input` (found via projections)
+    -- Now we have x, y, z in context with h : x = 5
```
pirapira: The new variables are not explicitly named. The user would need to call `rename_i` if they need names.

Clean/Gadgets/BLAKE3/ApplyRounds.lean
```diff
@@ -472,6 +456,18 @@ def Spec (input : Inputs (F p)) (out: BLAKE3State (F p)) :=
     flags.value ∧
   out.Normalized
 
+omit p_large_enough in
+@[circuit_norm]
+lemma eval_vector_eq_get {M : TypeMap} [NonEmptyProvableType M] {n : ℕ} (env : Environment (F p))
```
mitschabaude: Is this used here? I see that it is better-suited for circuit_norm than the similar getElem_eval_vector, because it doesn't introduce a potential cycle with breaking down the RHS eval. I just wonder if simp is smart enough to pass in the h_input hypothesis?

In any case, it belongs to Probable.lean, not here!

Clean/Gadgets/BLAKE3/ApplyRounds.lean
```diff
@@ -481,37 +477,19 @@ lemma initial_state_and_messages_are_normalized
...
-    h_eval_counter_low, h_eval_block_len, h_eval_flags⟩ := h_input
+  simp [circuit_norm, Inputs.mk.injEq] at h_input
 
   -- Helper to prove normalization of chaining value elements
   have h_chaining_value_normalized (i : ℕ) (h_i : i < 8) : (eval env chaining_value_var[i]).Normalized := by
-    have h : (eval env chaining_value_var : ProvableVector _ _ _) = chaining_value := h_eval_chaining_block_value
-    have h_i : (eval env chaining_value_var : ProvableVector _ _ _)[i] = chaining_value[i] := by
-      rw [h]
-    simp only [eval_vector, Vector.getElem_map] at h_i
+    simp_all [circuit_norm]
```
mitschabaude: Nit: non-terminal unrestricted simp

Clean/Utils/Tactics.lean
```diff
@@ -54,4 +56,192 @@ partial def SpecializeAuto (e_term : Lean.Term): Lean.Elab.Tactic.TacticM Unit :
...
   specializess it with them.
 -/
-elab "specialize_auto" e:term : tactic => withMainContext do specializeAuto e
+elab "specialize_auto" e:term : tactic => withMainContext do SpecializeAuto e
+
+/--
+  Find field projections in an expression and return the base variable if it's a structure
+-/
+partial def findStructVars (e : Lean.Expr) : Lean.MetaM (List Lean.FVarId) := do
+  let e' ← whnf e
```
mitschabaude: Wouldn't it make more sense to put this at the start of 'go'? It seems like in recursive calls, you're no longer operating on the whnf (weak head normal form)

Clean/Utils/Tactics.lean
```diff
@@ -54,4 +56,192 @@ partial def SpecializeAuto (e_term : Lean.Term): Lean.Elab.Tactic.TacticM Unit :
   `h1 -> h2 -> ... -> hn -> t'`, it searches in the context for `h1`, `h2`, ..., `hn` and
   specializess it with them.
 -/
-elab "specialize_auto" e:term : tactic => withMainContext do specializeAuto e
+elab "specialize_auto" e:term : tactic => withMainContext do SpecializeAuto e
+
+/--
+  Find field projections in an expression and return the base variable if it's a structure
+-/
+partial def findStructVars (e : Lean.Expr) : Lean.MetaM (List Lean.FVarId) := do
```
mitschabaude, line 64: Why is this a partial def?

Clean/Utils/TestDecomposeProvableStruct.lean
```diff
@@ -0,0 +1,149 @@
...
+theorem test_decompose_nested {F : Type} [Field F] (input : NestedInputs F) :
+    input.first.x + input.second.y = input.second.y + input.first.x := by
+  decompose_provable_struct
+  -- This should decompose input into first and second
+  rename_i first second
+  -- input should no longer exist
+  fail_if_success (exact input)
+  -- first and second should exist
+  have : TestInputs F := first
+  have : TestInputs F := second
```
mitschabaude: This test makes it look like the tactic fails to decompose nested structs! .x and .y appear as projections but we don't seem to get separate variables for them

Clean/Utils/TestDecomposeProvableStruct.lean
```diff
@@ -0,0 +1,149 @@
...
+  -- a and b should no longer exist
+  fail_if_success (exact a)
+  fail_if_success (exact b)
+  -- All components should exist
+  have : F := ax
+  have : F := bx
+  ring
+
+-- Test with multiple variables using automatic version
+theorem test_decompose_multiple_auto {F : Type} [Field F] (a : TestInputs F) (b : TestInputs F) :
```
mitschabaude: How does this test differ from the above?

Clean/Utils/TestDecomposeProvableStruct.lean
mitschabaude, line 1: Can you move this file to a folder Utils/Test?

mitschabaude (commented):
Very nice addition. Nested structs might not work yet.

Idea for later: I have the feeling that decomposeProvableStruct could also be implemented as a simproc. That would
- avoid iterating through the AST yourself
- instead, use simp's iteration which is probably more robust (might fix the issue with nested structs)
- let us apply the transformation at specific hypotheses instead of everywhere (could also be a downside)

Clean/Gadgets/BLAKE3/ApplyRounds.lean
pirapira: >  I just wonder if simp is smart enough to pass in the h_input hypothesis?

Yes, it is.

> it belongs to Probable.lean, not here!

I'll move.

Clean/Gadgets/BLAKE3/ApplyRounds.lean
pirapira: Fixed.

Clean/Utils/Tactics.lean
pirapira: Observations
1. if I move this to the first line of `go` (and fix variable names), I see kernel panic in `ApplyRounds`
2. if I remove this `whnf`, `input` and `input_var` are not found.

I'll investigate.

pirapira: I'll remove the `whnf`.

The tactic was finding projections that are not visible in the goal state (because `Assumptions` etc were not unfolded yet).

This one tactic is meant to be an element in a larger tactic construction, so it should be doing just one thing of dealing with projections that are already visible.

pirapira: Removed the `whnf`. The caller needs to unfold things until they see relevant projections, but that's a separate problem.

Clean/Utils/TestDecomposeProvableStruct.lean
pirapira: I'm thinking about using this under some `repeat` loop.

I'll document this limitation.

Clean/Utils/TestDecomposeProvableStruct.lean
pirapira: Removed.

Clean/Utils/TestDecomposeProvableStruct.lean
pirapira: will do.

Clean/Utils/Tactics.lean
pirapira: to get away with not proving termination.

I think this is very common in meta-programming situation. Usually the number of steps taken will be the blocker, and something eventually finishes (with very big bounds) doesn't matter much.

Clean/Utils/Tactics.lean
pirapira: Updated the document.

Clean/Utils/TestDecomposeProvableStruct.lean
pirapira: documented.

Clean/Utils/TestDecomposeProvableStruct.lean
pirapira: done.

Clean/Gadgets/BLAKE3/ApplyRounds.lean
pirapira: > I just wonder if simp is smart enough to pass in the h_input hypothesis?

Not reliably.

I have a place
```
        apply eval_vector_eq_get
        simp only [h_input]
```
works, but 
```
        simp only [eval_vector_eq_get, h_input]
```

It seems like Lean's `simp` doesn't try to solve side-conditions like Isabelle's `simp` despite the same name.

pirapira: When I tried adding this in `circuit_norm`, some existing proofs showed challenges like above.

For now I use this lemma explicitly.

mitschabaude (approved):

PR #227, opened 2025-07-21
Add tactics for simplifying struct equalities/evaluations

pirapira:
## Summary
Implements a comprehensive set of tactics for simplifying struct equalities and evaluations for types with `ProvableStruct` instances. These tactics combine common idiomatic practices for handling equality proofs and eval expressions.

### New tactics

**`circuit_proof_start`** - Automatically performs initial proof setup:
   - Introduces variables from the context
   - Decomposes struct and pair types using the following sub-tactic

**`provable_simp`** (renamed from `provable_struct_simp`) - Main simplification tactic that combines:
   - `split_provable_struct_eq` - splits struct equalities, if components are mentioned
   - `decompose_provable_struct` - simplifies struct projections, if projections exist in the goal state
   - `simplify_provable_struct_eval` - simplifies eval expressions with structs, if components are mentioned
   - `simp only [] at *` - normalizes field access `{x := X, ...}.x` into `X`.

### StructEquality Splitting
- **`struct literal = struct literal`**: Splits into component-wise equalities
- **`struct literal = variable`**: Applies cases on the variable then splits
- **Handles nested structures**: Works recursively on conjunctions

### Eval Simplification
- **`eval env struct_var = struct_literal`**: Decomposes and simplifies

### Circuit Proof Automation
- `circuit_proof_start` automatically sets up proofs by introducing and decomposing all relevant structures
- standardizes the initial proof steps for circuit soundness and completeness

## Test plan
- lake build succeeds

Co-Authored-By: Claude <noreply@anthropic.com>


Clean/Utils/Tactics/SpecializeAuto.lean
```diff
@@ -0,0 +1,58 @@
...
+    replaceMainGoal (mvarIds' ++ [mvarId])
+  else
+    throwError "'specialize_auto' requires a term that appears in the local context"
+
+/--
+  Takes in input a term `t` which must be present in the context. If `t` is of the form
+  `h1 -> h2 -> ... -> hn -> t'`, it searches in the context for `h1`, `h2`, ..., `hn` and
+  specializess it with them.
+-/
+elab "specialize_auto" e:term : tactic => withMainContext do SpecializeAuto e
```
pirapira: Add a newline at the end of the file.

Clean/Utils/Tactics/SplitProvableStructEq.lean
```diff
@@ -0,0 +1,199 @@
...
+      (h : TestInputs.mk 1 2 3 = input ∧ x = 7) : 
+      input.x = 1 := by
+    split_provable_struct_eq
+    -- input is automatically destructured via cases
+    -- The struct equality inside the conjunction is found and split
+    -- h.1 becomes: 1 = x ∧ 2 = y ∧ 3 = z
+    exact h.1.1.symm
+  ```
+-/
+elab "split_provable_struct_eq" : tactic => splitProvableStructEq
```
pirapira: Add a newline at the end of the file.

Clean/Utils/Tactics/DecomposeProvableStruct.lean
```diff
@@ -0,0 +1,202 @@
...
+    let fvarIds ← findProvableStructVars
+
+    if fvarIds.isEmpty then
+      logInfo "No ProvableStruct variables found in context or goal"
+    else
+      logInfo "Found ProvableStruct variables:"
+      for fvarId in fvarIds do
+        let localDecl ← fvarId.getDecl
+        let type ← inferType localDecl.toExpr
+        logInfo m!"  {localDecl.userName} : {type}"
```
pirapira: Add a newline at the end of the file.

Clean/Utils/Tactics.lean
```diff
@@ -1,254 +1,6 @@
...
-      for fvarId in fvarIds do
-        let localDecl ← fvarId.getDecl
-        let type ← inferType localDecl.toExpr
-        logInfo m!"  {localDecl.userName} : {type}"
+-- This file re-exports all tactics from the Tactics directory
+import Clean.Utils.Tactics.SpecializeAuto
+import Clean.Utils.Tactics.DecomposeProvableStruct
+import Clean.Utils.Tactics.SplitProvableStructEq
+import Clean.Utils.Tactics.SimplifyProvableStructEval
+import Clean.Utils.Tactics.ProvableStructSimp
```
pirapira: Add a newline at the end of the file.

Clean/Utils/Test/TestProvableStructSimp.lean
```diff
@@ -0,0 +1,185 @@
...
+    (input output : TestInputs F)
+    (h : input = output) :
+    input.x = output.x := by
+  -- First, let's create a hypothesis that looks like eval = eval
+  have h_eval : ProvableType.eval env (ProvableType.const input) = ProvableType.eval env (ProvableType.const output) := by
+    simp [ProvableType.eval_const, h]
+  -- Now try to apply simplify_provable_struct_eval - it should fail
+  -- because neither side has a struct literal
+  try simplify_provable_struct_eval
+  -- The tactic should have made no progress, so we prove it differently
```
pirapira: Assert failure of `progress (try simplify_provable_struct_eval)`.

Clean/Utils/Tactics/SimplifyProvableStructEval.lean
```diff
@@ -0,0 +1,137 @@
...
+            ProvableStruct.fromComponents,
+            ProvableStruct.eval.go,
+            ProvableType.eval_field
+          ] at $(mkIdent decl.userName):ident)
+          evalTactic tac
+          anyModified := true
+        catch _ => continue
+  
+  if !anyModified then
+    throwError "simplify_provable_struct_eval made no progress"
```
pirapira: Add a newline at the end of the file.

Clean/Utils/Tactics/ProvableStructSimp.lean
```diff
@@ -0,0 +1,39 @@
...
+  Simplify all provable struct expressions by repeatedly applying:
+  1. `split_provable_struct_eq` - splits struct equalities into field-wise equalities
+  2. `decompose_provable_struct` - destructures struct variables that appear in projections
+  3. `simplify_provable_struct_eval` - simplifies eval expressions in equalities with struct literals
+
+  The tactic continues until no transformation makes any more progress.
+
+  This is useful for normalizing proofs involving structures with ProvableStruct instances,
+  as it automatically:
+  - Splits equalities like `s1 = s2` into `s1.f1 = s2.f1 ∧ s1.f2 = s2.f2 ∧ ...`
```
pirapira: Add "if some components of `s1` or `s2` are explicitly mentioned".

Clean/Utils/Tactics/ProvableStructSimp.lean
```diff
@@ -0,0 +1,39 @@
...
+  - Destructures variables like `input` that appear in projections like `input.x`
+  - Simplifies `eval env var = struct_literal` to expose the structure
```
pirapira: "expose components" is clearer.

Clean/Utils/Tactics/ProvableStructSimp.lean
```diff
@@ -0,0 +1,39 @@
...
+
+/--
+  Simplify all provable struct expressions by repeatedly applying:
+  1. `split_provable_struct_eq` - splits struct equalities into field-wise equalities
+  2. `decompose_provable_struct` - destructures struct variables that appear in projections
+  3. `simplify_provable_struct_eval` - simplifies eval expressions in equalities with struct literals
+
+  The tactic continues until no transformation makes any more progress.
+
+  This is useful for normalizing proofs involving structures with ProvableStruct instances,
```
pirapira: "normalizing goal states" instead of "normalizing proofs"

Clean/Utils/Tactics/DecomposeProvableStruct.lean
```diff
@@ -0,0 +1,202 @@
...
+      let type' ← withTransparency .reducible (whnf type)
+      let typeCtor := type'.getAppFn
+      match typeCtor with
+      | .const name _ =>
+        try
+          let instType ← mkAppM ``ProvableStruct #[.const name []]
+          if let .some _ ← trySynthInstance instType then
+            result := result ++ [fvarId]
+        catch _ => continue
+      | _ => continue
```
pirapira: These lines are repeated for the goal and for each hypothesis. It's possible to refactor these.

Clean/Utils/Tactics/SimplifyProvableStructEval.lean
```diff
@@ -0,0 +1,137 @@
...
+        try
+          let tac ← `(tactic| simp only [
+            ProvableStruct.eval_eq_eval,
+            ProvableStruct.eval,
+            ProvableStruct.fromComponents,
+            ProvableStruct.eval.go,
+            ProvableType.eval_field
+          ] at $(mkIdent decl.userName):ident)
+          evalTactic tac
+          anyModified := true
```
pirapira: These lines are repeated above and here. A refactoring is possible.

pirapira:
Draft because some debugging is happening.

Clean/Utils/Tactics/SimplifyPairEval.lean
```diff
@@ -0,0 +1,155 @@
+import Lean
+import Clean.Circuit.Provable
+import Clean.Utils.Tactics.DecomposeProvablePair
+import Clean.Utils.Tactics.ProvableTacticUtils
+
+open Lean Meta Elab Tactic
+
+namespace ProvenZK
```
mitschabaude: why this namespace?

Clean/Utils/Tactics/SimplifyPairEval.lean
```diff
@@ -0,0 +1,155 @@
...
+      eval_pair_left_vector_expr,
+      eval_pair_right_vector_expr,
+      eval_pair_both_vector_expr,
+      Prod.fst,
+      Prod.snd,
+      Prod.mk.injEq
+    ] at $(mkIdent declName):ident)
+    evalTactic tac
+
+  -- Process each hypothesis
```
mitschabaude: is there any good reason this tactic is not just:
```lean
simp only [...lemmas] at *
```
or asked differently: is there a good reason not to simplify expressions like `(eval env x).1` to `eval env x.1` when they _don't_ appear in an equality?

personally I'd rather just add `Prod.mk.injEq` to `circuit_norm` and remove this tactic

Clean/Utils/Tactics/SimplifyPairEval.lean
```diff
@@ -0,0 +1,155 @@
...
+      ProvableType.eval_fieldTriple,
+      eval_pair,
+      eval_pair_left_expr,
+      eval_pair_right_expr,
+      eval_pair_both_expr,
+      eval_pair_left_vector_expr,
+      eval_pair_right_vector_expr,
+      eval_pair_both_vector_expr,
+      Prod.fst,
+      Prod.snd,
```
mitschabaude: why are `Prod.fst` and `Prod.snd` in this list?

Clean/Utils/Tactics/SplitPairEq.lean
mitschabaude, line 1: what is the benefit of this tactic, over doing
```
decompose_provable_pair
simp only [Prod.mk.injEq] at *
```

Clean/Utils/Tactics/SplitProvableStructEq.lean
```diff
@@ -0,0 +1,156 @@
...
+        catch e =>
+          trace[Meta.Tactic] "Failed to apply lemma {lemmaIdent}: {e.toMessageData}"
+          continue
+
+/--
+  Automatically split struct equalities (where struct has ProvableStruct instance) into field-wise equalities.
+  
+  This tactic:
+  1. First applies `cases` on struct variables that appear in equalities with struct literals
+  2. Then applies `mk.injEq` lemmas for all ProvableStruct types found in the context
```
mitschabaude: part (2) makes a lot of sense to me and is very useful

but why doesn't it just rely on `decompose_provable_struct` for part (1)?
in `provable_simp`, this tactic is used in conjunction with `decompose_provable_struct` anyway, so part (1) seems redundant.

Clean/Utils/Tactics/ProvableSimp.lean
```diff
@@ -0,0 +1,58 @@
...
+  ```
+-/
+macro "provable_simp" : tactic =>
+  `(tactic|
+    repeat (
+      fail_if_no_progress (
+      try split_provable_struct_eq;
+      try split_pair_eq;
+      try decompose_provable_struct;
+      try decompose_provable_pair;
```
mitschabaude: more logical to me would be to call `decompose_provable_struct` before `split_provable_struct_eq`. that way `split_provable_struct_eq` has more decomposed variables to operate on. (it will do that in the next loop iteration anyway, but more efficient to do it within the same loop)

similarly, `split_pair_eq` should come after `decompose_provable_pair`. however, I think `split_pair_eq` in that context can be replaced with `simp only `Prod.mk.injEq`, so probably it shouldn't exist.

Clean/Utils/Tactics/SimplifyProvableStructEval.lean
```diff
@@ -0,0 +1,127 @@
...
+  
+  -- Helper to apply the simp lemmas to a specific hypothesis
+  let applySimpToHyp (declName : Lean.Name) : Lean.Elab.Tactic.TacticM Unit := do
+    let tac ← `(tactic| simp only [
+      ProvableStruct.eval_eq_eval,
+      ProvableStruct.eval,
+      ProvableStruct.fromComponents,
+      ProvableStruct.eval.go,
+      ProvableType.eval_field
+    ] at $(mkIdent declName):ident)
```
mitschabaude: similar to `simplify_pair_eval`, this tactic reimplements
```lean
simp only [..lemmas] at *
```
with (*) replaced by a smaller set of hypothesis. why don't we use just use `at *`?

Clean/Utils/Tactics/ProvableSimp.lean
```diff
@@ -0,0 +1,58 @@
...
+      try simplify_provable_struct_eval;
+      try simplify_pair_eval;
+      try simp only [] at *
```
mitschabaude: you don't need the empty brackets:
```suggestion
      try simp only at *
```

Clean/Utils/Tactics/ProvableSimp.lean
```diff
@@ -0,0 +1,58 @@
...
+macro "provable_simp" : tactic =>
+  `(tactic|
+    repeat (
+      fail_if_no_progress (
+      try split_provable_struct_eq;
+      try split_pair_eq;
+      try decompose_provable_struct;
+      try decompose_provable_pair;
+      try simplify_provable_struct_eval;
+      try simplify_pair_eval;
```
mitschabaude: It seems to me these two tactics can be safely replaced by a `simp only [...lemmas] at *` with an appropriate set of lemmas

Clean/Utils/Test.lean
mitschabaude, line 1: this could be moved to the `Clean/` level so to import tests outside `Utils` as well eventually

Clean/Utils/Tactics/SplitProvableStructEq.lean
```diff
@@ -0,0 +1,156 @@
+import Lean.Elab.Tactic
+import Lean.Elab.Exception
+import Clean.Circuit.Provable
+import Clean.Utils.Tactics.ProvableTacticUtils
+
+open Lean.Elab.Tactic
+open Lean.Meta
+open Lean
+open ProvenZK
```
mitschabaude: can we get rid of the `ProvenZk` namespace?

Clean/Utils/Tactics/CircuitProofStart.lean
```diff
@@ -0,0 +1,159 @@
+import Lean
+import Clean.Circuit.Basic
+import Clean.Utils.Tactics.ProvableSimp
+
+namespace ProvenZK
```
mitschabaude: remove namespace

Clean/Utils/Tactics/CircuitProofStart.lean
```diff
@@ -0,0 +1,159 @@
...
+
+open Lean Elab Tactic Meta
+open Circuit
+
+/--
+  Apply intro until we've introduced all the standard parameters for
+  soundness or completeness proofs, then apply additional simplifications
+-/
+partial def circuitProofStartCore : TacticM Unit := do
```
mitschabaude: I like this tactic a lot!

personally I would prefer a shorter name for `offset`. something that is not likely to collide with other index variable names like `i` and `j`. for example, `i0` or `i₀` or `n₀`.

shorter variable names cause easier-to-read goal states, which has a real impact on usability IMO.

also, personally, I would prefer the descriptive name `h_assumptions` over `h_asm`. sounds like contradicting myself :D but `h_assumptions` won't appear multiple times as part of expressions in the goal state, as opposed to `offset`, so the longer name doesn't matter

Clean/Utils/Tactics/CircuitProofStart.lean
```diff
@@ -0,0 +1,159 @@
...
+        try
+          -- Check if this constant exists and is a definition
+          let info ← getConstInfo candidate
+          match info with
+          | .defnInfo _ =>
+            -- It's a definition, try to unfold it
+            let ident := mkIdent candidate
+            try (evalTactic (← `(tactic| simp only [$ident:ident] at *))) catch _ => pure ()
+          | _ => pure ()
+        catch _ => pure ()
```
mitschabaude: is this any different than
```lean
unfold ...names at *
```
?

Clean/Utils/Tactics/CircuitProofStart.lean
```diff
@@ -0,0 +1,159 @@
...
+  -- Then apply additional unfolding and simplification
+  -- Unfold the circuit definition and common definitions
+  try (evalTactic (← `(tactic| dsimp only [ElaboratedCircuit.main, Circuit.bind_def, Circuit.output, main] at *))) catch _ => pure ()
+  -- Try to unfold Assumptions and Spec as local definitions
+  tryUnfoldLocalDefs [`Assumptions, `Spec]
+  -- Also try with delta reduction which is more aggressive
+  try (evalTactic (← `(tactic| delta Assumptions Spec))) catch _ => pure ()
+  -- Unfold the elaborated circuit definition
+  try (evalTactic (← `(tactic| unfold elaborated at *))) catch _ => pure ()
+  try (evalTactic (← `(tactic| provable_simp))) catch _ => pure ()
```
mitschabaude: I'd love if we treated core proving infra like this with a lot of care, in particular each line's impact here and their orders should be tested and justified.

```lean
try (evalTactic (← `(tactic| simp only [circuit_norm] at *))) catch _ => pure ()
```
why do this at the beginning? it would be much more powerful after unfolding some definitions and decomposing structs, as you (partially) do below.
is there any particular impact that calling this tactic at the beginning has, that you wanted?

```
try (evalTactic (← `(tactic| dsimp only [ElaboratedCircuit.main, Circuit.bind_def, Circuit.output, main] at *))) catch _ => pure ()
```
the snippet seems to always fail, since the name `main` is not available. probably it has to be quoted somehow, like you do with Spec and Assumptions below?

@pirapira please test stuff like this before having a PR reviewed!

also, what is the motivation for the other lemmas here? `ElaboratedCircuit.main, Circuit.bind_def, Circuit.output` are all part of `circuit_norm` and so would already be unfolded by the line above.

Maybe remove this line, and properly unfold `main` together with `Assumptions` and `Spec` below

```
-- Also try with delta reduction which is more 
try (evalTactic (← `(tactic| delta Assumptions Spec))) catch _ => 
```
this doesn't do anything, because `Assumptions` and `Spec` are not recognized. If I remove the `tryUnfoldLocalDefs` line above, they don't get unfolded.

again, a `try .. catch` line that always fails shouldn't be in a PR at review time

```
try (evalTactic (← `(tactic| unfold elaborated at *))) catch _ => pure ()
```
what's the motivation for this line? I didn't check, but based on the above I would assume it always fails

```
try (evalTactic (← `(tactic| provable_simp))) catch _ => pure ()
```
`provable_simp` is a very useful tactic. at the same time, I wonder why you don't call `circuit_norm` at the end after unfolding `Spec` and `Assumptions` (and unsuccessfully trying to unfold `main` and `elaborated`).
Do you think `circuit_norm` here does too much / is too opinionated?

Clean/Utils/Tactics/DecomposeProvableStruct.lean
mitschabaude, line 1: in `decompose_provable_pair`, you introduce names for the newly introduced variables, like `input_fst` and `input_snd`. we could do something very similar for structs, and I think that could be a big improvement to `circuit_proof_start` as a whole

Clean/Utils/Tactics/ProvableSimp.lean
```diff
@@ -0,0 +1,58 @@
...
+      try simp only [] at *
+      )
+    )
+  )
+
+-- Backward compatibility alias
+macro "provable_struct_simp" : tactic => `(tactic| provable_simp)
```
mitschabaude: this seems like a temporary hack that should be removed by now

Clean/Utils/Tactics/SimplifyPairEval.lean
pirapira: Just doing
```
simp only [...lemmas] at *
```
would expand all of the pairs, even when no `.1` or `.2` appear in the goal state.

> asked differently: is there a good reason not to simplify expressions like (eval env x).1 to eval env x.1 when they don't appear in an equality?

No good reasons against that. I just never saw an example that would be useful. All of the development here is motivated by the existing proofs.

mitschabaude:
Another note: This contains too much AI slop for my taste. When coding with AI, I think an initial thorough review should be done by the PR author @pirapira. Not doing so effectively offloads a lot of work to the PR reviewer.

Clean/Utils/Tactics/SplitProvableStructEq.lean
pirapira: `decompose_provable_struct` does not decompose `v` even when there's a hypothesis `h_eq : {x := .., y := .., z := ..} = v`.

I [added](https://github.com/Verified-zkEVM/clean/pull/227/commits/f043cad76dc3c90c1ee6647dff7eea908034d306#diff-53935aca88fab306d888540a2a5b6865dc8668f75246a8b47dcd737bb08484eeR192) a test witnessing that.

Clean/Utils/Tactics/SimplifyPairEval.lean
mitschabaude: > would expand all of the pairs, even when no .1 or .2 appear in the goal state.

this tactic already expands all eval pair equalities, regardless of whether .1 or .2 appear in the goal state

mitschabaude: > No good reasons against that. I just never saw an example that would be useful. All of the development here is motivated by the existing proofs.

got it! In that case let's just use a `simp only [...]`, and let's add `Prod.mk.injEq` to `circuit_norm` as I think it should always be useful

Clean/Utils/Tactics/CircuitProofStart.lean
pirapira: > but h_assumptions won't appear multiple times as part of expressions in the goal state, as opposed to offset, so the longer name doesn't matter

That makes a lot of sense. I'll come back here after some other more important things.

Clean/Utils/Tactics/CircuitProofStart.lean
pirapira: > why do this at the beginning?

I tried not doing `[circuit_norm]` at all. This caused some errors I didn't understand. I'll remove `[circuit_norm]` again and debug the issue in ernest.

> the snippet seems to always fail, since the name main is not available. probably it has to be quoted somehow, like you do with Spec and Assumptions below?

I will confirm this, and deal with the rest later.

Clean/Utils/Tactics/SplitProvableStructEq.lean
mitschabaude: thanks, understood! even then, after only doing part (2) you get `.x` in the hypotheses, so `decompose_provable_struct` will already introduce variables for them when they are used in conjunction. so when the two are used together, part (1) still looks redundant to me

Clean/Utils/Tactics/CircuitProofStart.lean
pirapira: I see that
```
try (evalTactic (← `(tactic| dsimp only [ElaboratedCircuit.main, Circuit.bind_def, Circuit.output, main] at *))) catch _ => pure ()
```
fails even when `#check main` succeeds. I think I was mistaken about `(tactic| ...main...` and `` `main``.

I am becoming curious because
```
  try (evalTactic (← `(tactic| simp only [circuit_norm] at *))) catch _ => pure ()
```
sometimes succeeds although it mentions `circuit_norm` in the same way as the above mentioned `main`.

Clean/Utils/Tactics/SimplifyPairEval.lean
pirapira: I thought odd and I left it.

Now I've removed this namespace all over.

Clean/Utils/Tactics/SplitProvableStructEq.lean
pirapira: The description (2) was wrong.

>  2. Then applies `mk.injEq` lemmas for all ProvableStruct types found in the context

This does too much. That is not desired. 2. would make `provable_simp` decompose `ProvableStruct` down to single field. I changed the description.

Clean/Utils/Tactics/SimplifyPairEval.lean
pirapira: > this tactic already expands all eval pair equalities,

It does. That's a bug. Given `pair_a` and `pair_b` Lean variables, A pair equality `pair_a = pair_b` or `eval env pair_a = pair_b` should not trigger decomposition of `pair_a` and `pair_b`.

I'll add a test against this, and repair the tactic.

mitschabaude:
> For this particular one, I want to turn U32 into a ProvableStruct.

Here's some background why it isn't a `ProvableStruct`, and making it one wouldn't be a good idea:
* We designed `ProvableStruct` such that it would always be unfolded by `circuit_norm`. Which gives you a very convenient way of defining "packages" of types that will be unpacked automatically (the package is necessary because certain things require a single type, like the Input and Output arguments of `FormalCircuit`)
* However, there are ProvableTypes which you typically do _not_ want to unfold, because they represent abstractions that you want to keep. U32 is a perfect example. If it's possible to reason about entire U32s at once, you want to do that. Unpacking them would unnecessarily complicate the proof statement and make it harder for Lean and humans to process.

In summary, as long as we keep the convenient default of always unpacking provable structs, types like U32 shouldn't be provable structs. On the other hand, provable structs that should be default not be unpacked can just be made provable types instead, so I don't see a particular reason to change the default of always unpacking provable structs!

pirapira:
> We designed ProvableStruct such that it would always be unfolded by circuit_norm

That's unexpected.

> there are ProvableTypes which you typically do not want to unfold

It depends on the context. Whether or not anything in the goal state is interested in its component. For any `ProvableStruct` or `U32`, there are internal proofs that look into the member fields, and users who don't want to see the member fields at all. It's not a distinction between `U32` and other records.

mitschabaude:
> It depends on the context. Whether or not anything in the goal state is interested in its component. For any ProvableStruct or U32, there are internal proofs that look into the member fields, and users who don't want to see the member fields at all. It's not a distinction between U32 and other records.

Yes, to be fair, when we designed this we didn't yet have/knew of the advanced tools that you are now using to detect whether something should be unfolded based on the circuit (i.e. looking at whether members are accessed).

So what we have is a heuristic that works well as a default, and pretty easy ways to step out of that default. (If you don't want your provable struct to always be unpacked: make it a provable type. If you want to unpack a provable type: just add the necessary lemmas to simp)

Clean/Utils/Tactics/SimplifyPairEval.lean
mitschabaude: I do think it's fine to always decompose them! using a pair is a very easy and lazy way of packaging up things, which signals "I need to package some things here but I don't care about keeping them packaged" (because, if their package would represent a meaningful abstraction, you'd use a custom `structure` instead of a pair)

Clean/Utils/Tactics/SimplifyPairEval.lean
```diff
@@ -0,0 +1,151 @@
...
+      Prod.mk.injEq
+    ] at $(mkIdent declName):ident)
```
pirapira: This tactic gets used in `h_eq : (pairA = pairB) /\ (c, d) = (e, f)` because `c, d, e, f` are mentioned. This causes `pairA` and `pairB` to get decomposed, and that's not desirable.

pirapira:
I'm taking pairs out of the scope.

Clean/Utils/Tactics/SimplifyPairEval.lean
pirapira: I'm going somewhat conservative here.

In general, it's more of roundabout getting automation that does too much, adjusting proofs, and getting the automation cleverer so it does less, and adjusting proofs back. I like it better when the automation does less than desired, gets more powerful and the proofs get simpler all the time.

pirapira:
> There seems to be the misunderstanding that ProvableStruct is fundamentally easier to deal with than ProvableType. It isn't.

`ProvableStruct` is usually a `structure`. And the shape of the projection is more predictable. General `ProvableType` will come at the end after dealing with `ProvableStruct`, `ProvablePair` and `ProvableVector`. Each of these has their own way of accessing components. There's a good argument that `ProvablePair` should always be decomposed component-wise. It's dubious decomposing `ProvableVector` always.

And then, the general `ProvableType` that doesn't fit in any of the above will get the most conservative treatment.

Clean/Utils/Tactics/SimplifyPairEval.lean
mitschabaude: fair enough! ✅

Clean/Utils/Tactics/SplitProvableStructEq.lean
mitschabaude: the way provable structs were originally intended this is exactly what it should do 😅 but fine to keep it at a smaller scope for now, and keep the tactic as is.

however, if this tactic doesn't end up reusing `decompose_provable_type`, then I'm requesting the same for it as for `decompose_provable_type`: automatically rename the decomposed variables

Clean/Utils/Tactics/SimplifyPairEval.lean
pirapira: This also goes together with keeping the implementation somehow simple enough, and I'm working on that.

Clean/Utils/Tactics/SimplifyPairEval.lean
pirapira: I vaguely remember `(x, y).1` or something, but unsure. I'm postponing `ProvablePairs`. They'll get a simpler treatment.

Clean/Utils/Tactics/ProvableSimp.lean
pirapira: These lines are gone, and the `provable_simp` is now called `provable_struct_simp`.

Clean/Utils/Tactics/SimplifyPairEval.lean
mitschabaude: I see, but that's handled by `simp only` already

Clean/Utils/Tactics/CircuitProofStart.lean
pirapira: I've renamed `offset` and `h_asm`.

Clean/Utils/Tactics/SimplifyPairEval.lean
pirapira: This thread is postponed till I get to `ProvablePair` in another PR.

Clean/Utils/Tactics/SimplifyPairEval.lean
pirapira: This thread is also postponed till I get to work with `ProvablePair`.

Clean/Utils/Tactics/SplitPairEq.lean
pirapira: Pairs are out of scope for now. Maybe ^ is plausible.

Clean/Utils/Tactics/ProvableSimp.lean
pirapira: >  call decompose_provable_struct before split_provable_struct_eq. that way split_provable_struct_eq has more decomposed variables to operate on.

The same reasoning can be used to support the current sequence. This is because there's a reverse dependency.

Starting from `h_eq : struct_a = {x := X, y := Y}`, first doing `split_provable_struct_eq` will get `h_eq : struct_a.x = X /\ struct_a.y = Y`. And then, `decompose_provable_struct` finds the projection.

Unless there's a projection, `decompose_provable_struct` doesn't decompose struct variables.

Clean/Utils/Tactics/SimplifyProvableStructEval.lean
pirapira: No, that decomposes all `ProvableStruct` variables, and that does too much.

Clean/Utils/Tactics/ProvableSimp.lean
pirapira: done.

Clean/Utils/Tactics/ProvableSimp.lean
pirapira: No, that decomposes all ProvableStruct variables, and that does too much.

Clean/Utils/Test.lean
pirapira: > import tests out side `Utils`

In which file do we want to import tests?

Clean/Utils/Tactics/CircuitProofStart.lean
pirapira: Writing
```
  try (evalTactic (← `(tactic| unfold `Assumptions `Spec at *))) catch _ => pure ()
```
give me an error message
```
unexpected token; expected identifier
```
at `` `Assumptions``.

So the biggest difference is in the parser's behavior in
```
tryUnfoldLocalDefs [...]
```
vs
```
  try (evalTactic (← `(tactic| unfold ... at *))) catch _ => pure ()
```

Either I read about `(tactic|...` or I experiment.

Clean/Utils/Tactics/ProvableSimp.lean
mitschabaude: yes I didn't realize this initially ✅

Clean/Utils/Tactics/CircuitProofStart.lean
pirapira: I got
```
  try (evalTactic (← `(tactic| unfold $(mkIdent `Assumptions):ident at *))) catch _ => pure ()
  try (evalTactic (← `(tactic| unfold $(mkIdent `Spec):ident at *))) catch _ => pure ()

```
This is already an improvement over having a 19 line funciton.

Clean/Utils/Tactics/CircuitProofStart.lean
pirapira: Now the tactic is doing
```
try (evalTactic (← `(tactic| simp only [circuit_norm] at *))) catch _ => pure ()
```
at the end. This got a fixable build only after I removed the always failing tactics.

Clean/Utils/Tactics/SimplifyProvableStructEval.lean
mitschabaude: I don't think you're picking specific variables to decompose here? Rather, you're picking specific hypotheses to apply these lemmas to. But why would we simplify eval of provable struct at some hypotheses and not others?

Specifically, this seems designed to apply to `h_input`, and it will decompose all structs there. Right?

Clean/Utils/Test.lean
mitschabaude: `Clean/Tests.lean` could import all tests, and `Clean.lean` could import `Clean.Tests`

Clean/Utils/Tactics/CircuitProofStart.lean
mitschabaude: nice!

Clean/Utils/Tactics/SimplifyProvableStructEval.lean
pirapira: > Specifically, this seems designed to apply to h_input, and it will decompose all structs there. Right?

That's what it does at the moment and that is a bug.

I think it's acceptable to have these simplifications to work on the specific `\alpha : ProvableStruct` whose components are mentioned.

pirapira: But I think I can specify an argument of `eval_eq_eval` so that this gets triggered only on a specific literal or variable.

pirapira: Then I found a test file is missing for `simplify_provable_struct_eval`.

pirapira: > But I think I can specify an argument of eval_eq_eval so that this gets triggered only on a specific literal or variable.

This approach is working now. There's a test case `test_conjunction_with_base_and_non_base` for that.

Clean/Utils/Test.lean
pirapira: I see that make sense.

pirapira: moved. Further reorganization is for later.

Clean/Utils/Tactics/CircuitProofStart.lean
pirapira: > @pirapira please test stuff like this before having a PR reviewed!

I generally do not try removing every line to see if that makes a difference.

Clean/Test.lean
```diff
@@ -0,0 +1,5 @@
+-- This file imports all test modules
+import Clean.Utils.Test.TestDecomposeProvableStruct
+import Clean.Utils.Test.TestSplitProvableStructEq
+import Clean.Utils.Test.TestCircuitProofStart
+import Clean.Utils.Test.TestSimplifyProvableStructEval
```
pirapira: Add a newline at the end of the file.

Clean/Utils/Tactics/CircuitProofStart.lean
```diff
@@ -0,0 +1,93 @@
+import Lean
+import Clean.Circuit.Basic
+import Clean.Utils.Tactics.ProvableSimp
+
+open Lean Elab Tactic Meta
+open Circuit
+
+/--
+  Apply intro until we've introduced all the standard parameters for
+  soundness or completeness proofs, then apply additional simplifications
```
pirapira: Better phrasing is:

Introduce all standard parameters and hypotheses for Soundness or Completeness.

Clean/Utils/Tactics/CircuitProofStart.lean
```diff
@@ -0,0 +1,93 @@
...
+partial def circuitProofStartCore : TacticM Unit := do
+  withMainContext do
+    let goal ← getMainGoal
+    let goalType ← goal.getType
+
+    -- First check if this is a Soundness or Completeness type that needs unfolding
+    -- We need to check the head constant of the expression
+    let headConst? := goalType.getAppFn.constName?
+
+    -- Check if this is a Soundness or Completeness proof
```
pirapira:
```suggestion
```

Clean/Utils/Tactics/CircuitProofStart.lean
```diff
@@ -0,0 +1,93 @@
...
+    let isSoundness := headConst? == some ``Soundness
+    let isCompleteness := headConst? == some ``Completeness
+
+    if isSoundness then
+      -- This is a Soundness proof, unfold it and introduce all parameters with names
```
pirapira:
```suggestion
```

Clean/Utils/Tactics/CircuitProofStart.lean
```diff
@@ -0,0 +1,93 @@
...
+      evalTactic (← `(tactic| unfold Soundness))
+      -- Introduce parameters with explicit names using introN
```
pirapira:
```suggestion
```

Clean/Utils/Tactics/CircuitProofStart.lean
```diff
@@ -0,0 +1,93 @@
...
+      let names := [`i₀, `env, `input_var, `input, `h_input, `h_assumptions, `h_holds]
+      for name in names do
+        evalTactic (← `(tactic| intro $(mkIdent name):ident))
+      return
+    else if isCompleteness then
+      -- This is a Completeness proof, unfold it and introduce all parameters with names
```
pirapira:
```suggestion
```

Clean/Utils/Tactics/CircuitProofStart.lean
```diff
@@ -0,0 +1,93 @@
...
+      evalTactic (← `(tactic| unfold Completeness))
+      -- Introduce parameters one by one
```
pirapira:
```suggestion
```

Clean/Utils/Tactics/CircuitProofStart.lean
```diff
@@ -0,0 +1,93 @@
...
+    else
+      -- Goal is neither Soundness nor Completeness
+      throwError "circuitProofStartCore can only be used on direct Soundness or Completeness goals"
+
+/--
+  Standard tactic for starting soundness and completeness proofs.
+
+  This tactic:
+  1. Automatically introduces all parameters for `Soundness` or `Completeness` goals
+  2. Applies provable_simp to decompose structs/pairs and simplify eval
```
pirapira:
```suggestion
  2. Applies provable_simp to decompose structs and decompose eval that mention struct components
```

Clean/Utils/Tactics/CircuitProofStart.lean
```diff
@@ -0,0 +1,93 @@
...
+  3. Unfolds circuit definitions using circuit_norm
+  4. Unfolds local Assumptions and Spec definitions
```
pirapira:
```suggestion
  3. Unfolds local Assumptions and Spec definitions
  4. Normalized the goal state using circuit_norm
```

Clean/Utils/Tactics/CircuitProofStart.lean
```diff
@@ -0,0 +1,93 @@
...
+
+  For completeness proofs, it introduces:
+  - i₀ (offset in the environment)
+  - env (environment)
+  - input_var (variable)
+  - henv (UsesLocalWitnessesCompleteness ...)
+  - input (value)
+  - h_input (eval env input_var = input)
+  - h_assumptions (Assumptions input)
+
```
pirapira:
```suggestion
```

Clean/Utils/Tactics/DecomposeProvableStruct.lean
```diff
@@ -0,0 +1,204 @@
...
+    for localDecl in ctx do
+      if localDecl.isImplementationDetail then
+        continue
+
+      -- Search in the type of each hypothesis
+      let varsInType ← findStructVars localDecl.type
+      -- Filter to only include those with ProvableStruct instances
+      for fvarId in varsInType do
+        if ← hasProvableStructInstance fvarId then
+          result := result ++ [fvarId]
```
pirapira: Taking cons is probably faster.

Clean/Utils/Tactics/DecomposeProvableStruct.lean
```diff
@@ -0,0 +1,204 @@
...
+
+    -- Also search for projections in the goal
+    let goal ← getMainGoal
+    let target ← goal.getType
+    let varsInGoal ← findStructVars target
+    -- Filter to only include those with ProvableStruct instances
+    for fvarId in varsInGoal do
+      if ← hasProvableStructInstance fvarId then
+        result := result ++ [fvarId]
```
pirapira: Taking cons is probably faster.

Clean/Utils/Tactics/DecomposeProvableStruct.lean
```diff
@@ -0,0 +1,204 @@
...
+
+    -- Remove duplicates and return
+    return result.eraseDups
```
pirapira: After we build the list using cons, we can reverse `result` here.

Clean/Utils/Tactics/DecomposeProvableStruct.lean
```diff
@@ -0,0 +1,204 @@
...
+
+/--
+  Decompose all variables with ProvableStruct instances in the context
```
pirapira:
```suggestion
  Decompose all ProvableStruct variables that appear in projections
```

Clean/Utils/Tactics/DecomposeProvableStruct.lean
```diff
@@ -0,0 +1,204 @@
...
+    let mut currentGoal := goal
+    let mut decomposed := false
+
+    for fvarId in fvarIds do
+      let localDecl ← fvarId.getDecl
+      let userName := localDecl.userName
+
+      try
+        -- Use cases tactic on the variable
+        let casesResult ← currentGoal.cases fvarId
```
pirapira: This is where the newly produced variables need to be named nicely.

Clean/Utils/Tactics/DecomposeProvableStruct.lean
```diff
@@ -0,0 +1,204 @@
...
+        -- Skip variables that can't be decomposed
+        -- This can happen with type synonyms or other non-destructurable types
+        trace[Meta.Tactic] "Failed to decompose struct variable {userName}: {e.toMessageData}"
+        continue
+
+    if not decomposed then
+      throwError "Failed to decompose any variables with ProvableStruct instances"
+
+    replaceMainGoal [currentGoal]
+
```
pirapira, line 165:
```suggestion
```

Clean/Utils/Tactics/DecomposeProvableStruct.lean
```diff
@@ -0,0 +1,204 @@
...
+
+/--
+  Automatically decompose ALL variables with ProvableStruct instances that either:
+  1. Are in the context as variables with ProvableStruct instances
+  2. Appear in field projections in hypotheses (e.g., h : input.x = 5)
+  3. Appear in field projections in the goal
```
pirapira:
```suggestion
  1. appear in field projections in hypotheses (e.g., h : input.x = 5)
  2. appear in field projections in the goal
```

Clean/Utils/Tactics/DecomposeProvableStruct.lean
```diff
@@ -0,0 +1,204 @@
...
+
+  Note: 
+  - The new variables introduced by decomposition are not explicitly named. 
+    Use `rename_i` after this tactic to give them names.
+  - This tactic only performs one level of decomposition. For nested structures,
+    it will decompose the outer structure but not the inner ones. Use under
+    `repeat` to decompose nested structures fully.
```
pirapira:
```suggestion
  - This tactic only performs one level of decomposition. For nested structures,
    it will decompose the outer structure but not the inner ones. Use under
    `repeat` to decompose nested structures fully. `provable_struct_simp` is one
    implementation of such a loop.
```

Clean/Utils/Tactics/DecomposeProvableStruct.lean
```diff
@@ -0,0 +1,204 @@
...
+  theorem example_theorem (input : Inputs (F p)) (h : input.x = 5) :
+    input.y + input.z = someValue := by
+    decompose_provable_struct  -- This will destruct `input` (found via projections)
+    -- Now we have auto-named variables for x, y and z components in context
+    -- and h has been updated to reference the new x component
+    sorry
+  ```
+-/
+elab "decompose_provable_struct" : tactic => decomposeProvableStruct
+
```
pirapira, line 191:
```suggestion
```

Clean/Utils/Tactics/DecomposeProvableStruct.lean
```diff
@@ -0,0 +1,204 @@
...
+
+/--
+  Print all ProvableStruct variables found in the context and goal
```
pirapira:
```suggestion
  Print all ProvableStruct variables found in projections in the context and goal
```

Clean/Utils/Tactics/ProvableSimp.lean
```diff
@@ -0,0 +1,38 @@
...
+/--
+  Simplify all provable struct expressions by repeatedly applying:
+  1. `split_provable_struct_eq` - splits struct equalities into field-wise equalities
+  2. `decompose_provable_struct` - destructures struct variables that appear in projections
+  3. `simplify_provable_struct_eval` - simplifies eval expressions in equalities with struct literals
+
+  The tactic continues until no transformation makes any more progress.
+
+  This is useful for normalizing goal states involving structures with ProvableStruct instances
+  or pairs (ProvablePair, fieldPair, Prod), as it automatically:
```
pirapira:
```suggestion
 as it automatically:
```

Clean/Utils/Tactics/SplitProvableStructEq.lean
```diff
@@ -0,0 +1,155 @@
...
+                if ← hasProvableStructInstance typeExpr' then
+                  let lemmaIdent := mkIdent mkInjEqName
+                  if !lemmasToApply.any (fun l => l.getId == mkInjEqName) then
+                    lemmasToApply := lemmaIdent :: lemmasToApply
+            | _ => pure ()
+
+      -- Apply all the lemmas we found
+      for lemmaIdent in lemmasToApply do
+        try
+          evalTactic (← `(tactic| simp only [$lemmaIdent:ident] at *))
```
pirapira, line 125: This might do too much when there are multiple conjuncts that mention the same struct type, and only some of them talk about the components. We need first a test case for this.

Clean/Utils/Test/TestSplitProvableStructEq.lean
```diff
@@ -0,0 +1,94 @@
...
+  -- h1 becomes: 1 = x₁ ∧ 2 = y₁ ∧ 3 = z₁
+  -- h2 becomes: x₂ = 4 ∧ y₂ = 5 ∧ z₂ = 6
+  constructor
+  · exact h1.1.symm
+  · exact h2.2.1
+
+-- Test with type synonyms (like Var)
+@[reducible]
+def TestVar (F : Type) := TestInputs F
+
```
pirapira: The codebase doesn't have this pattern, so no need to consider this in the tests.

Clean/Utils/Tactics/SplitProvableStructEq.lean
pirapira: I added a test case `test_with_base_and_non_base`.

The simplification leaves unrelated conjuncts intact. So it's already behaving as expected.

Clean/Utils/Tactics/ProvableSimp.lean
```diff
@@ -0,0 +1,38 @@
...
+ as it automatically:
+  - Splits equalities like `s1 = s2` into `s1.f1 = s2.f1 ∧ s1.f2 = s2.f2 ∧ ...` if some components of `s1` or `s2` are explicitly mentioned
+  - Destructures variables like `input` that appear in projections like `input.x` or `input.1` to expose components
+  - Simplifies `eval env var = struct_literal` to expose components
+  - Handles these transformations even inside conjunctions
+
+  Example with struct:
+  ```lean
+  theorem example (a b : MyStruct F) (h : a = b ∧ a.x = 5) : b.x = 5 := by
+    provable_simp
```
pirapira:
```suggestion
    provable_struct_simp
```

Clean/Utils/Tactics/DecomposeProvableStruct.lean
pirapira: Now new names are named like `input_low_counter` etc.

Clean/Utils/Test/TestCircuitProofStart.lean
```diff
@@ -0,0 +1,194 @@
...
+
+def Spec (input : unit (F p)) (output : unit (F p)) : Prop :=
+  TestSpec input output
+
+def testCircuit : ElaboratedCircuit (F p) unit unit :=
+  { main := fun _ => pure (), output := fun _ _ => (), localLength := 0, output_eq := by simp }
+
+example : Soundness (F p) testCircuit Assumptions Spec := by
+  circuit_proof_start
+  -- Assumptions and Spec should be unfolded to TestAssumptions and TestSpec
```
pirapira, line 127: Assert this somehow using tactics.

Clean/Utils/Test/TestCircuitProofStart.lean
```diff
@@ -0,0 +1,194 @@
...
+def Spec (input : unit (F p)) (output : unit (F p)) : Prop :=
+  TestSpec input output ∧
+  TestSpec input output
+
+def testCircuit : ElaboratedCircuit (F p) unit unit :=
+  { main := fun _ => pure (), output := fun _ _ => (), localLength := 0, output_eq := by simp }
+
+example : Soundness (F p) testCircuit Assumptions Spec := by
+  circuit_proof_start
+  -- Should unfold nested references
```
pirapira, line 151: Assert this somehow using tactics.

Clean/Utils/Test/TestCircuitProofStart.lean
```diff
@@ -0,0 +1,194 @@
...
+
+def elaborated : ElaboratedCircuit (F p) unit unit :=
+  testCircuit
+
+def TestAssumptions (_ : unit (F p)) : Prop := True
+def TestSpec (_ : unit (F p)) (_ : unit (F p)) : Prop := True
+
+example : Soundness (F p) elaborated TestAssumptions TestSpec := by
+  circuit_proof_start
+  -- elaborated should be unfolded to testCircuit
```
pirapira, line 170: Assert this using tactics.

Clean/Utils/Test/TestDecomposeProvableStruct.lean
```diff
@@ -130,10 +129,71 @@ theorem test_selective_decompose {F : Type} [Field F] (a : TestInputs F) (b : Te
...
+  -- input should NOT be decomposed because no projections are used
+  assumption
+
+-- Test with multiple struct variables, none with projections
+theorem test_multiple_no_projection {F : Type} [Field F]
+    (a : TestInputs F) (b : TestInputs F) (c : TestInputs F)
+    (ha : a = ⟨1, 2, 3⟩) (hb : b = ⟨4, 5, 6⟩) (hc : c = ⟨7, 8, 9⟩) :
+    a = ⟨1, 2, 3⟩ ∧ b = ⟨4, 5, 6⟩ ∧ c = ⟨7, 8, 9⟩ := by
+  try decompose_provable_struct
+  -- None should be decomposed
```
pirapira: Assert this using tactics. Maybe fail_if_success and progress is useful.

Clean/Utils/Test/TestDecomposeProvableStruct.lean
```diff
@@ -130,10 +129,71 @@ theorem test_selective_decompose {F : Type} [Field F] (a : TestInputs F) (b : Te
...
+section NoProjectionTests
+-- Test cases where the tactic errors because projections are never used
+-- decompose_provable_struct throws an error when no variables need decomposition
+
+-- Test struct variable with no projections used
+theorem test_no_projection_struct {F : Type} [Field F] (input : TestInputs F)
+    (h : input = ⟨5, 3, 1⟩) :
+    input = ⟨5, 3, 1⟩ := by
+  try decompose_provable_struct
+  -- input should NOT be decomposed because no projections are used
```
pirapira, line 151: Assert this using tactics. Maybe fail_if_success and progress is useful.

Clean/Utils/Test/TestDecomposeProvableStruct.lean
```diff
@@ -130,10 +129,71 @@ theorem test_selective_decompose {F : Type} [Field F] (a : TestInputs F) (b : Te
...
+  try decompose_provable_struct
+  -- None should be decomposed
+  exact ⟨ha, hb, hc⟩
+
+-- Test where only the whole struct is used in operations
+theorem test_whole_struct_equality {F : Type} [Field F]
+    (x : TestInputs F) (y : TestInputs F) (h : x = y) :
+    x = y := by
+  try decompose_provable_struct
+  -- No decomposition should happen
```
pirapira, line 168: Assert this using tactics. Maybe fail_if_success and progress is useful.

Clean/Utils/Test/TestDecomposeProvableStruct.lean
```diff
@@ -130,10 +129,71 @@ theorem test_selective_decompose {F : Type} [Field F] (a : TestInputs F) (b : Te
...
+
+-- Test with struct used in function application (not projection)
+def processStruct {F : Type} [Field F] (input : TestInputs F) : F :=
+  input.x + input.y + input.z
+
+theorem test_struct_in_function {F : Type} [Field F]
+    (input : TestInputs F) (h : processStruct input = 10) :
+    processStruct input = 10 := by
+  try decompose_provable_struct
+  -- input should NOT be decomposed because it's used as a whole in processStruct
```
pirapira, line 179: Assert this using tactics. Maybe fail_if_success and progress is useful.

Clean/Utils/Test/TestDecomposeProvableStruct.lean
```diff
@@ -130,10 +129,71 @@ theorem test_selective_decompose {F : Type} [Field F] (a : TestInputs F) (b : Te
...
+  assumption
+
+-- Test with struct comparison
+theorem test_struct_comparison {F : Type} [Field F]
+    (a : TestInputs F) (b : TestInputs F) :
+    a = b ↔ a = b := by
+  fail_if_success decompose_provable_struct
+  -- No decomposition should happen
```
pirapira, line 187: Assert this using tactics. Maybe fail_if_success and progress is useful.

Clean/Utils/Tactics/DecomposeProvableStruct.lean
pirapira: done

Clean/Utils/Tactics/SplitProvableStructEq.lean
pirapira: Now new variables are named.

Clean/Utils/Tactics/CircuitProofStart.lean
```diff
@@ -0,0 +1,73 @@
...
+  circuitProofStartCore
+  -- Try to unfold main, Assumptions and Spec as local definitions
+  try (evalTactic (← `(tactic| unfold $(mkIdent `Assumptions):ident at *))) catch _ => pure ()
+  try (evalTactic (← `(tactic| unfold $(mkIdent `Spec):ident at *))) catch _ => pure ()
+  try (evalTactic (← `(tactic| simp only [circuit_norm] at *))) catch _ => pure ()
+  -- circuit_norm exposes a `main` usually
+  try (evalTactic (← `(tactic| unfold $(mkIdent `main):ident at *))) catch _ => pure ()
+  try (evalTactic (← `(tactic| provable_struct_simp))) catch _ => pure ()
+  -- Additional simplification for common patterns in soundness proofs
+  try (evalTactic (← `(tactic| simp only [circuit_norm, h_input] at h_holds henv ⊢))) catch _ => pure ()
```
mitschabaude: this line doesn't work for reasons that should be obvious by now

@pirapira I can push a fix, but just wanted to double-check: did you actually want to `simp only [circuit_norm, h_input]` on the goals? because you [wrote before](https://github.com/Verified-zkEVM/clean/pull/227#issuecomment-3112872988) that you didn't want to do this for now

(in particular, `circuit_norm` will unfold all provable structs and pairs, which you've been trying to avoid)

pirapira: I do not have an understanding about the quoting.
```
`(tactic| ...)
```

I'll experiment.

pirapira: If it behaves like Scheme or Lisp, it should quote everything in `...`. It doesn't. I don't know what it does.

mitschabaude: ok I'll push my changes

pirapira: > in particular, circuit_norm will unfold all provable structs and pairs,

`circuit_norm` does not unfold all provable structs and pairs.

On this goal state
```
s1 s2 : Var SimpleStruct F
h : ({ x := Expression.eval env x_var, y := Expression.eval env y_var, z := Expression.eval env z_var } =
      { x := 1, y := 2, z := 3 } ∧
    x = 7) ∧
  eval env s1 = eval env s2
⊢ Expression.eval env x_var = 1
```
the tactic `simp only [circuit_norm]` does not make a progress (experimented in `TestSimplifyProvableStructEval.lean`).

pirapira: > ok I'll push my changes

You don't have to because I have a good guess.

pirapira: The experiment involved
```
  evalTactic (← `(tactic| fail_if_success simp only [circuit_norm, h_input] at h_holds henv ⊢))
```
with the whole build. The simp never succeeded.

pirapira: I think the `catch` all is quite bad. Maybe there's a variant of `simp` that doesn't fail just because it makes no progress.

pirapira:
```
  evalTactic (← `(tactic| try ...))
```
doesn't fail more often than
```
  try (evalTactic (← `(tactic| try ...))) catch _ => pure
```
The second form with `catch` is better because there's a possibility to analyze the exception in Lean.

pirapira: Now I see this working.

About `circuit_norm` and `ProvableStruct`. We need a way to scale the development. The main issue is to keep the goal state small enough.

Decomposing all `ProvableStruct` will cause tricky situations. I don't think it's feasible to add `ProvableStruct` while we want to decompose them and then remove the `ProvableStruct` instance when we no longer want to destruct them.

Fortunately at the moment `circuit_norm` doesn't decompose all `ProvableStruct`s.

`provable_struct_simp` in this PR is one scaling solution.

pirapira:
@mitschabaude Does this coincide with you change? https://github.com/Verified-zkEVM/clean/pull/227/commits/26c67c069c8782bfbfff7b76cd656e5bb11b068c Are there any remaining issues?

mitschabaude:
> @mitschabaude Does this coincide with you change? [26c67c0](https://github.com/Verified-zkEVM/clean/commit/26c67c069c8782bfbfff7b76cd656e5bb11b068c) Are there any remaining issues?

it didn't work yet for soundness proofs because of assuming `henv` was in scope, I fixed that as well and had to refactor the blake3b apply rounds proof a bit to make it work again

Clean/Utils/Tactics/CircuitProofStart.lean
mitschabaude: > the tactic simp only [circuit_norm] does not make a progress (experimented in TestSimplifyProvableStructEval.lean).

it does, probably you forgot to point `simp only [circuit_norm]` at the right hypothesis (`h`)

mitschabaude: > Fortunately at the moment circuit_norm doesn't decompose all ProvableStructs.

it does! by design, bc of the circuit_norm annotations here and on similar lemmas:

https://github.com/Verified-zkEVM/clean/blob/c7e643f2992fd38654ae9f8e1541a1c77f57b642/Clean/Circuit/Provable.lean#L223-L224

https://github.com/Verified-zkEVM/clean/blob/c7e643f2992fd38654ae9f8e1541a1c77f57b642/Clean/Circuit/Provable.lean#L285-L289

pirapira: That sounds problematic on two counts. Most lines in this PR are moot. This causes performance problems when nested ProvableStruct's are around.

I think this required artificial techniques like `FormalCircuit.concat`.

pirapira: So I'll drop `circuit_norm` from this PR. I think `circuit_norm` is for happy situation where decomposing all `ProvableStruct` works. BLAKE is already beyond that stage.

pirapira:
Another problem is that this PR does not have test cases agains undesired decompositions.

Clean/Utils/Tactics/CircuitProofStart.lean
mitschabaude: ok 😂 I think the tactic was nice but I get your viewpoint

mitschabaude (approved):
as soon as the proofs are fixed (should be trivial) this is good to go from my side. great work @pirapira!

note for myself later: I'd like to split out `Tests.lean` from `Clean.lean` compilation, because many of these tests have `sorry` in them by design which is noisy when compiling, and I generally like to target a `sorry`-free state of the main library.
this can be done by having CI run `lake build Clean.Test` separately, need more fine-grained CI for that

pirapira:
I'll follow up with two things.
* simplify the implementation of `start_circuit_proof`. `circuit_norm` decomposes lots of `ProvableStruct` anyway, so most detailed goal state analysis is not needed.
  * I'll keep the small tactics.
* investigate the two-stage situation where some proofs expand more `ProvableType`s using `explicit_provable_type` while others don't.

PR #228, opened 2025-07-23
Add BLAKE3 chunk processing specification

pirapira:
## Summary
- Implemented BLAKE3 chunk processing specification matching a Python implementation
- Added comprehensive test suite with test vectors from the Python implementation
- Fixed edge case where 64-byte inputs were handled differently than the Python version

### Core Components
- `ChunkState`: Structure tracking chaining value, chunk counter, blocks compressed, and buffer
- `processBlock`: Processes a single 64-byte block
- `splitIntoBlocks`: Splits input into complete blocks with proper handling of exact block-length inputs
- `updateChunk`: Main incremental processing function
- `finalizeChunk`: Finalizes chunk with CHUNK_END flag
- `processChunk`: High-level function for complete chunk processing

### Key Fix
Modified `splitIntoBlocks` to match Python behavior: when exactly `blockLen` bytes remain, they are kept as remainder for finalization with CHUNK_END flag, rather than being processed as a complete block.

### Tests
Added extensive test coverage including:
- Empty input
- Single byte inputs with various parameters
- Block boundary cases (63, 64, 65 bytes)
- Multiple block sizes (127, 128, 129, 256, 512 bytes)
- Full chunk sizes (1023, 1024 bytes)
- All test vectors validated against Python implementation

## Test plan
- [x] `lake build` succeeds
- [x] All test cases pass with `native_decide`
- [x] Test vectors match Python implementation


Clean/Specs/BLAKE3/ChunkProcessingTests.lean
```diff
@@ -0,0 +1,207 @@
...
+-- cv = test_process_chunk(bytes(range(64)), 0, 0)  # Full block
+-- cv = test_process_chunk(bytes(), 0, 0)  # Empty input
+-- ```
+
+-- Test: One byte [0x00], chunk_counter=0, flags=0
+example :
+    let input := [0x00]
+    let cv := processChunk testCV 0 input 0
+    cv = Vector.ofFn (fun i => [0x88a7f10d, 0x87d2711d, 0xfcc2afd0, 0x283dd2d7,
+                                0x1a402ef1, 0x26ca58b8, 0xf1c5117f, 0x15f30d71][i.val]!) := by
```
mitschabaude: can you use vector literals instead of `.ofFn`?

Clean/Specs/BLAKE3.lean
```diff
@@ -197,6 +197,130 @@ def compress (chaining_value: Vector Nat 8) (block_words: Vector Nat 16) (counte
...
+-- CHUNK PROCESSING
+--------
+
+/--
+State maintained during chunk processing.
+Tracks the current chaining value, number of blocks compressed,
+and any partial block data.
+-/
+structure ChunkState where
+  chaining_value : Vector Nat 8      -- Current chaining value (8 × 32-bit words)
```
mitschabaude: nit: use ℕ instead of Nat

Clean/Specs/BLAKE3.lean
```diff
@@ -197,6 +197,130 @@ def compress (chaining_value: Vector Nat 8) (block_words: Vector Nat 16) (counte
...
+Pads with zeros if less than 64 bytes.
+-/
+def bytesToWords (bytes : List Nat) : Vector Nat 16 :=
+  let paddedBytes := bytes ++ List.replicate (64 - bytes.length) 0
+  Vector.ofFn fun i =>
+    let byteIdx := i.val * 4
+    paddedBytes[byteIdx]! +
+    paddedBytes[byteIdx + 1]! * 256 +
+    paddedBytes[byteIdx + 2]! * 256^2 +
+    paddedBytes[byteIdx + 3]! * 256^3
```
mitschabaude: since this is a spec, it would be really nice to implement it without `!`. maybe using `Vector.toChunks`?

Clean/Specs/BLAKE3/ChunkProcessingTests.lean
pirapira: Done. `#v` wasn't known at the time.

Clean/Specs/BLAKE3.lean
pirapira: done.

Clean/Specs/BLAKE3.lean
pirapira: I couldn't find `Vector.toChunks`. Implementation with `List.toChunks` requires extra proof steps later.

I'll remove `!`s and prove bounds.

pirapira: I removed `!`s. I proved the list is long. Lean figured out the bounds.

mitschabaude: nice! btw `Vector.toChunks` is our own addition: https://github.com/Verified-zkEVM/clean/blob/e31284da96e2417d2059faf2c5c7047f6396cd01/Clean/Utils/Vector.lean#L303-L304
(but it's fine either way you proved it :))

pirapira: I see. I missed it.

PR #229, opened 2025-07-23
Implement BLAKE3 processBlocks as InductiveTable

### LINKED ISSUE #124: [BLAKE3] Example Inductive Table

pirapira:
## Summary
- Implements BLAKE3's processBlocks function as an InductiveTable with exactly 17 rows
- Updates InductiveTable framework to include initial state parameter in Spec

This PR closes #124.

### ProcessBlocksInductive Table
- One row per processBlock call (16 rows for blocks + 1 for final output)
- Uses BlockInput type with `block_exists` flag to handle variable-length inputs
- Implements conditional logic using arithmetic constraints

### Framework Updates
- Modified InductiveTable.Spec to include initial state parameter
- Added InitialStateAssumptions field for completeness
- Updated all existing InductiveTable instances (Fibonacci, Keccak)

## Test Plan
- [x] Lake build succeeds
- [x] ProcessBlocksInductive soundness proof
- [x] ProcessBlocksInductive completeness proof

### DIFF
--- a/Clean/Gadgets/IsZero.lean
+++ b/Clean/Gadgets/IsZero.lean
@@ -34,8 +34,8 @@ instance elaborated : ElaboratedCircuit F M field where
 
 def Assumptions (_ : M F) : Prop := True
 
-def Spec (input : M F) (output : F) : Prop :=
-  output = if (∀ i : Fin (size M), (toElements input)[i] = 0) then 1 else 0
+def Spec [DecidableEq (M F)] (input : M F) (output : F) : Prop :=
+  output = if input = 0 then 1 else 0
 
 /--
 lemma for soundness. Separate because the statement is optimized for induction.
@@ -53,10 +53,11 @@ lemma foldl_isZero_eq_one_iff {n : ℕ} {vars : Vector (Expression F) n} {vals :
       (Fin.foldl n
         (fun acc i => acc * (IsZeroField.circuit.output vars[i] (i₀ + i * IsZeroField.circuit.localLength vars[i]) : Var field F))
         1) =
-    if ∀ (i : Fin n), vals[i] = 0 then 1 else 0 := by
+    if ∀ (i : ℕ) (x : i < n), vals[i] = 0 then 1 else 0 := by
   simp only [IsZeroField.circuit, IsZeroField.Assumptions, IsZeroField.Spec] at h_isZero
   induction n generalizing i₀
   · simp only [id_eq, Fin.getElem_fin, Fin.foldl_zero, IsEmpty.forall_iff, ↓reduceIte, Expression.eval]
+    simp only [not_lt_zero', IsEmpty.forall_iff, implies_true, ↓reduceIte]
   · rename_i pre h_ih
     simp only [Fin.foldl_succ_last, Expression.eval]
     let vars_pre := vars.take pre |>.cast (by simp : min pre (pre + 1) = pre)
@@ -81,24 +82,39 @@ lemma foldl_isZero_eq_one_iff {n : ℕ} {vars : Vector (Expression F) n} {vals :
     specialize h_isZero pre trivial
     norm_num at h_isZero ⊢
     simp only [h_isZero, Fin.forall_fin_succ']
-    norm_num
-    split
-    · rename_i h
-      simp only [h]
-      simp only [implies_true, true_and, ← h_eval, Vector.getElem_map]
-    · rename_i h
-      simp only [h]
-      simp only [false_and, ↓reduceIte]
+    split_ifs <;> try rfl
+    · rename_i h_smaller h_last h_all
+      apply False.elim
+      apply h_all
+      intro i
+      by_cases h : i < pre
+      · intro _
+        aesop
+      intro _
+      have : i = pre := by omega
+      aesop
+    · rename_i h_smaller h_last h_all
+      apply False.elim
+      apply h_last
+      aesop
+    · aesop
 
-theorem soundness : Soundness F (elaborated (M := M)) Assumptions Spec := by
+theorem soundness [DecidableEq (M F)] : Soundness F (elaborated (M := M)) Assumptions Spec := by
   circuit_proof_start
   simp only [explicit_provable_type, ProvableType.fromElements_eq_iff] at h_input
+  conv_rhs =>
+    arg 1
+    rw [Zero.toOfNat0, OfNat.ofNat]
+    simp only [Zero.zero]
+    rw [ProvableType.fromElements_eq_iff']
+    rw [Vector.ext_iff]
+    simp only [Vector.getElem_replicate]
   apply foldl_isZero_eq_one_iff <;> assumption
 
 theorem completeness : Completeness F (elaborated (M := M)) Assumptions := by
   circuit_proof_start [IsZeroField.circuit, IsZeroField.Assumptions]
 
-def circuit : FormalCircuit F M field := {
+def circuit [DecidableEq (M F)] : FormalCircuit F M field := {
   elaborated with Assumptions, Spec, soundness, completeness
 }
 
--- a/Clean/Table/Inductive.lean
+++ b/Clean/Table/Inductive.lean
@@ -37,7 +37,7 @@ def InductiveTable.Completeness (F : Type) [Field F] (State Input : Type → Typ
   env.UsesLocalWitnessesCompleteness ((size State) + (size Input)) (step acc_var x_var |>.operations ((size State) + (size Input))) →
   -- assuming the spec on the current row, the input_spec on the input, and initial state assumptions
   InitialStateAssumptions initialState ∧
-  (∀ prev_row_index (h: prev_row_index < row_index), InputAssumptions prev_row_index (xs[prev_row_index]'(by simp only [xs_len, h]))) ∧ Spec initialState xs row_index xs_len acc ∧ InputAssumptions row_index x →
+  Spec initialState xs row_index xs_len acc ∧ InputAssumptions row_index x →
   -- the constraints hold
   Circuit.ConstraintsHold.Completeness env (step acc_var x_var |>.operations ((size State) + (size Input)))
 
--- a/Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
+++ b/Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
@@ -114,11 +114,8 @@ instance : ProvableStruct BlockInput where
     intros
     rfl
 
-/--
-Predicate that all components of BlockInput are well-formed.
--/
 def BlockInput.Normalized (input : BlockInput (F p)) : Prop :=
-  (input.block_exists = 0 ∨ input.block_exists = 1) ∧
+  IsBool input.block_exists ∧
   (∀ i : Fin 16, input.block_data[i].Normalized)
 
 namespace BLAKE3StateFirstHalf
@@ -439,7 +436,7 @@ lemma completeness : InductiveTable.Completeness (F p) ProcessBlocksState BlockI
     dsimp only [BlockInput.Normalized, ProcessBlocksState.Normalized] at h_assumptions
     dsimp only [IsZero.circuit, IsZero.Assumptions, BLAKE3.Compress.circuit, BLAKE3.Compress.Assumptions, BLAKE3.ApplyRounds.Assumptions]
     constructor
-    · have : x_block_exists = 0 ∨ x_block_exists = 1 := by tauto
+    · have : IsBool x_block_exists := by simp [h_assumptions]
       cases this with
       | inl h =>
           simp only [h]
--- a/Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
+++ b/Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
@@ -306,7 +306,7 @@ private lemma step_process_block (env : Environment (F p))
     simp only [Nat.ofNat_pos, and_true, true_and]
     constructor
     · linarith
-    · simp only [Expression.eval, chunkStart]
+    · simp only [circuit_norm, chunkStart]
       split at h_iszero
       · norm_num at h_iszero ⊢
         simp only [h_iszero, ZMod.val_one]
@@ -331,37 +331,31 @@ private lemma step_process_block (env : Environment (F p))
   simp only [h_vector_cond, h_u32_cond] at h_addition ⊢
   dsimp only [BLAKE3StateFirstHalf.circuit] at h_addition ⊢
   simp only [h_first_half, Addition32.circuit]
+  simp only [ProcessBlocksState.Normalized] at ⊢ acc_Normalized
+  simp only [ProcessBlocksState.toChunkState] at ⊢ h_addition blocks_compressed_not_many
+  dsimp only [BLAKE3.BLAKE3State.value] at h_compress
   constructor
-  · simp only [ProcessBlocksState.toChunkState, ↓reduceIte, id_eq, Nat.reduceMul, List.sum_cons, List.sum_nil, add_zero,
+  · simp only [↓reduceIte, id_eq, Nat.reduceMul, List.sum_cons, List.sum_nil, add_zero,
       Nat.reduceAdd, Vector.take_eq_extract, Vector.map_extract, Pi.zero_apply] at ⊢ h_addition
     simp only [h_addition, processBlockWords]
     clear h_addition
     norm_num at ⊢ h_compress
+    simp only [h_compress.1]
+    simp only [startFlag, U32_blockLen_value, circuit_norm]
+    norm_num at h_iszero
+    simp only [mul_zero, add_zero, id_eq]
+    simp only [circuit_norm, BLAKE3BlockInputNormalized.circuit] at ⊢ h_iszero
+    norm_num
     constructor
-    · dsimp only [BLAKE3.BLAKE3State.value] at h_compress
-      simp only [h_compress.1]
-      clear h_compress
-      simp only [startFlag, U32_blockLen_value, circuit_norm]
-      norm_num at h_iszero
-      simp only [mul_zero, add_zero, id_eq]
-      simp only [circuit_norm, BLAKE3BlockInputNormalized.circuit] at ⊢ h_iszero
-      norm_num
-      rw [eval_acc_blocks_compressed env (acc_chaining_value:=acc_chaining_value) (acc_chunk_counter:=acc_chunk_counter)]
-      simp only [ProcessBlocksState.Normalized] at acc_Normalized
-      simp only [IsZero.circuit, circuit_norm]
-      simp only [h_iszero]
+    · rw [eval_acc_blocks_compressed env (acc_chaining_value:=acc_chaining_value) (acc_chunk_counter:=acc_chunk_counter)]
+      · simp_all only [IsZero.circuit, circuit_norm]
       · simp_all [circuit_norm, IsZero.circuit]
-      · simp_all
-      · simp_all
-    · simp only [ProcessBlocksState.toChunkState] at blocks_compressed_not_many
-      simp only [circuit_norm]
-      omega
-  · simp only [ProcessBlocksState.Normalized]
-    constructor
+      · simp_all [ProcessBlocksState.Normalized]
+    · omega
+  · constructor
     · aesop
     · simp only [↓reduceIte, Nat.reduceMul, List.sum_cons, List.sum_nil, add_zero,
         Nat.reduceAdd, id_eq, Pi.zero_apply] at ⊢ h_addition
-      dsimp only [ProcessBlocksState.Normalized] at acc_Normalized
       simp [h_addition.2, acc_Normalized]
 
 /--
--- a/Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
+++ b/Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
@@ -30,7 +30,8 @@ private lemma ZMod_val_64 :
   have := p_large.elim
   linarith
 
-attribute [local circuit_norm] blockLen ZMod.val_zero ZMod.val_one ZMod_val_64 add_zero zero_add chunkStart -- only in the current section
+attribute [local circuit_norm] blockLen ZMod.val_zero ZMod.val_one ZMod_val_64 add_zero zero_add chunkStart List.concat_eq_append List.length_append List.length_cons List.length_nil
+  List.concat_eq_append -- only in the current section
 
 private lemma U32_blockLen_value (env : Environment (F p)) :
     (eval (α := U32) env { x0 := Expression.const 64, x1 := 0, x2 := 0, x3 := 0 }).value = 64 := by
@@ -351,12 +352,10 @@ lemma soundness : InductiveTable.Soundness (F p) ProcessBlocksState BlockInput S
   intro initialState row_index env acc_var x_var acc x xs xs_len h_eval h_holds spec_previous initial_Normalized inputs_short
   specialize spec_previous (by assumption)
   specialize spec_previous (by
-    simp only [List.concat_eq_append, List.length_append, List.length_cons, List.length_nil,
-      Nat.reducePow] at inputs_short
+    simp only [circuit_norm] at inputs_short
     omega)
-  rw [List.concat_eq_append, List.filter_append]
-  have : (xs.concat x)[row_index]'(by simp_all) = x := by simp_all
-  simp only [this]
+  simp only [circuit_norm]
+  rw [List.filter_append]
   have : (List.take row_index xs) = xs := by simp_all
   simp only [this] at spec_previous
   have input_normalized : x.Normalized := by
@@ -376,21 +375,22 @@ lemma soundness : InductiveTable.Soundness (F p) ProcessBlocksState BlockInput S
     have one_op := step_process_block env acc_var x_var acc x h_eval h_x h_holds
       spec_previous.2.2.2 input_normalized
         (by
-          simp only [List.concat_eq_append, List.length_append, List.length_cons, List.length_nil,
+          simp only [List.concat_eq_append, circuit_norm,
             Nat.reducePow] at inputs_short
           omega)
+    simp only [circuit_norm] at one_op
     simp only [one_op]
     constructor
-    · simp only [processBlockWords, List.concat_eq_append, List.length_append, List.length_cons, List.length_nil,
+    · simp only [processBlockWords, List.concat_eq_append, circuit_norm,
         add_lt_add_iff_right]
       omega
     simp [spec_previous, List.map_append, List.map_cons, List.map_nil, processBlocksWords, List.foldl_append, List.foldl_cons, List.foldl_nil]
   · simp only [h_x, decide_false, cond_false, List.append_nil]
     have no_op := step_skip_block env acc_var x_var acc x (by aesop) h_eval h_x h_holds
+    simp only [circuit_norm] at no_op
     simp only [no_op]
     constructor
-    · simp only [List.concat_eq_append, List.length_append, List.length_cons, List.length_nil]
-      omega
+    · omega
     · simp [spec_previous]
 
 def InitialStateAssumptions (initialState : ProcessBlocksState (F p)) := initialState.Normalized
@@ -400,7 +400,7 @@ def InputAssumptions (i : ℕ) (input : BlockInput (F p)) :=
 
 lemma completeness : InductiveTable.Completeness (F p) ProcessBlocksState BlockInput InputAssumptions InitialStateAssumptions Spec step := by
     intro initialState row_index env acc_var x_var acc x xs xs_len h_eval h_witnesses h_assumptions
-    dsimp only [InitialStateAssumptions, InputAssumptions] at *
+    dsimp only [InitialStateAssumptions, InputAssumptions, Addition32.Assumptions] at *
     rcases h_assumptions with ⟨ h_init, ⟨ h_assumptions, ⟨ h_input, h_small ⟩ ⟩ ⟩
     specialize h_assumptions (by assumption)
     have := p_large.elim
@@ -461,10 +461,9 @@ lemma completeness : InductiveTable.Completeness (F p) ProcessBlocksState BlockI
           omega
         constructor
         · trivial
-        constructor
-        · simp only [circuit_norm]
-          omega
         simp only [circuit_norm]
+        constructor
+        · omega
         constructor
         · split at h_witnesses_iszero
           · simp only [h_witnesses_iszero]
@@ -475,7 +474,7 @@ lemma completeness : InductiveTable.Completeness (F p) ProcessBlocksState BlockI
             norm_num
         · norm_num)
       simp only [h_compress]
-    simp_all [Addition32.circuit, Addition32.Assumptions, h_assumptions, circuit_norm, Conditional.circuit, Conditional.Assumptions, IsBool]
+    simp_all [Addition32.circuit, Addition32.Assumptions, h_assumptions, circuit_norm, Conditional.circuit, Conditional.Assumptions]
 
 /--
 The InductiveTable for processBlocks.
--- a/Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
+++ b/Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
@@ -31,7 +31,7 @@ private lemma ZMod_val_64 :
   linarith
 
 attribute [local circuit_norm] blockLen ZMod.val_zero ZMod.val_one ZMod_val_64 add_zero zero_add chunkStart List.concat_eq_append List.length_append List.length_cons List.length_nil
-  List.concat_eq_append -- only in the current section
+  -- only in the current section
 
 private lemma U32_blockLen_value (env : Environment (F p)) :
     (eval (α := U32) env { x0 := Expression.const 64, x1 := 0, x2 := 0, x3 := 0 }).value = 64 := by
@@ -375,13 +375,13 @@ lemma soundness : InductiveTable.Soundness (F p) ProcessBlocksState BlockInput S
     have one_op := step_process_block env acc_var x_var acc x h_eval h_x h_holds
       spec_previous.2.2.2 input_normalized
         (by
-          simp only [List.concat_eq_append, circuit_norm,
+          simp only [circuit_norm,
             Nat.reducePow] at inputs_short
           omega)
     simp only [circuit_norm] at one_op
     simp only [one_op]
     constructor
-    · simp only [processBlockWords, List.concat_eq_append, circuit_norm,
+    · simp only [processBlockWords, circuit_norm,
         add_lt_add_iff_right]
       omega
     simp [spec_previous, List.map_append, List.map_cons, List.map_nil, processBlocksWords, List.foldl_append, List.foldl_cons, List.foldl_nil]
--- a/Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
+++ b/Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
@@ -182,7 +182,6 @@ def step (state : Var ProcessBlocksState (F p)) (input : Var BlockInput (F p)) :
   -- Compute CHUNK_START flag (1 if blocks_compressed = 0, else 0)
   let isFirstBlock ← IsZero.circuit state.blocks_compressed
 
-  -- TODO: check state.Normalized if it's the first block -- maybe check always, because the circuit size would be similar?
   let startFlagU32 : Var U32 (F p) :=  ⟨isFirstBlock * (Expression.const (F:=F p) chunkStart), 0, 0, 0⟩
 
   -- Prepare constants
--- a/Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
+++ b/Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
@@ -172,15 +172,6 @@ def circuit : FormalAssertion (F p) BlockInput where
 
 end BLAKE3BlockInputNormalized
 
-omit p_large in
-lemma eval_vector_takeShort {M : TypeMap} [NonEmptyProvableType M] {n : ℕ} (env : Environment (F p))
-    (vars : Var (ProvableVector M n) (F p)) (i : ℕ) (h_i : i < n) :
-    (eval env (vars.takeShort i h_i) : ProvableVector _ _ _) = (eval env vars).takeShort i h_i := by
-  simp only [Vector.takeShort]
-  simp only [eval_vector]
-  ext j h_j
-  simp only [Vector.getElem_map, Vector.getElem_cast, Vector.map_take, Vector.getElem_map]
-
 attribute [local circuit_norm] eval_vector_takeShort Vector.map_takeShort
 
 /--
@@ -371,7 +362,7 @@ lemma soundness : InductiveTable.Soundness (F p) ProcessBlocksState BlockInput S
     simp only [x_block_exists_zero] at *
     simp only [Conditional.circuit, Conditional.Assumptions, Conditional.Spec, h_eval, step, circuit_norm] at h_holds ⊢
     simp only [step, circuit_norm, h_holds, h_eval, ProcessBlocksState.toChunkState] at ⊢ spec_previous
-    norm_num at h_holds ⊢ spec_previous
+    norm_num at h_holds ⊢
     simp_all only [circuit_norm]
     omega
 
@@ -404,25 +395,24 @@ lemma completeness : InductiveTable.Completeness (F p) ProcessBlocksState BlockI
       constructor
       · simp only [h_assumptions]
         trivial
+      simp_all only [circuit_norm]
       constructor
-      · simp only [circuit_norm]
-        omega
+      · omega
       constructor
-      · simp_all
-      constructor
-      · simp only [circuit_norm]
-        omega
+      · omega
       rcases h_witnesses with ⟨ h_witnesses_iszero, h_witnesses ⟩
       simp only [IsZero.circuit, IsZero.Assumptions] at h_witnesses_iszero
       specialize h_witnesses_iszero (by simp_all)
       simp only [IsZero.Spec] at h_witnesses_iszero
       constructor
       · split at h_witnesses_iszero
-        · simp only [h_witnesses_iszero]
+        · simp only [circuit_norm] at h_witnesses_iszero
+          simp only [h_witnesses_iszero]
           norm_num
           simp only [circuit_norm]
           omega
-        · simp only [h_witnesses_iszero]
+        · simp only [circuit_norm] at h_witnesses_iszero
+          simp only [h_witnesses_iszero]
           norm_num
       · norm_num
     simp_all only [Addition32.circuit, Addition32.Assumptions, Conditional.circuit, Conditional.Assumptions]
@@ -456,20 +446,13 @@ The InductiveTable for processBlocks.
 -/
 def table : InductiveTable (F p) ProcessBlocksState BlockInput where
   step
-
   Spec
-
   InitialStateAssumptions initialState := initialState.Normalized
-  InputAssumptions i input :=
-    input.Normalized ∧ i < 2^32
-
+  InputAssumptions i input := input.Normalized ∧ i < 2^32
   soundness
   completeness
   subcircuitsConsistent := by
-    intros
-    dsimp only [step]
-    simp only [circuit_norm]
+    simp only [step, circuit_norm]
     omega
-
 end
 end Tables.BLAKE3.ProcessBlocksInductive
--- a/Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
+++ b/Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
@@ -288,6 +288,8 @@ private lemma step_process_block (env : Environment (F p))
   simp only [step, circuit_norm] at ⊢ h_holds
   provable_struct_simp
   simp only [h_eval, h_x] at ⊢ h_holds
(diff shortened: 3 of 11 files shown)


Clean/Table/Inductive.lean
```diff
@@ -272,20 +274,20 @@ lemma table_soundness_aux (table : InductiveTable F State Input) (input output:
...
 def toFormal (table : InductiveTable F State Input) (input output: State F) : FormalTable F (ProvablePair State Input) where
   constraints := table.tableConstraints input output
-  Assumption N := N > 0 ∧ table.Spec 0 input [] rfl
-  Spec {N} trace := table.Spec (N-1) output (traceInputs trace.tail) (traceInputs_length trace.tail)
+  Assumption N := N > 0 ∧ table.Spec input [] 0 rfl input
+  Spec {N} trace := table.Spec input (traceInputs trace.tail) (N-1) (traceInputs_length trace.tail) output
 
   soundness N trace env assumption constraints :=
-    table.table_soundness input output ⟨N, assumption.left⟩ trace env assumption.right constraints
+    table_soundness table input output ⟨N, assumption.left⟩ trace env assumption.right constraints
```
pirapira: Is this change needed?

Clean/Gadgets/ConditionalU32.lean
```diff
@@ -0,0 +1,88 @@
...
+Uses the formula: result = cond * ifTrue + (1 - cond) * ifFalse
+-/
+def main (input : Var Inputs (F p)) : Circuit (F p) (Var U32 (F p)) := do
+  let { cond, ifTrue, ifFalse } := input
+
+  -- For each limb, compute: result_limb = cond * ifTrue_limb + (1 - cond) * ifFalse_limb
+  let x0 <== cond * ifTrue.x0 + (1 - cond) * ifFalse.x0
+  let x1 <== cond * ifTrue.x1 + (1 - cond) * ifFalse.x1
+  let x2 <== cond * ifTrue.x2 + (1 - cond) * ifFalse.x2
+  let x3 <== cond * ifTrue.x3 + (1 - cond) * ifFalse.x3
```
pirapira: It's more modular to define `F * U32` operation and `U32 + U32` operation.

[edit] It should be element-wise scaling and element-wise addition, so not to use too many constraints.

Clean/Gadgets/ConditionalU32.lean
```diff
@@ -0,0 +1,88 @@
...
+def Spec (input : Inputs (F p)) (output : U32 (F p)) : Prop :=
+  output = if input.cond = 1 then input.ifTrue else input.ifFalse
+
+omit [Fact (p > 512)] in
+theorem soundness : Soundness (F p) elaborated Assumptions Spec := by
+  intro offset env
+    ⟨ cond_var, ⟨ x0_var, x1_var, x2_var, x3_var ⟩, ⟨ y0_var, y1_var, y2_var, y3_var ⟩ ⟩
+    ⟨ cond, ⟨ x0, x1, x2, x3 ⟩, ⟨ y0, y1, y2, y3 ⟩ ⟩
+    h_eval h_assumptions h_holds
+  simp only [circuit_norm, Inputs.mk.injEq, U32.mk.injEq, explicit_provable_type] at h_eval
```
pirapira: `circuit_proof_start` is probably useful.

Clean/Gadgets/ConditionalU32.lean
```diff
@@ -0,0 +1,88 @@
...
+  simp only [main, circuit_norm, explicit_provable_type, Spec, Assumptions, h_eval] at *
+  simp only [h_holds]
+  rcases h_assumptions with h_asm|h_asm <;> simp [h_asm]
+
+omit [Fact (p > 512)] in
+theorem completeness : Completeness (F p) elaborated Assumptions := by
+  intro offset env input_var h_env_uses input h_eval h_assumptions
+  simp only [Assumptions] at h_assumptions
+  decompose_provable_struct
+  simp only [circuit_norm, Inputs.mk.injEq] at h_eval
```
pirapira: `circuit_proof_start` is probably useful.

Clean/Gadgets/IsZero.lean
```diff
@@ -0,0 +1,147 @@
...
+
+  -- The witnesses are exactly as specified
+  simp [h_eval] at h_wit_isZero h_wit_x_inv
+
+  refine ⟨?_, ?_, ?_⟩
+  · -- First constraint: isZero * x = 0
+    rw [h_wit_isZero]
+    by_cases hx : x = 0
+    · simp [hx]
+    · simp [hx]
```
pirapira: Since both cases are doing the same thing, can be refactored.

Clean/Gadgets/IsZero.lean
```diff
@@ -0,0 +1,147 @@
...
+  · -- Second constraint: isZero * (isZero - 1) = 0
+    rw [h_wit_isZero]
+    by_cases hx : x = 0
+    · simp only [circuit_norm] at h_eval ⊢
+      simp only [h_eval]
+      simp [hx]
+    · simp only [circuit_norm] at h_eval ⊢
+      simp only [h_eval]
+      simp [hx]
```
pirapira: Both cases are doing the same thing, so can be refactored.

Clean/Gadgets/IsZero.lean
```diff
@@ -0,0 +1,147 @@
...
+  · -- Third constraint: (1 - isZero) * x * x_inv = (1 - isZero)
+    simp only [circuit_norm] at h_eval ⊢
+    simp only [h_eval]
+    rw [h_wit_isZero, h_wit_x_inv]
+    by_cases hx : x = 0
+    · simp only [h_eval]
+      simp [hx]
+    · simp only [h_eval]
+      simp [hx]
```
pirapira: Both cases are doing the same thing, so can be refactored.

Clean/Gadgets/IsZeroU32.lean
```diff
@@ -0,0 +1,168 @@
...
+    (h_normalized : (U32.mk x0 x1 x2 x3).Normalized) :
+    x0 ≠ 0 → (U32.mk x0 x1 x2 x3).value ≠ 0 := by
+  intro h_nonzero h_value_eq_zero
+  -- If value = 0, then by injectivity, U32.mk x0 x1 x2 x3 = U32.mk 0 0 0 0
+  have h_eq : U32.mk x0 x1 x2 x3 = U32.mk 0 0 0 0 := by
+    apply U32.value_injective_on_normalized
+    · exact h_normalized
+    · simp [U32.Normalized]
+    · rw [h_value_eq_zero]
+      simp [U32.value]
```
pirapira: This part is common. Maybe state this part only, and then use it instead of the four similar lemmas.

Clean/Gadgets/IsZero.lean
pirapira: This has been refactored away.

Clean/Gadgets/IsZero.lean
pirapira: This has been refactored away.

Clean/Gadgets/IsZero.lean
pirapira: This has been refactored away.

Clean/Gadgets/ConditionalU32.lean
pirapira: Implemented #241 for scaling.

pirapira: Now being addressed in #242.

Clean/Gadgets/ConditionalU32.lean
pirapira: This file is gone.

Clean/Gadgets/ConditionalU32.lean
pirapira: This file is gone.

Clean/Gadgets/IsZeroU32.lean
pirapira: The four similar theorems are now gone.

Clean/Gadgets/ConditionalU32.lean
pirapira: This PR uses the conditional gadget in #242 .

Clean/Gadgets/IsZeroU32.lean
```diff
@@ -0,0 +1,86 @@
...
+
+namespace Gadgets.IsZeroU32
+
+variable {p : ℕ} [Fact p.Prime] [p_large_enough: Fact (p > 512)]
+
+/--
+Main circuit that checks if a U32 (32-bit unsigned integer) is zero.
+Returns 1 if all limbs are 0, otherwise returns 0.
+-/
+def main (x : Var U32 (F p)) : Circuit (F p) (Var field (F p)) := do
```
pirapira: `IsZero` can be defined for all `ProvableType`s generally. Similarly to #242 .

pirapira: Now this PR is using the generic IsZero from #247

Clean/Table/Inductive.lean
```diff
@@ -9,6 +9,38 @@ Thus far, only the common `k=2` case is handled.
 import Clean.Table.Theorems
 import Clean.Gadgets.Equality
 
+def InductiveTable.Soundness (F : Type) [Field F] (State Input : Type → Type) [ProvableType State] [ProvableType Input]
+    (Spec : (initialState : State F) → (xs : List (Input F)) → (i : ℕ) → (xs.length = i) → (currentState : State F) → Prop)
```
mitschabaude, line 13: nit: can we move the `currentState` argument to the beginning? it's sort of the "main" argument that the Spec is about, and the one that will always be used.

or I don't know, maybe putting the important arguments at the beginning is a JS instinct. but it's more easily readable that way

Clean/Table/Inductive.lean
```diff
@@ -9,6 +9,38 @@ Thus far, only the common `k=2` case is handled.
...
+  ∀ (initialState : State F) (row_index : ℕ) (env : Environment F),
+  -- for all rows and inputs
+  ∀ (acc_var : Var State F) (x_var : Var Input F)
+    (acc : State F) (x : Input F) (xs : List (Input F)) (xs_len : xs.length = row_index),
+    (eval env acc_var = acc) ∧ (eval env x_var = x) →
+  -- when using honest-prover witnesses
+  env.UsesLocalWitnessesCompleteness ((size State) + (size Input)) (step acc_var x_var |>.operations ((size State) + (size Input))) →
+  -- assuming the spec on the current row, the input_spec on the input, and initial state assumptions
+  InitialStateAssumptions initialState ∧
+  (∀ prev_row_index (h: prev_row_index < row_index), InputAssumptions prev_row_index (xs[prev_row_index]'(by simp only [xs_len, h]))) ∧ Spec initialState xs row_index xs_len acc ∧ InputAssumptions row_index x →
```
mitschabaude: do we need the input assumptions on all previous rows here? the current circuit is not really supposed to depend on previous inputs

Clean/Tables/Fibonacci32Inductive.lean
```diff
@@ -25,16 +25,24 @@ def table : InductiveTable (F p) Row unit where
     let z ← Addition32.circuit { x := row.x, y := row.y }
     return { x := row.y, y := z }
 
-  Spec i row _ _ : Prop :=
+  Spec initialState _ i _ row : Prop :=
```
mitschabaude: nit: leave unused arguments blank _

Clean/Tables/Fibonacci32Inductive.lean
```diff
@@ -25,16 +25,24 @@ def table : InductiveTable (F p) Row unit where
...
     row.x.value = fib32 i ∧
     row.y.value = fib32 (i + 1) ∧
     row.x.Normalized ∧ row.y.Normalized
 
-  soundness := by simp_all [fib32, circuit_norm,
-    Addition32.circuit, Addition32.Assumptions, Addition32.Spec]
+  InitialStateAssumptions _ := True
 
-  completeness := by simp_all [fib32, circuit_norm,
-    Addition32.circuit, Addition32.Assumptions, Addition32.Spec]
```
mitschabaude: can you move back to the old simpler proofs by passing in `InductiveTable.Soundness`/`Completeness`?

Clean/Types/U32.lean
```diff
@@ -181,6 +181,175 @@ lemma fromByte_normalized {x : Fin 256} : (fromByte x).Normalized (p:=p) := by
...
+  simp_all
+
+end ValueInjectivity
+
+omit p_large_enough in
+/--
+Lemma showing that U32 Normalized property is equivalent to all components being < 256
+-/
+@[circuit_norm]
+lemma Normalized_componentwise (env : Environment (F p)) (a b c d : Var field (F p)):
```
mitschabaude: this should be spelled lower-case

Clean/Types/U32.lean
```diff
@@ -181,6 +181,175 @@ lemma fromByte_normalized {x : Fin 256} : (fromByte x).Normalized (p:=p) := by
...
+  cases x; cases y
+  simp_all
+
+end ValueInjectivity
+
+omit p_large_enough in
+/--
+Lemma showing that U32 Normalized property is equivalent to all components being < 256
+-/
+@[circuit_norm]
```
mitschabaude, line 212: so I guess it's ok to add this to `circuit_norm`, because it will only be applied if the U32 is already decomposed?

mitschabaude: but I want to question if we really need this. the lemma is just the definition of `U32.Normalized` + field access commuting with `eval`. It feels like we should achieve this automatically with the techniques of the `simp_provable_struct` tactic, for _all_ provable types and not just U32.

Clean/Types/U32.lean
```diff
@@ -181,6 +181,175 @@ lemma fromByte_normalized {x : Fin 256} : (fromByte x).Normalized (p:=p) := by
...
+lemma Normalized_componentwise (env : Environment (F p)) (a b c d : Var field (F p)):
+    (eval (α := U32) env
+    { x0 := a, x1 := b, x2 := c, x3 := d }).Normalized ↔
+    ((eval env a).val < 256 ∧ (eval env b).val < 256 ∧ (eval env c).val < 256 ∧ (eval env d).val < 256) := by
```
mitschabaude: maybe it's cleaner to state this as
```lean
(eval env (U32.mk a b c d)).Normalized ↔
  (U32.mk (eval env a) (eval env b) (eval env c) (eval env d)).Normalized
```

Clean/Types/U32.lean
```diff
@@ -181,6 +181,175 @@ lemma fromByte_normalized {x : Fin 256} : (fromByte x).Normalized (p:=p) := by
...
+  simp only [Parser.Attr.explicit_provable_type, ProvableType.eval, fromElements, toVars, toElements, Vector.map]
+  simp only [List.map_toArray, List.map_cons, List.map_nil, U32.Normalized]
+
+/--
+Lemma: For a normalized U32, value = 0 iff all components are 0
+-/
+lemma value_zero_iff_components_zero {x : U32 (F p)} (hx : x.Normalized) :
+    x.value = 0 ↔ (∀ i : Fin (size U32), (toElements x)[i] = 0) := by
```
mitschabaude: wouldn't it be simpler to state this as
```lean
x.value = 0 ↔ x = .mk 0 0 0 0
```

in any case, we should always use `toLimbs` (the U32-specific concept) rather than `toElements` for consistency with other theory here

Clean/Types/U32.lean
```diff
@@ -181,6 +181,175 @@ lemma fromByte_normalized {x : Fin 256} : (fromByte x).Normalized (p:=p) := by
...
+    simp only [this] at h_3
+    simp only [List.getElem_cons_succ, List.getElem_cons_zero] at h_3
+    simp only [h_0, h_1, h_2, h_3, ZMod.val_zero]
+    norm_num
+
+lemma constU32_value (env : Environment (F p)) (n0 n1 n2 n3 : ℕ)
+    (h0 : n0 < p) (h1 : n1 < p) (h2 : n2 < p) (h3 : n3 < p) :
+    (eval (α := U32) env { x0 := Expression.const ↑n0, x1 := Expression.const ↑n1,
+                           x2 := Expression.const ↑n2, x3 := Expression.const ↑n3 }).value =
+    n0 + n1 * 256 + n2 * 256^2 + n3 * 256^3 := by
```
mitschabaude: again this feels like an overly specific application of the general `eval_const` lemma for all provable types

Clean/Types/U32.lean
```diff
@@ -181,6 +181,175 @@ lemma fromByte_normalized {x : Fin 256} : (fromByte x).Normalized (p:=p) := by
...
+
+lemma one_value (env : Environment (F p)) :
+    (eval (α := U32) env { x0 := 1, x1 := 0, x2 := 0, x3 := 0 }).value = 1 := by
+  have := p_large_enough.elim
+  have h := constU32_value env 1 0 0 0 (by omega) (by omega) (by omega) (by omega)
+  simp at h
+  exact h
+
+lemma const_value (env : Environment (F p)) (n : ℕ) (h : n < 256) :
+    (eval (α := U32) env { x0 := Expression.const ↑n, x1 := 0, x2 := 0, x3 := 0 }).value = n := by
```
mitschabaude: really not a fan of all these U32 specific lemmas which `simp only [circuit_norm, explicit_provable_type]`  should prove for _any_ provable type

Clean/Tables/KeccakInductive.lean
```diff
@@ -14,20 +14,23 @@ def table : InductiveTable (F p) KeccakState KeccakBlock where
...
     AbsorbBlock.circuit { state, block }
 
-  Spec i state blocks _ : Prop :=
+  Spec initialState blocks i _ state : Prop :=
     state.Normalized
     ∧ state.value = absorbBlocks (blocks.map KeccakBlock.value)
 
   InputAssumptions i block := block.Normalized
+  
+  InitialStateAssumptions _ := True
```
mitschabaude: can we make this the default, so you don't always need to add `InitialStateAssumptions`?

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
```diff
@@ -0,0 +1,555 @@
...
+  components := [ProvableVector U32 8, U32, U32]
+  toComponents := fun { chaining_value, chunk_counter, blocks_compressed } =>
+    .cons chaining_value (.cons chunk_counter (.cons blocks_compressed .nil))
+  fromComponents := fun xss =>
+    match xss with
+    | .cons cv (.cons cc (.cons bc .nil)) =>
+      { chaining_value := cv, chunk_counter := cc, blocks_compressed := bc }
+  fromComponents_toComponents := by
+    intros
+    rfl
```
mitschabaude: wait, isn't that the default tactic?

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
```diff
@@ -0,0 +1,555 @@
...
+Predicate that all components of BlockInput are well-formed.
+-/
+def BlockInput.Normalized (input : BlockInput (F p)) : Prop :=
+  (input.block_exists = 0 ∨ input.block_exists = 1) ∧
+  (∀ i : Fin 16, input.block_data[i].Normalized)
+
+namespace BLAKE3StateFirstHalf
+
+def main (x : Var Gadgets.BLAKE3.BLAKE3State (F p)) : Circuit (F p) (Var (ProvableVector U32 8) (F p)) := do
+  return x.take 8
```
mitschabaude: this has no circuit operations, why is it a circuit?

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
```diff
@@ -0,0 +1,555 @@
...
+-/
+def step (state : Var ProcessBlocksState (F p)) (input : Var BlockInput (F p)) :
+    Circuit (F p) (Var ProcessBlocksState (F p)) := do
+
+  -- Add constraint that block_exists is boolean
+  input.block_exists * (input.block_exists - 1) === 0
+
+  -- Compute CHUNK_START flag (1 if blocks_compressed = 0, else 0)
+  let isFirstBlock ← IsZero.circuit state.blocks_compressed
+  let startFlagU32 : Var U32 (F p) := ⟨Expression.mul isFirstBlock (Expression.const chunkStart), 0, 0, 0⟩
```
mitschabaude: why not
```suggestion
  let startFlagU32 : Var U32 (F p) := ⟨isFirstBlock * chunkStart, 0, 0, 0⟩
```

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
```diff
@@ -0,0 +1,555 @@
...
+end BLAKE3StateFirstHalf
+
+/--
+The step function that processes one block or passes through the state.
+-/
+def step (state : Var ProcessBlocksState (F p)) (input : Var BlockInput (F p)) :
+    Circuit (F p) (Var ProcessBlocksState (F p)) := do
+
+  -- Add constraint that block_exists is boolean
+  input.block_exists * (input.block_exists - 1) === 0
```
mitschabaude: would be nicer to use
```suggestion
  assertBool input.block_exists
```

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
```diff
@@ -0,0 +1,555 @@
...
+        omega
+      · simp [spec_previous]
+  · aesop
+
+def InitialStateAssumptions (initialState : ProcessBlocksState (F p)) := initialState.Normalized
+
+def InputAssumptions (i : ℕ) (input : BlockInput (F p)) :=
+    input.Normalized ∧ i < 2^32
+
+set_option maxHeartbeats 300000 in
```
mitschabaude: eww 😢

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
```diff
@@ -0,0 +1,555 @@
...
+      { block_exists := block_exists, block_data := data }
+  fromComponents_toComponents := by
+    intros
+    rfl
+
+/--
+Predicate that all components of BlockInput are well-formed.
+-/
+def BlockInput.Normalized (input : BlockInput (F p)) : Prop :=
+  (input.block_exists = 0 ∨ input.block_exists = 1) ∧
```
mitschabaude: why not
```suggestion
  IsBool input.block_exists ∧
```

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
```diff
@@ -0,0 +1,555 @@
...
+    blocks_compressed := muxedBlocksCompressed
+  }
+
+def Spec (initialState : ProcessBlocksState (F p)) (inputs : List (BlockInput (F p))) i (_ : inputs.length = i) (state : ProcessBlocksState (F p)) :=
+    initialState.Normalized →
+    (∀ input ∈ inputs, input.Normalized) →
+    inputs.length < 2^32 →
+    -- The spec relates the current state to the mathematical processBlocksWords function
+    -- applied to the first i blocks from inputs (where block_exists = 1)
+    let validBlocks := inputs.take i |>.filter (·.block_exists = 1)
```
mitschabaude: why do you `.take i` when we have the hypothesis that `inputs.length = i`?

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
```diff
@@ -0,0 +1,555 @@
...
+    -- Extract the word data directly - no conversion needed!
+    let blockWords := validBlocks.map (fun b => b.block_data.map (·.value))
```
mitschabaude: that comment sounds like an LLM talking to itself XD
```suggestion
    let blockWords := validBlocks.map (fun b => b.block_data.map (·.value))
```

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
```diff
@@ -0,0 +1,555 @@
...
+    selector := input.block_exists
+    ifTrue := newState.chaining_value
+    ifFalse := state.chaining_value
+  }
+
+  let muxedBlocksCompressed ← Conditional.circuit (M := U32) {
+    selector := input.block_exists
+    ifTrue := newBlocksCompressed
+    ifFalse := state.blocks_compressed
+  }
```
mitschabaude: why is this type argument needed? the left and right inputs are clearly U32
```suggestion
  let muxedBlocksCompressed ← Conditional.circuit {
    selector := input.block_exists
    ifTrue := newBlocksCompressed
    ifFalse := state.blocks_compressed
  }
```

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
```diff
@@ -0,0 +1,555 @@
...
+  -- Compute CHUNK_START flag (1 if blocks_compressed = 0, else 0)
+  let isFirstBlock ← IsZero.circuit state.blocks_compressed
+  let startFlagU32 : Var U32 (F p) := ⟨Expression.mul isFirstBlock (Expression.const chunkStart), 0, 0, 0⟩
+
+  -- Prepare constants
+  let zeroU32 : Var U32 (F p) := ⟨0, 0, 0, 0⟩
+  let blockLenU32 : Var U32 (F p) := ⟨(blockLen : F p), 0, 0, 0⟩
+
+  -- Prepare inputs for compress
+  let compressInput : Var Gadgets.BLAKE3.ApplyRounds.Inputs (F p) := {
```
mitschabaude: isn't `Gadgets` opened?
```suggestion
  let compressInput : Var BLAKE3.ApplyRounds.Inputs (F p) := {
```

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
```diff
@@ -0,0 +1,555 @@
...
+    chaining_value := state.chaining_value
+    block_words := input.block_data
+    counter_high := zeroU32
+    counter_low := state.chunk_counter
+    block_len := blockLenU32
+    flags := startFlagU32
+  }
+
+  -- Apply compress to get new chaining value
+  let newCV16 ← Gadgets.BLAKE3.Compress.circuit compressInput
```
mitschabaude: isn't `Gadgets` opened?
```suggestion
  let newCV16 ← BLAKE3.Compress.circuit compressInput
```

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
```diff
@@ -0,0 +1,555 @@
...
+  -- Increment blocks_compressed
+  let one32 : Var U32 (F p) := ⟨1, 0, 0, 0⟩
+  let newBlocksCompressed ← Addition32.circuit { x := state.blocks_compressed, y := one32 }
+
+  -- Create the new state (if block exists)
+  let newState : Var ProcessBlocksState (F p) := {
+    chaining_value := newCV8
+    chunk_counter := state.chunk_counter
+    blocks_compressed := newBlocksCompressed
+  }
```
mitschabaude: why is this struct created? it looks like we only ever take `newState.chaining_value` from it

Clean/Table/Inductive.lean
pirapira: In the completeness proof, `Spec` needs to be used. In the current example, `Spec` is defined with
```
    (∀ input ∈ inputs, input.Normalized) →
```
so, using `Spec` requires information about all the previous inputs.

If we remove
```
    (∀ input ∈ inputs, input.Normalized) →
```
from `Spec`, there's a problem in the soundness proof when the input has at least two elements. Soundness requires proving `Spec` of the current step using `Spec` of the previous  step. The `Spec` for the previous steps demands the previous input is `Normalized`. This is not available because the `Spec` of the current step now only assumes that the last input is `Normalized`.

Clean/Types/U32.lean
pirapira: >  it will only be applied if the U32 is already decomposed?

That's the expectation. Maybe I should try.

> if we really need this.

`simp` uses discrimination trees to optimize for bigger simpsets (aiming at less than linear time on the size of simpsets). `simp_provable_struct`, defined current way, runs every check on all subterms by default, so can grow slower as we put more features into it.

Clean/Types/U32.lean
pirapira: I'll try `.mk 0 0 0 0` first.

Clean/Types/U32.lean
pirapira: I guess there should be a simplification rule for `eval env (U32.mk _ _ _ _)`, and another simplification rule.for `(U32.mk _ _ _ _).value`.

Clean/Table/Inductive.lean
pirapira: Now I think about it's somewhat consistent with `Spec input output`.

Here we're looking at this as computation, with progress of time. First there's `initialState`, then we get inputs `xs` and do something and get `currentState` at the end. The order of arguments follows the time sequence.

Clean/Types/U32.lean
mitschabaude: > simp uses discrimination trees to optimize for bigger simpsets (aiming at less than linear time on the size of simpsets).

fair enough, but beyond simp efficiency, I think it's more elegant and easier to maintain to have a strategy that does something for all provable types, instead of reimplementing it for every individual type

pirapira: ~My take here is that `U32` should be a `ProvableStruct`. But I don't implement that in this PR.~ The only reason for implementing a custom `simp_provable_struct` is because Lean simp cannot name new variables, or it cannot split subgoals. Everything that can be done with a simplification lemma should be done with one.

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
pirapira: done.

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
pirapira: And then I removed the check. The surrounding circuit (in this case `updateChunk`) should establish the `IsBool` condition.

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
pirapira: The typeclass resolution wasn't clever enough so currently it's
`⟨isFirstBlock * (Expression.const (F:=F p) chunkStart)`

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
pirapira: This is gone and nothing else changed.

Clean/Table/Inductive.lean
mitschabaude: Ok I think you shouldn't actually have those assumptions in your `Spec`. If your local soundness proof needs the fact that the input is normalized, this just means the _circuit has to constrain that property_. (How else would we ever convince the verifier of this property of the inputs, which is a witness?)

Compare how the Keccak table adds a similar constraint on the input, and then doesn't need it as an assumption in the Spec: https://github.com/Verified-zkEVM/clean/blob/1cfc31f703f186e82285a75e797eebf09e3942ac/Clean/Tables/KeccakInductive.lean#L14

mitschabaude: to summarize, in my opinion, the need to add this property comes downstream of misstructuring statements in the Blake3 table, and should be reverted

pirapira: > Ok I think you shouldn't actually have those assumptions in your Spec.

I see.

The assumptions appeared because I wanted to use the table as a sub-circuit, and there was inertia about that. We have a different plan now, and the table won't share the environment with other circuits. I'll remove the assumptions, also will revert the `InductiveTable` type.

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
pirapira: This wasn't needed.

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
pirapira: These are now gone.

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
pirapira: These are now gone.

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
pirapira: This one and the one below are gone.

Clean/Types/U32.lean
pirapira: lowered.

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
```diff
@@ -0,0 +1,509 @@
...
+  linarith
+
+attribute [local circuit_norm] blockLen ZMod.val_zero ZMod.val_one ZMod_val_64 -- only in the current section
+
+private lemma U32_blockLen_value (env : Environment (F p)) :
+    (eval (α := U32) env { x0 := Expression.const 64, x1 := 0, x2 := 0, x3 := 0 }).value = 64 := by
+  simp only [circuit_norm]
+  rw [U32.value_of_literal] -- why is this needed though value_of_literal is in circuit_norm?
+  simp only [circuit_norm]
+  omega
```
mitschabaude: simpler proof that doesn't need `value_of_literal`:
```suggestion
private lemma U32_blockLen_value (env : Environment (F p)) :
    (eval (α := U32) env { x0 := Expression.const 64, x1 := 0, x2 := 0, x3 := 0 }).value = 64 := by
  simp only [circuit_norm, U32.value]
  rfl
```

pirapira: The next step here is to tweak the statement of `U32.vavlue_of_literal` so that it does something within `simp only [circuit_norm]`. The step after that is to try to remove this lemma because by that time `circuit_norm` is strong enough to crunch this form.

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
```diff
@@ -0,0 +1,513 @@
...
+  return {
+    chaining_value := muxedCV
+    chunk_counter := state.chunk_counter  -- Never changes
+    blocks_compressed := muxedBlocksCompressed
+  }
+
+def Spec (initialState : ProcessBlocksState (F p)) (inputs : List (BlockInput (F p))) i (_ : inputs.length = i) (state : ProcessBlocksState (F p)) :=
+    initialState.Normalized → -- TODO: drop, assert in the circuit
+    (∀ input ∈ inputs, input.Normalized) → -- TODO: drop, assert in the circuit
+    inputs.length < 2^32 →
```
pirapira, line 230: I'm wondering whether the table needs a counter so that the number of rows can be limited by constraints.

That feels excessive because the verifier usually knows the dimension of tables.

Clean/Table/Inductive.lean
pirapira: I changed my mind on this.

Other circuits can constrain inputs to be nice, and lookup `(inputs, output)` from the table. In this combination, the table doesn't need to constrain the inputs to be nice.

Clean/Tables/Fibonacci32Inductive.lean
pirapira: done.

Clean/Types/U32.lean
pirapira: I think `circuit_norm` should contain one lemma like this for every structure. That's the idiomatic way across multiple interactive theorem provers especially in the simplification-centric approach.

The usual way to avoid repetition is to add an attribute
```
@[distribute_eval]
struct S { ... }
```
that automatically generates
```
eval env (S.mk _ _ ...) = S.mk (eval env _) (eval env _) ...
```

My personal preference is verbose repetition.

Clean/Types/U32.lean
pirapira: ^ isn't needed because
```
@[circuit_norm]
lemma eval_of_literal (env : Environment (F p)) (a b c d : Var field (F p)) :
    eval env (U32.mk a b c d) =
    U32.mk (eval env a) (eval env b) (eval env c) (eval env d) 
```
does exactly that simplfiication.

Clean/Types/U32.lean
pirapira: This has disappeared.

Clean/Types/U32.lean
pirapira: This has also disappeared.

Clean/Tables/KeccakInductive.lean
pirapira: The default existed and this line disappeared.

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
pirapira: removed these three lines.

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
pirapira: Just to split subproofs away from the big main proofs (which require bigger heartbeat).

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
pirapira: The `take` is gone together with a dozen lines in the proof.

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
pirapira: No longer wondering about this. All the assumptions will remain here.

Prover can put non-BLAKE in the table as long as one of the inputs are not normalized.

Verifier will the spec of `splitIntoBlockWords` to see the inputs are normalized, and then looks up entries from `ProcessBlocksInductive` table.

Clean/Table/Inductive.lean
pirapira: Let's allow the prover to put non-BLAKE entries as long as some entries are not normalized.

The verifier needs to use the spec of `splitIntoBlockWords` gadgets to make sure all inputs are normalized, and then look up `ProcessBlocksInductive` table.

Clean/Types/U32.lean
pirapira:
```
(∀ i : Fin (size U32), (toElements x)[i] = 0)
```
comes from `IsZero.Spec`.

This suggests `IsZero.Spec` should be rewritten in terms of `input = fromElements ...`.

pirapira: I'm modifying `IsZero.Spec` in #257.

Clean/Table/Inductive.lean
mitschabaude: > Other circuits can constrain inputs to be nice, and lookup (inputs, output) from the table. In this combination, the table doesn't need to constrain the inputs to be nice.

I agree this should be possible in principle, but `InductiveTable` specifically is not designed to be "looked up" into. It's designed to be a final, end-to-end interface.

If we want to change that, and let `InductiveTable.Soundness` accept input assumptions, then we should add those input assumptions as a statement separate from the `Spec`. Adding them to the Spec is wrong, because you get this awkward inductive step.

Clean/Types/U32.lean
mitschabaude: I'm pretty sure this can be done with simplification lemmas that work for all provable types!

mitschabaude: the question is not if it can be done, but _when_ it should be done. we don't always want to decompose provable types, but when we do want it, just adding `explicit_provable_type` to our simp set should do the job

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
mitschabaude: it's already strong enough if you add `U32.value`

pirapira: The game I was playing here is:
* to avoid adding `U32.value` in `circuit_norm`
* to avoid the need to say `U32.value` in simp arguments
* to make `eval env { x0 := ...}` reduced automatically with `circuit_norm`.

To be honest, I'm not doing this consistently. I think I did `[circuit_norm, U32.value]` in other places.

mitschabaude: makes sense!

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
pirapira: I'll drop these assumptions, so that `ProcessBlocksInductive` table makes sense stand-alone.

edit: though `input.length < 2^32` will remain.

Clean/Types/U32.lean
pirapira: If we are already seeing `S.mk _ _ _ _` with elements, always distributing `eval env` will make a normal form. The goal state will be slightly bigger with repeated `eval env` but that's rather small.

If we are just seeing `evan env someStruct` with a variable, that alone won't trigger the simplification.

Clean/Table/Inductive.lean
pirapira: The awkwardness came from `Spec` being `SomethingBig -> SomethingUseful`. Every use of `Spec` involved `SomethingBig`, and that's burdensome.

I'll make the circuit check the input against `InputAssumption`.

And then,
```
Spec initialState xs row_index xs_len acc
```
inplies all inputs in `xs` satisfy `InputAssumption`.

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
pirapira: Now these lines say
```
    initialState.Normalized →
    inputs.length < 2^32 →
    (∀ input ∈ inputs, input.Normalized) ∧
```

The condition on the `initialState` still remains. It doesn't cause awkward induction steps though.

Clean/Table/Inductive.lean
pirapira: Now the awkward induction is gone.

Clean/Tables/Fibonacci32Inductive.lean
pirapira: Reverted.

Clean/Types/U32.lean
pirapira: Now I have
```
@[circuit_norm]
lemma value_zero_iff_zero {x : U32 (F p)} (hx : x.Normalized) :
    x.value = 0 ↔ x = U32.mk 0 0 0 0
```

I decided to reduce `0 : U32 F` to `U32.mk 0 0 0 0`. The result plays nicely with the existing simplification lemmas about `eval env (U32.mk a b c d)`.

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
pirapira: The soundness proof this circuit is basically dealing with:
```
eval env (Vector.take ...
```

I'll try adding a lemma about the above form in `circuit_norm`.

When the soundness proof becomes just several lines, I'm comfortable inlining it.

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
pirapira: I added back this check together with the check all input blocks are normalized.

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
pirapira: This is gone.

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
pirapira: This lemma is gone. The key seems to be adding `id_eq` in `local circuit_norm`.

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
pirapira: It's easy to make the circuit check `initialState.Normalized`. Maybe I'll do that.

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
pirapira: The main benefit of this subcircuit is that the casting business
`Vector _ 8` vs `Vector _ (min 8 16)`
is isolated in this small proof script.

pirapira: Maybe I can use a variant of `Vector.take` that doesn't result in `Vector _ (min _ _)` (but with an extra hypothesis that take argument is not big).

pirapira: I was able to remove the take-first-eight subcircuit.

I made a variant of `Vector.take` called `v.takeShort j` where `j` has to be smaller than the length of `v`. The nice thing is that the type of `v.takeShort j` is `Vector _ j` instead of `Vector _ (min _ j)`.

Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
pirapira: Now the only assumption is
```
inputs.length < 2^32 →
```

I'll leave it.

Clean/Table/Inductive.lean
pirapira: Also arguably the main argument should come at the end because
`pred x0 x1 x2 main` can be seen as `pred_about_main main` where `pred_about_main = pred x0 x1 x2`.

mitschabaude (approved):

PR #242, opened 2025-07-31
Add generic conditional selector on ProvableType

pirapira:
## Summary

This PR introduces conditional binary multiplexer for `ProvableType`s.

## Changes

   - `Conditional`: Selects between two ProvableTypes based on a binary selector

## Test plan
- [x] All files compile without errors


Clean/Gadgets/ElementwiseScalarMul.lean
```diff
@@ -0,0 +1,135 @@
...
+    simp only [h_spec]
+    simp only [Vector.getElem_map]
+    simp only [h_one]
+    simp only [one_mul]
+
+/--
+Binary scalar multiplication circuit with stronger assumptions and guarantees.
+-/
+def binaryCircuit : FormalCircuit F (Inputs M) M :=
+  (circuit (F := F) (M := M)).strengthenAssumption
```
pirapira: weakenSpec suffices

Clean/Circuit/StructuralLemmas.lean
```diff
@@ -99,3 +99,59 @@ lemma FormalCircuit.weakenSpec_assumptions {F Input Output} [Field F] [ProvableT
...
+and potentially a different specification. This is useful when:
+- You need to add preconditions to a circuit
+- You want to specialize a general circuit for a specific use case
+- The stronger assumptions enable proving additional properties in the spec
+
+The requirements are:
+- The stronger assumptions must imply the original assumptions
+- The stronger assumptions and original spec together must imply the new spec
+-/
+def FormalCircuit.strengthenAssumption
```
pirapira: I think strengthenAssumption is not needed for now.

Clean/Gadgets/Conditional.lean
```diff
@@ -0,0 +1,114 @@
...
+  let scaledFalse ← Gadgets.ElementwiseScalarMul.binaryCircuit { scalar := 1 - selector, data := ifFalse }
+
+  -- Add them together (using circuitWithZeroSpec which handles zero cases)
+  ElementwiseAdd.circuitWithZeroSpec { a := scaledTrue, b := scaledFalse }
+
+/--
+Assumes the selector is binary (0 or 1).
+-/
+def Assumptions (input : Inputs M F) : Prop :=
+  IsBool input.selector
```
pirapira: I think this can be just True.

Clean/Circuit/Provable.lean
```diff
@@ -120,7 +126,7 @@ def varFromOffset (α : TypeMap) [ProvableType α] (offset : ℕ) : Var α F :=
 attribute [explicit_provable_type] Vector.mapRange_succ Vector.mapRange_zero
 end ProvableType
 
-export ProvableType (eval const varFromOffset)
+export ProvableType (eval const zero zeroVar varFromOffset)
```
pirapira: `export`ing `zero` is probably not a good idea. The name is too generic.

pirapira: This turned into `allZero`.

Clean/Gadgets/Conditional.lean
```diff
@@ -0,0 +1,98 @@
...
+  ElementwiseAdd.circuitWithZeroSpec { a := scaledTrue, b := scaledFalse }
+
+def Assumptions (_ : Inputs M F) : Prop := True
+
+/--
+Specification: If selector is 1, output equals ifTrue; if selector is 0, output equals ifFalse.
+-/
+def Spec (input : Inputs M F) (output : M F) : Prop :=
+  (input.selector = 1 → output = input.ifTrue) ∧
+  (input.selector = 0 → output = input.ifFalse)
```
pirapira: Probably `output = if ... then ... else ...` is better.

pirapira: however, that requires dealing with the case `input.selector` is not 0 or 1.

mitschabaude: I think the gadget should just assume it's 0 or 1 (it won't be used in any other way)

pirapira: Now it assumes `IsBool` and the spec is if then else.

Clean/Circuit/Provable.lean
```diff
@@ -118,9 +124,38 @@ def varFromOffset (α : TypeMap) [ProvableType α] (offset : ℕ) : Var α F :=
...
+-/
+def elementwiseAdd [Field F] (a b : α F) : α F :=
+  fromElements (Vector.ofFn fun i => (toElements a)[i] + (toElements b)[i])
+
+/--
+Element-wise scalar multiplication for ProvableTypes.
+Multiplies each element by a scalar field element. No carries. Each element wraps around.
+-/
+def elementwiseScalarMul [Field F] (s : F) (v : α F) : α F :=
+  fromElements ((toElements v).map (s * ·))
```
mitschabaude: very clean definitions!

Clean/Circuit/Provable.lean
```diff
@@ -118,9 +124,38 @@ def varFromOffset (α : TypeMap) [ProvableType α] (offset : ℕ) : Var α F :=
...
-export ProvableType (eval const varFromOffset)
+/--
+Notation for element-wise addition of ProvableTypes.
+-/
+infixl:65 " .+ " => ProvableType.elementwiseAdd
+
+/--
+Notation for element-wise scalar multiplication of ProvableTypes.
+-/
+infixl:70 " .* " => ProvableType.elementwiseScalarMul
```
mitschabaude: is there a particular reason against implementing the `Add` and `HMul` type classes, instead of creating custom operators?

(I guess it could be confusing sometimes, e.g. `+` on two U32s not doing carries..)

Clean/Circuit/Provable.lean
```diff
@@ -118,9 +124,38 @@ def varFromOffset (α : TypeMap) [ProvableType α] (offset : ℕ) : Var α F :=
...
+
+export ProvableType (eval const allZero allZeroVar varFromOffset elementwiseAdd elementwiseScalarMul)
```
mitschabaude: instead of exporting `allZero` and `allZeroVar`, I think you could implement the `Zero` type class!

Clean/Circuit/Provable.lean
mitschabaude: actually maybe `SMul` is the right type class for multiplication? Written • and typed `\smul`

Clean/Circuit/Theorems.lean
```diff
@@ -630,3 +631,78 @@ def FormalAssertion.isGeneralFormalCircuit (F : Type) (Input : TypeMap) [Field F
...
   }
+
+-- Theorems about elementwiseAdd and allZero
+
+namespace ProvableType
+
+variable {M : TypeMap} [ProvableType M]
+
+@[circuit_norm]
+theorem allZero_elementwiseAdd (a : M F) : allZero .+ a = a := by
```
mitschabaude: very nice!

after seeing all these theorems I'm pretty sure we should implement `Zero`, `Add` and `SMul` for provable types, because they define a vector space over `F`, which we can show using a handful of lemmas like these, and then we get Mathlibs theory about vector spaces for free

Clean/Gadgets/ElementwiseAdd.lean
mitschabaude, line 1: this defines a circuit that doesn't do any circuit operations. it seems to me that this mainly achieves the rewrite 
```lean
eval env (x_var + y_var) = (eval env x_var) + (eval env y_var)
```
which could just be a simp lemma instead? (assuming we define `Add` for both `M (Expression F)` and `M F`)

Clean/Gadgets/ElementwiseScalarMul.lean
mitschabaude, line 1: same comment as for `ElementwiseAdd`. I'm sceptical of creating circuits without circuit operations

Clean/Gadgets/Conditional.lean
```diff
@@ -0,0 +1,103 @@
...
+  fromComponents := fun (.cons selector (.cons ifTrue (.cons ifFalse .nil))) => { selector, ifTrue, ifFalse }
+
+def main [DecidableEq F] (input : Var (Inputs M) F) : Circuit F (Var M F) := do
+  let { selector, ifTrue, ifFalse } := input
+
+  let scaledTrue ← Gadgets.ElementwiseScalarMul.binaryCircuit { scalar := selector, data := ifTrue }
+  let scaledFalse ← Gadgets.ElementwiseScalarMul.binaryCircuit { scalar := 1 - selector, data := ifFalse }
+
+  -- Add them together (using circuitWithZeroSpec which handles zero plus something)
+  ElementwiseAdd.circuitWithZeroSpec { a := scaledTrue, b := scaledFalse }
```
mitschabaude: [not urgent, can be done later] so this does `s * a + (1 - s) * b`. I'd prefer to replace that with `s * (a - b) + b`, which replaces (`size` multiplications + 1 addition) with (`size` additions), which is more efficient in most proof systems

Clean/Gadgets/Conditional.lean
mitschabaude, line 1: for some reason I'm fine with this being a circuit without circuit operations. feels like the assumptions + specwill make it easier to use like this than as a function

Clean/Gadgets/Conditional.lean
pirapira: will do, I think omega etc will absorb this difference.

Clean/Circuit/Theorems.lean
pirapira: I'll try that. I think there's a way to turn off `+` that I worried about.

Clean/Circuit/Provable.lean
pirapira: `SMul` sounds right. I think there's a way to say it's a vector space.

I'll try to find a way to avoid `+` somehow.

Clean/Gadgets/ElementwiseAdd.lean
pirapira: I'll try inlining this.

Clean/Gadgets/ElementwiseScalarMul.lean
pirapira: I'll try inlining this, too.

Clean/Circuit/Provable.lean
pirapira: `.*` disappeared. Now it's `\smul` and there's a `Module` instance.

The elementwise `+` is hidden under a special namespace, and it's not available unless you open the namespace.

Clean/Circuit/Provable.lean
pirapira: These are `0` now.

Clean/Circuit/Theorems.lean
pirapira: Now there's a `Module` instance. Mathlib 4 chose not to use a special abbreviation for vector spaces (because it caused type inference issues, according to [a comment](https://github.com/leanprover-community/mathlib4/blob/5e4041f47b10a268633b29d7cc3c91a94e984e32/Mathlib/Algebra/Module/Defs.lean#L27)).

Clean/Gadgets/ElementwiseAdd.lean
pirapira: inlined and didn't cause much troubles later.

Clean/Gadgets/Conditional.lean
pirapira: done

Clean/Gadgets/ElementwiseScalarMul.lean
pirapira: the file is gone.

Clean/Circuit/Provable.lean
pirapira: But all these appeared only in intermediate specification, and now these are all gone.

Clean/Gadgets/Conditional.lean
```diff
@@ -0,0 +1,94 @@
+import Clean.Circuit.Provable
+import Clean.Gadgets.Boolean
+import Clean.Utils.Tactics
+
+namespace Conditional
```
mitschabaude: can you make the namespace `Gadgets.Conditional`? We did that throughout clean so far, so that individual gadgets don't mix into the core library namespace.

Clean/Gadgets/Conditional.lean
```diff
@@ -0,0 +1,94 @@
...
+
+def main [DecidableEq F] (input : Var (Inputs M) F) : Circuit F (Var M F) := do
+  let { selector, ifTrue, ifFalse } := input
+
+  -- Inline element-wise scalar multiplication
+  let trueVars := toVars ifTrue
+  let falseVars := toVars ifFalse
+
+  -- Inline element-wise addition
+  let resultVars := Vector.ofFn fun i => selector * (trueVars[i] - falseVars[i]) + falseVars[i]
```
mitschabaude, line 35: so you just got rid of the element-wise add and scalar mul entirely? 😅

mitschabaude (approved):

Clean/Gadgets/Conditional.lean
pirapira: They were on `F` and here I need something for `Expression F`. The thing was only used in the spec of the removed intermediate circuits.

PR #247, opened 2025-08-05
IsZero gadget for ProvableType in general

pirapira:
This will be useful in #229 because `IsZeroU32` can be replaced with the general version in this PR.

### DIFF
--- a/Clean/Circuit/Theorems.lean
+++ b/Clean/Circuit/Theorems.lean
@@ -5,8 +5,28 @@ For more complicated interconnected theorems, we have separate files,
 such as `Circuit.Subcircuit` which focuses on establishing the foundation for subcircuit composition.
 -/
 import Clean.Circuit.Basic
+import Clean.Circuit.Provable
 
 variable {F : Type} [Field F] {α β : Type}
+
+-- ProvableType lemmas
+
+namespace ProvableType
+
+/--
+A useful lemma for proving equality via `fromElements`.
+States that `fromElements A = B` is equivalent to `A = toElements B`.
+This forms one direction of the Galois connection between `Vector F (size M)` and `M F`.
+-/
+theorem fromElements_eq_iff {M : TypeMap} [ProvableType M] {F : Type} {A : Vector F (size M)} {B : M F} :
+    fromElements A = B ↔ A = toElements B := by
+  constructor
+  · intro h
+    rw [← h, toElements_fromElements]
+  · intro h
+    rw [h, fromElements_toElements]
+
+end ProvableType
 
 -- basic simp lemmas
 
--- a/Clean/Circuit/Theorems.lean
+++ b/Clean/Circuit/Theorems.lean
@@ -9,27 +9,6 @@ import Clean.Circuit.Provable
 
 variable {F : Type} [Field F] {α β : Type}
 
--- ProvableType lemmas
-
-namespace ProvableType
-
-/--
-A useful lemma for proving equality via `fromElements`.
-States that `fromElements A = B` is equivalent to `A = toElements B`.
-This forms one direction of the Galois connection between `Vector F (size M)` and `M F`.
--/
-theorem fromElements_eq_iff {M : TypeMap} [ProvableType M] {F : Type} {A : Vector F (size M)} {B : M F} :
-    fromElements A = B ↔ A = toElements B := by
-  constructor
-  · intro h
-    rw [← h, toElements_fromElements]
-  · intro h
-    rw [h, fromElements_toElements]
-
-end ProvableType
-
--- basic simp lemmas
-
 namespace Operations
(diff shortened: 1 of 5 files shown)


Clean/Gadgets/IsZeroField.lean
```diff
@@ -0,0 +1,61 @@
...
+  -- When x ≠ 0, we need x_inv such that x * x_inv = 1
+  -- When x = 0, x_inv can be anything (we use 0)
+  let x_inv ← witness fun env =>
+    if x.eval env = 0 then 0 else (x.eval env : F)⁻¹
+
+  isZero * x === 0  -- If isZero = 1, then x must be 0
+  isZero * (isZero - 1) === 0  -- isZero must be boolean (0 or 1)
+
+  -- If isZero = 0 (i.e., x ≠ 0), then x * x_inv = 1, so x is non-zero
+  (1 - isZero) * x * x_inv === (1 - isZero)
```
mitschabaude: this is less efficient than possible, see `IsZero` in `Circomlib/Comparators.lean` which only uses 2 constraints

Clean/Gadgets/IsZero.lean
```diff
@@ -0,0 +1,105 @@
...
+    · rename_i h
+      simp only [h]
+      simp only [implies_true, true_and, ← h_eval, Vector.getElem_map]
+    · rename_i h
+      simp only [h]
+      simp only [false_and, ↓reduceIte]
+
+theorem soundness : Soundness F (elaborated (M := M)) Assumptions Spec := by
+  circuit_proof_start
+  simp only [Parser.Attr.explicit_provable_type, ProvableType.eval, ProvableType.fromElements_eq_iff] at h_input
```
mitschabaude: nit: you can just pass `explicit_provable_type` outside a namespace, and `eval` shouldn't be necessary as it's part of that simp set

Clean/Gadgets/IsZero.lean
```diff
@@ -0,0 +1,105 @@
...
+      IsZeroField.circuit.Assumptions (Expression.eval (F:=F) env vars[i]) →
+        IsZeroField.circuit.Spec (Expression.eval (F:=F) env vars[i])
+          (Expression.eval (F:=F) env
+            (IsZeroField.circuit.output vars[i]
+              (i₀ + i * IsZeroField.circuit.localLength vars[i])))) :
+    Expression.eval env
+      (Fin.foldl n
+        (fun acc i => acc * (IsZeroField.circuit.output vars[i] (i₀ + i * IsZeroField.circuit.localLength vars[i]) : Var field F))
+        1) =
+    if ∀ (i : Fin n), vals[i] = 0 then 1 else 0 := by
```
mitschabaude, line 56: this seems like it's almost all of the soundness proof. why separate it into a lemma? doing so adds verbosity (you have to state all the hypotheses here)

Clean/Circuit/Theorems.lean
```diff
@@ -5,9 +5,29 @@ For more complicated interconnected theorems, we have separate files,
...
+States that `fromElements A = B` is equivalent to `A = toElements B`.
+This forms one direction of the Galois connection between `Vector F (size M)` and `M F`.
+-/
+theorem fromElements_eq_iff {M : TypeMap} [ProvableType M] {F : Type} {A : Vector F (size M)} {B : M F} :
+    fromElements A = B ↔ A = toElements B := by
+  constructor
+  · intro h
+    rw [← h, toElements_fromElements]
+  · intro h
+    rw [h, fromElements_toElements]
```
mitschabaude: all of the other lemmas about provable types are in `Provable.lean` for now, maybe we can keep it that way for consistency? and create a separate file for those lemmas at some point

Clean/Gadgets/IsZeroField.lean
pirapira: done this and the proofs are shorter too

Clean/Gadgets/IsZero.lean
pirapira: That worked.

Clean/Gadgets/IsZero.lean
pirapira: For mathematical induction. In the original context, `induction n` doesn't work because the goal contains `size M` instead of `n`.

I spend twenty minutes trying to replace `size M` with a fresh variable and replace that everywhere but none of the obvious things worked.

Clean/Circuit/Theorems.lean
pirapira: Moved.

Clean/Gadgets/IsZero.lean
pirapira: I guess it's because `size M` appears in the types, and the whole goal needs to be casted or something.

Issue #250, opened 2025-08-06
[BLAKE3] processChunk gadget


pirapira:
This issue implements and verifies a circuit for `processChunk`, on top of #124 .

- [ ] implement and verify a gadget for `initialChunkState`
- [ ] implement and verify a gadget for `updateChunk`
- [ ] implement and verify a gadget for `finalizeChunk`

PR #253, opened 2025-08-07
Implement Or32 gadget for bitwise OR on U32 values

pirapira:
## Summary
- Implements Or32 gadget for performing bitwise OR operations on U32 values

## Motivation
The Or32 gadget is needed for `BLAKE3's finalizeChunk` implementation to combine flags using bitwise OR (e.g., `base_flags ||| start_flag ||| CHUNK_END`). This was identified as a dependency while working on issue #250.

## Implementation Details

The implementation follows the same pattern as the existing Xor32 gadget:

## Test Plan
- [x] `lake build` succeeds

### DIFF
--- a/Clean.lean
+++ b/Clean.lean
@@ -18,6 +18,8 @@ import Clean.Tables.Fibonacci32Inductive
 import Clean.Tables.KeccakInductive
 import Clean.Gadgets.Bits
 import Clean.Gadgets.Conditional
+import Clean.Gadgets.IsZeroField
+import Clean.Gadgets.IsZero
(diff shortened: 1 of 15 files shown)


pirapira:
I see, in that case, I'll reduce the whole thing into
```
  let xor := 2*or - x - y
  lookup ByteXorTable (x, y, xor)
```
so it's more like a twin of `AND`.

mitschabaude:
great! the proof will be an interesting challenge :D

pirapira:
Superseded by #254.

PR #257, opened 2025-08-15
Improve UX of IsZero gadget

pirapira:
The specification of `IsZero` gadget contained an awkward expression
```
 (∀ i : Fin (size M), (toElements input)[i]
```
that caused [more work](https://github.com/Verified-zkEVM/clean/pull/229/files#r2276250071) downstream.

This PR modifies the Spec to say just
```
input = 0
```

For that purpose this PR also introduces a `Zero` instance for `ProvableType`s.

### DIFF
--- a/Clean/Circuit/Provable.lean
+++ b/Clean/Circuit/Provable.lean
@@ -423,9 +423,7 @@ lemma fromElements_eq_iff' {M : TypeMap} [ProvableType M] {F : Type} {B : Vector
   · intro h
     rw [← h, fromElements_toElements]
 
-
 -- basic simp lemmas
-
 
 @[circuit_norm]
 theorem eval_const {F : Type} [Field F] {α : TypeMap} [ProvableType α] {env : Environment F} {x : α F} :


Clean/Gadgets/IsZero.lean
```diff
@@ -34,8 +34,8 @@ instance elaborated : ElaboratedCircuit F M field where
 
 def Assumptions (_ : M F) : Prop := True
 
-def Spec (input : M F) (output : F) : Prop :=
-  output = if (∀ i : Fin (size M), (toElements input)[i] = 0) then 1 else 0
+def Spec [DecidableEq (M F)] (input : M F) (output : F) : Prop :=
```
mitschabaude, line 37: maybe we could also prove
```lean
instance [DecidableEq F] [ProvableType M] : DecidableEq (M F)
```

pirapira: I think this instance is provable.

The typeclass resolution might become slower (especially in the failing case). This might and might not be an issue, but it's hard to notice and debug. An experiment is time-consuming too.

The merit is unclear. All existing `ProvableType`s are already `DecidableEq` because they are nested structures, vectors or pairs of fields.

I guess it's reasonable to wait and see if we get exotic `ProvableType`s that aren't trivially `DecidableEq`.

mitschabaude: ok got it, so you're saying Lean can infer these instances, without us providing them explicitly?

pirapira: mixed results
* `#check (inferInstance : DecidableEq (ProvablePair field field F))` succeeds
* `#check (inferInstance : DecidableEq (ProvableVector field 10 F))` succeeds
* `#check (inferInstance : DecidableEq (U32 (F p)))` doesn't succeed
  * `deriving DecidableEq` after the `structure U32` definition helps

I'm still afraid of duplicate instances, so I'll take the `deriving DecidableEq` route.

Issue #252, opened 2025-08-07
Lookup tables with multiplicities


pirapira:
There's a frequent pattern involving send/receive of relation tuples. This issue keeps track of supporting that pattern in Clean.

mitschabaude:
related, and to be considered when investigating this: https://github.com/Verified-zkEVM/clean/issues/162

PR #261, opened 2025-08-20
Counting tuples

pirapira:
This is one candidate solution to https://github.com/Verified-zkEVM/clean/issues/252

The implementation is based on @mitschabaude 's idea. A new operation `count` is introduced. At the witness generation time, tuples of field elements (usually encoding a `ProvableType`) will be counted.

At the moment, the accounting is based on a pair of `label, arity` (called `Account`).


pirapira:
#309 supersedes this.

PR #274, opened 2025-09-16
Update clean logo

marcobesier:
- Adds logo files to /images directory.
- Updates logo in Readme.
- Removes old logo from repo.


images/logo-main.png
mitschabaude, line 1: for the README, a version with smaller upper and lower margin would be nice!

README.md
```diff
@@ -9,6 +7,8 @@
 
 </div>
 
+## About clean
```
mitschabaude: agree with removing the centered header. for this intro, we could either remove the header entirely, or change it to just "clean". both seem cleaner to me than "About clean"

marcobesier: yeah, I was thinking about that, too. You're right, let's just get rid of the header. 👍

images/logo-main.png
marcobesier: Agreed. Will push rectangular versions of the logos as well, as soon as we've fixed the misaligned letters. Apparently, that happened during export - in edit mode, the letters were perfectly aligned. Might have to switch to another editor, since this one is giving us a hard time with the export. 😅

PR #275, opened 2025-09-16
Add proof-of-concept formalization of a minimal VM

gio54321:
Add a new example: a formalization of an AIR implementing a VM for the femtoCairo CPU architecture.

femtoCairo is a Cairo-style ISA, and in particular, the memory is read-only. Ideally, we want to do a minimal VM which is similar to risc-v, but we still need to figure out how to formalize read-write memory and memory consistency.

This example still shows the basic structure of a zkVM. For each cycle, the VM does the following things.
- Fetching the instruction at the current program counter.
- Decoding the instruction.
- Performing relevant memory operations in the data memory (in this case, only reads).
- Checking that the relevant arithmetic operations between the values retrieved from the memory are correct.

Notice that, in the case of read-write memory, the structure of the main table remains the same: the memory read lookups are replaced by different lookups for reads and writes.


Clean/Examples/FemtoCairo/Types.lean
```diff
@@ -0,0 +1,189 @@
...
+  Contains the current machine state, the offset operand, and the addressing mode.
+-/
+structure MemoryReadInput (F : Type) where
+  state : State F
+  offset : F
+  mode : DecodedAddressingMode F
+
+/--
+  Input structure for checking the validity of a state transition.
+  Contains the current state, the next state, the decoded instruction, and the values read from memory.
```
pirapira: I see only one `State F` in the implementation. So maybe either `the current state` or `the next state` should be dropped from the comment?

gio54321: Right, thanks! This is a leftover from a refactoring.
Will fix before merging

PR #277, opened 2025-09-18
Generalization of lookup (yielding and using tuples with properties)

pirapira:
Work in progress with some sorry's.

I'm opening this PR already so we can decide **whether to follow through this approach.**

Maybe the complication is not worth it.

Clean/Circuit/PropertyLookup.lean explains the approach.

Clean/Examples/FibonacciYield.lean is the example.

An alternative: the soundness proof doesn't use induction on general ordering on tuples, but the order of witness-generation, where only already `yield`ed tuples can be `use`ed with confidence that the tuple satisfies the nice property. This relieves the user from specifying an ordering and proving well-foundedness.

## Status Sep 26, 2025

The example is finished without sorrys. Currently proving more circuits yield nothing.


pirapira:
> it would be so nice if we could get rid of the new dependent type arguments (PropertySet, SequenceOrder)

These can also default to the empty `PropertySet` and the empty `SequenceOrderer`.

mitschabaude:
another feedback:
it'd be nice if the example weren't a single circuit that hard-codes a bunch of uses/yields, but something more generic and powerful, like implementing the memory lookups in femtoCairo with use/yield

mitschabaude:
do I see this correctly: right now there's this global `checked` set used in the `AllDependenciesChecked` hypothesis, but actually providing `checked` and proving `AllDependenciesChecked` is expected to happen somewhere outside? the fibonacci example would be better if it did that I feel

mitschabaude:
> Whether/how to do mathematical induction would be the user's choice.

fair enough but this PR should have an example of doing so successfully!

pirapira:
> soundness proofs breaks our abstraction of subcircuits

This is intended. There can be mutually recursive inductive reasoning that involve multiple circuits.

For a successful `use` guarantee anything, all of the `yield`s everywhere have to be checked inside or outside of the current subcircuit.

pirapira:
> checked will appear as an argument of FormalCircuit. And then Assumption and Spec will talk about it.

This doesn't work when the same circuit is used in multiple steps in the inductive reasoning. For keeping a circuit's ability to contain arbitrary sub-circuits, we need circuits that take multiple steps in the inductive reasoning.

Say one circuit contains 128 CPU steps, so that we need soundness for various `checked` values like `checked = empty`, `checked = {<timestamp=0, *>}`, `checked={<timestamp=1, *>}`, `checked = {<timestamp=2, *>}` ... .

We have just one circuit with a parametrized soundness theorem. assumption and the spec get one additional parameter.

* `FormalCircuit` gets additional type parameter `ReasoningState : Type`
* `Assumption` takes `r : ReasoningState`
* `Spec` takes `r : ReasoningState`
* `soundness` statemment holds for every `ReasoningState` with `forall r, Assumption r ... -> Spec r ...`.

This is not yet dependent type.

pirapira:
Relevant techniques, we don't need yet, but nice to keep in sight:
* [nominal logic](https://www.cl.cam.ac.uk/~amp12/talks/tacs01.pdf) supports allocating a "fresh" name that is guaranteed not to collide with any other name
* [Iris](https://iris-project.org/) supports associating an access-controlled resource to such a new name

pirapira:
Another try is running in https://github.com/Verified-zkEVM/clean/pull/286

PR #283, opened 2025-09-24
Proved soundness and completeness of circomlib/Sign

matteobusi:
Added soundness and completeness for the `Sign` circuit in `circomlib`.  Solves #235


pirapira (approved):
Looks great. Thank you very much.

PR #291, opened 2025-10-14
Foundational theorems for offline memory checking

gio54321:
The purpose of this PR is to formally define and prove the equivalence between "online" memory checking and "offline" memory checking. Both of those memory checking operations operate on tuples of "memory operations" `(t, a, r, w)`, where `t` is the timestamp of the operation, `a` is the address, and for each memory operation, we read the current value `r` and write a new value `w`. This is a common trick used by zkVMs, as memory reads can be emulated by enforcing that `r = w` in the operation.

There are two different consistency checks:
- **Online memory checking**: for each operation `(t, a, r, w)`, we need to find the most recent operation on the same address `(t', a, r', w')`, with `t' < t`, and we need to enforce that `r = w'`. If this is the first operation done on address `a`, then we ensure that `r = 0`.
- **Offline memory checking** (aka memory-in-the-head): we sort the operations in ascending order first by address and then by timestamp. For the first row `(t, a, r, w)`, it must hold that `r = 0`. Then, we look at all subsequent pairs of consecutive rows `(t1, a1, r1, w1)` and `(t2, a2, r2, w2)`.
  - If `a1 = a2`, then it must hold that `r2 = w1`
  - If `a1 != a2`, them it must hold that `r2 = 0`

The goal of this PR is to define in Lean those two statements and prove the following theorem: given a timestamp-sorted list of memory operations, they are consistent according to online memory checking if and only if there exists a permutation of that list, such that the elements are address-timestamp-sorted, and they are consistent according to offline memory checking.

Notice that the offline memory checking can be directly used as a spec for an inductive AIR table, without any modifications. Furthermore, online memory checking is much closer to the low-level VM spec, for example, of [riscv-sail-lean](https://github.com/opencompl/sail-riscv-lean/blob/main/LeanRV64D/Sail/Sail.lean). The idea is that the main trace spec uses online memory consistency (or something similar), and the memory trace spec uses offline memory consistency. Then, given that we add a permutation argument, we can do a final "connection" of the two specs using these foundational theorems.

### DIFF
--- a/Clean/Utils/OfflineMemory.lean
+++ b/Clean/Utils/OfflineMemory.lean
@@ -225,6 +225,18 @@ theorem MemoryAccessList.noTimestampDup_perm (l1 l2 : MemoryAccessList)
   obtain ⟨t_y, a_y, _r_y, _w_y⟩ := y
   simp_all only [eq_comm, not_false_eq_true]
 
+theorem MemoryAccessList.noTimestampDup_of_cons (head : MemoryAccess) (tail : MemoryAccessList)
+    (h : Notimestampdup (head :: tail)) :
+    Notimestampdup tail := by
+  simp only [Notimestampdup] at *
+  exact List.Pairwise.of_cons h
+
+theorem MemoryAccessList.isAddressTimestampSorted_of_cons (head : MemoryAccess) (tail : MemoryAccessList)
+    (h : isAddressTimestampSorted (head :: tail)) :
+    isAddressTimestampSorted tail := by
+  simp only [isAddressTimestampSorted] at *
+  exact List.Sorted.of_cons h
+
 theorem MemoryAccessList.noTimestampDup_of_TimestampSorted
     (accesses : MemoryAccessList) (h_sorted : accesses.isTimestampSorted) :
     accesses.Notimestampdup := by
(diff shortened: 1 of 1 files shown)


Clean/Utils/OfflineMemory.lean
```diff
@@ -0,0 +1,613 @@
...
+      simp only [is_consistent_online_full.left]
+      rw [eq_comm]
+      apply MemoryAccessList.lastWriteValue_filter
+
+/--
+  Offline version of memory consistency checking.
+-/
+def MemoryAccessList.isConsistentOffline (accesses : MemoryAccessList) (h_sorted : accesses.isAddressTimestampSorted) : Prop := match accesses with
+  | [] => True -- no memory access is trivially consistent
+  | (_timestamp, _addr, readValue, _writeValue) :: [] => readValue = 0
```
pirapira, line 571: This line is actually looking at the last access, usually not the first. So I have a feeling something else is wanted.

pirapira: never mind, I wasn't aware the sort is reversed.

pirapira (approved):
Looks good to me.

PR #292, opened 2025-10-16
chore: Remove consecutive empty lines

pirapira:
`doc/conventions.md` says one empty line is enough so this PR removes consecutive empty lines.


mitschabaude:
would be so nice to have an auto-formatter for stuff like that

Issue #209, opened 2025-07-14
[circomlib] Formalize mux3.circom


pirapira:
Formalize the Mux3 (3-bit multiplexer) circuit from circomlib.

## Implementation details
- Located in circomlib/circuits/mux3.circom
- Natural progression from Mux1 and Mux2
- Takes 8 inputs per output (2^3) and a 3-bit selector
- Implements MultiMux3 template that handles n parallel 3-bit multiplexers

## References
- Parent epic: #155 (Formalize all of circomlib)
- Related: #197 (Mux1 properties), #200 (Formalize mux1.circom), #208 (Formalize mux2.circom)
- [Original circomlib source](https://github.com/iden3/circomlib/blob/master/circuits/mux3.circom)

Issue #208, opened 2025-07-14
[circomlib] Formalize mux2.circom


pirapira:
Formalize the Mux2 (2-bit multiplexer) circuit from circomlib.

## Implementation details
- Located in circomlib/circuits/mux2.circom
- Natural progression from Mux1 (which is already formalized)
- Takes 4 inputs per output (2^2) and a 2-bit selector
- Implements MultiMux2 template that handles n parallel 2-bit multiplexers

## References
- Parent epic: #155 (Formalize all of circomlib)
- Related: #197 (Mux1 properties), #200 (Formalize mux1.circom)
- [Original circomlib source](https://github.com/iden3/circomlib/blob/master/circuits/mux2.circom)

PR #297, opened 2025-10-22
Formalized mux2.circom and mux3.circom

BrunoCaste:
Implementation and soundness/completeness proofs for MultiMux2, MultiMux3, Mux2 and Mux3. Solves #208 and #209.


mitschabaude:
if it's ready, I can merge @BrunoCaste

Issue #207, opened 2025-07-14
Formalize RotR (rotate right) circuit


pirapira:
Formalize the RotR (rotate right) circuit from circomlib SHA256 components.

## Implementation details
- Located in circomlib/circuits/sha256/rotate.circom
- Very simple circuit (29 lines total)
- Takes n-bit input and rotates it right by r positions
- out[i] = in[(i+r) % n]

## References
- Parent epic: #155 (Formalize all of circomlib)
- [Original circomlib source](https://github.com/iden3/circomlib/blob/master/circuits/sha256/rotate.circom)

PR #298, opened 2025-10-24
Circomlib RotR implementation

finiteprods:
implements circomlib's sha256/RotR template
* refer to #207 
* todo: soundness & completeness proofs


pirapira (approved):
Looks good to me.

(I was checking the direction of the rotation, looks also good.)

PR #312, opened 2025-12-16
Soundness Proof for `Bits2Num_strict`

Koukyosyumei:


Clean/Circomlib/Bitify2.lean
```diff
@@ -121,6 +121,20 @@ def main (input : Vector (Expression (F p)) 254) := do
...
+
+lemma fieldFromBits_eq_mapFinRange_cast {n} {f : Fin n → F p} :
+    fieldFromBits (Vector.mapFinRange n f) = (fromBits (Vector.mapFinRange n fun i => (f i).val) : F p) := by
+  unfold fieldFromBits
+  apply congrArg (Nat.cast : ℕ → F p)
+  apply congrArg fromBits
+  ext i
+  simp only [Vector.getElem_map, Vector.getElem_mapFinRange]
+
 def circuit : FormalCircuit (F p) (fields 254) field where
```
pirapira, line 140: The CI is failing around here. For me locally, adding this option here fixes the issue. I don't fully understand what's going on so I created [an issue](https://github.com/Verified-zkEVM/clean/issues/313) for investigating this. In any case it's about a syntax linter.

```suggestion
set_option linter.constructorNameAsVariable false

def circuit : FormalCircuit (F p) (fields 254) field where
```

Clean/Circomlib/Bitify2.lean
```diff
@@ -135,8 +149,25 @@ def circuit : FormalCircuit (F p) (fields 254) field where
...
+    rw[← mapFinRange_eq_map]
+    rw[← fieldFromBits_eq_mapFinRange_cast]
+    conv =>
+          rhs
+          congr
+          congr
+          congr
+          ext i
+          rw [← h_input]
+          simp
```
pirapira: When `simp` doesn't close a goal completely, we prefer `simp only [...]`. This makes the script robust against `simp` annotation changes.
```suggestion
          simp only [Fin.getElem_fin, Vector.getElem_map]
```

PR #314, opened 2025-12-19
Soundness Proof for `Num2BitsNeg`

Koukyosyumei:
This PR completes the soundness proof for the `Num2BitsNeg` circuit in Clean/Circomlib/Bitify2.lean.

*Proof sketch*

1. Extract bit constraints and the zero-test invariant from h_holds.
2. Split on n = 0 and discharge the trivial case by simplification.
3. For n > 0:
   - Define the evaluated bit vector from the environment.
   - Show the folded linear combination equals fieldFromBits.
   - Split on whether input = 0:
      - If input = 0, show the output bits encode 0.
      - Otherwise, show the output bits encode 2^n - input.


pirapira (approved):
I applied some cosmetic changes, but the essence is unchanged. Thank you very much.

PR #317, opened 2025-12-19
FemtoCairo refinements

### LINKED ISSUE #311: femto: readFromMemoryCircuit to perform reads only on the actual mode
Currently, `readFromMemoryCircuit` performs four reads corresponding to four modes and then creates a linear combination of the four read values. This requires luck because all four reads need to be located within the memory.

This issue keeps track of modifying `readFromMemoryCircuit` so that the irrelevant modes don't add requirements about addresses.

mitschabaude:
* simplify some FemtoCairo proofs
* remove unrealistic assumptions for completeness; namely, that multiple unrelated memory accesses all have to succeed, since the circuit uses lookups for all addressing modes. in the new version, the circuit selects the correct address before the lookup, so that only lookups related to the runtime value of the addressing modes are performed

closes #311

### DIFF
--- a/Clean/Circomlib/Bitify2.lean
+++ b/Clean/Circomlib/Bitify2.lean
@@ -127,6 +127,7 @@ lemma mapFinRange_eq_map {α β : Type} {n : ℕ} (v : Vector α n) (f : α → 
   simp only [Vector.getElem_mapFinRange, Vector.getElem_map]
   simp
 
+omit [Fact (p < 2 ^ 254)] [Fact (p > 2 ^ 253)] in
 lemma fieldFromBits_eq_mapFinRange_cast {n} {f : Fin n → F p} :
     fieldFromBits (Vector.mapFinRange n f) = (fromBits (Vector.mapFinRange n fun i => (f i).val) : F p) := by
   unfold fieldFromBits
--- a/Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
+++ b/Clean/Tables/BLAKE3/ProcessBlocksInductive.lean
@@ -278,7 +278,6 @@ private lemma step_process_block (env : Environment (F p))
     simp [acc_normalized, circuit_norm])
   dsimp only [Addition32.Spec] at h_addition ⊢
   rcases h_holds with ⟨ h_vector_cond, h_u32_cond ⟩
-  dsimp only [Conditional.Spec] at h_vector_cond h_u32_cond
   specialize h_vector_cond (by simp only [circuit_norm])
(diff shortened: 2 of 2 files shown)


Clean/Examples/FemtoCairo/FemtoCairo.lean
```diff
@@ -1024,117 +755,51 @@ def femtoCairoStepCircuit
...
+  step state _ :=
+    femtoCairoStep program h_programSize memory h_memorySize state
+
+  Spec initialState _ i _ state : Prop :=
+    (Spec.femtoCairoMachineBoundedExecution program memory (some initialState) i) = some state
+
+  -- For completeness, we assume that execution on the initial state succeeds, for all steps up to a maximum N
+  InputAssumptions i _ := i < n
+  InitialStateAssumptions initialState :=
+    ∀ i ≤ n, (Spec.femtoCairoMachineBoundedExecution program memory (some initialState) i).isSome
```
mitschabaude, line 769: the previous assumptions stated that execution succeeds for _all_ initial states, which is much stronger than what is needed: that execution succeeds for the _given_ initial state

Clean/Examples/FemtoCairo/FemtoCairo.lean
```diff
@@ -401,44 +251,63 @@ def fetchInstructionCircuit
...
+      mode.isFpRelative * (state.fp + offset)
+
+    let value1 ← witness fun eval => memory <| Fin.ofNat _ (eval addr1).val
+
+    let addr2 <== mode.isDoubleAddressing * value1
+
+    let value2 ← witness fun eval => memory <| Fin.ofNat _ (eval addr2).val
+
+    lookup memoryTable ⟨addr1, value1⟩
+    lookup memoryTable ⟨addr2, value2⟩
```
mitschabaude, line 285: I refined the circuit here to only do the lookups that are actually relevant, which allows us to get rid of unrealistic memory assumptions

Clean/Gadgets/Conditional.lean
mitschabaude, line 1: I actually wanted to use this conditional gadget in the femto cairo circuit, but the way it's currently written makes that pretty annoying (it puts the entire computation into a complicated output expression), so I abandoned that attempt

pirapira: I see complicated output expressions in many other places too. One possible way to avoid this is to put a new variable in the context and say "the new variable satisfies the spec of the sub-circuit". #322

pirapira:
Merging so that I can start updating other PRs.

PR #320, opened 2025-12-23
Soundness and completeness for BinSub

matteobusi:
This PR provides soundness and completeness proofs for the BinSub circuit, along with some minor fixes and proofs. Also, it adds a number of lemmas to `Clean/Circuit/Expression.lean`, `Clean/Utils/Fin.lean`, and `Clean/Utils/Bits.lean` that we believe may be generally useful.

These proofs also constitute a first attempt at (partially) automating circuit proofs using AI. Loosely following [this paper](https://arxiv.org/abs/2504.21801), a general-purpose LLM was used to decompose soundness and completeness statements into intermediate goals, which were then handed to a specialized AI for theorem proving in Lean ([Aristotle by Harmonic](https://aristotle.harmonic.fun/)). The resulting proofs were then optimized, simplified, and partially re-engineered by hand.


pirapira (approved):
Looks good to me.

PR #324, opened 2026-01-12
Preserve nesting in subcircuit operations

mitschabaude:
This solves the issue that we couldn't map clean circuits to plonk custom gates, because we didn't preserve the subcircuit nesting structure inside subcircuits itself.
the solution is surprisingly easy

### DIFF
--- a/Clean/Circuit/Operations.lean
+++ b/Clean/Circuit/Operations.lean
@@ -23,8 +23,7 @@ inductive NestedOperations (F : Type) where
 
 def NestedOperations.toFlat {F : Type} : NestedOperations F → List (FlatOperation F)
   | .single op => [op]
-  | .nested (_, []) => []
-  | .nested (s, op :: ops) => toFlat op ++ toFlat (.nested (s, ops))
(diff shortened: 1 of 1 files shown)


Clean/Circuit/Operations.lean
```diff
@@ -17,6 +17,15 @@ inductive FlatOperation (F : Type) where
...
   | lookup : Lookup F → FlatOperation F
 
+inductive NestedOperations (F : Type) where
+  | single : FlatOperation F → NestedOperations F
+  | nested : String × List (NestedOperations F) → NestedOperations F
+
+def NestedOperations.toFlat {F : Type} : NestedOperations F → List (FlatOperation F)
+  | .single op => [op]
+  | .nested (_, []) => []
+  | .nested (s, op :: ops) => toFlat op ++ toFlat (.nested (s, ops))
```
pirapira:
```suggestion
  | .nested (_, lst) => List.flatMap toFlat lst
```
This pattern already has a name, so let's use it. This doesn't affect the build.

mitschabaude: nice!

Issue #319, opened 2025-12-22
Enable dynamic lookup tables


mitschabaude:
`FemtoCairo` revealed that clean's lookup definition does not properly support an essential pattern: lookups into tables that are only filled at runtime by the prover.

The problem is that for such dynamic tables, there is no good way to capture the statement

> a lookup implies that the value is part of the table that the prover provided

because `Table.Contains` does not give you access to the runtime table values.

`Table.Contains` is currently only useful for cases where "the element is in the table" has an equivalent formulation that only uses the content of the current row.

In the case of the memory lookups used in `FemtoCairo` however, the lookups assert _nothing_ about the elements other than being in the memory table, which in and of itself is a non-trivial consistency property.

Mitigation:
* pass in the table's actual values at prover runtime to the `Contains` (and `Soundness` and `Completeness`) statement
* enable us to use these values in Specs and Assumptions, supporting statements like "when this circuit's constraints holds, then x is in table T"
* therefore, we need slots to hold these values in FormalCircuit and relatives, and in Spec/Assumptions
* for example, we could widen the definition of `Environment` to not only contain the current circuit's witness values, but also the actual values of lookup tables. This could be realized as a hashmap where you use a table's name as key to get the table's runtime value (an array of vectors)

PR #325, opened 2026-01-19
Environment with tables

mitschabaude:
fixes https://github.com/Verified-zkEVM/clean/issues/319, using the strategy already laid out there

an `Environment` (which basically represents the information provided by the prover when instantiating a circuit with concrete values) now not only carries the flat `\N -> F` map which represents witnesses for the circuit's own variables, but also the contents of an unspecified number of additional tables the prover has access to, keyed by table name.

this is used in femtocairo to replace the hard-coded memory table with a dynamic one that can be picked at proving time and remain unknown to the verifier.

### DIFF
--- a/Clean/Table/WitnessGeneration.lean
+++ b/Clean/Table/WitnessGeneration.lean
@@ -58,7 +58,7 @@ def generateNextRow (tc : TableConstraint W S F Unit) (cur_row : Array F) : Arra
 
   -- evaluate the witness generators
   let (_, next_row) := generators.foldl (fun (idx, next_row) compute =>
-      let wit := compute ⟨ ⟨env⟩, ⟨fun _ _ => #[]⟩ ⟩
+      let wit := compute ⟨ env, fun _ _ => #[] ⟩
 
       -- insert the witness value to the next row
       let next_row := if h : idx < assignment.offset then


mitschabaude:
> I'm wondering whether the same can be achieved with a `ProvableVector` of rows with one special operation for lookup (this will be translated into, for example, channel mechanism).

not following how this would achieve the same

mitschabaude:
> That's new. I don't know a backend implementation that supports changing table lengths.

all AIR backends support this. they usually require the length to be a power of two but support many different powers. so you need both:
* flexibility about the table length
* ability to disable the constraints in a given row

Clean/Circuit/Basic.lean
```diff
@@ -388,8 +388,8 @@ add the range assumption to the soundness statement, thus making the circuit har
 -/
 structure GeneralFormalCircuit (F : Type) (Input Output : TypeMap) [Field F] [ProvableType Input] [ProvableType Output]
     extends elaborated : ElaboratedCircuit F Input Output where
-  Assumptions : Input F → Prop -- the statement to be assumed for completeness
-  Spec : Input F → Output F → Prop -- the statement to be proved for soundness. (Might have to include `Assumptions` on the inputs, as a hypothesis.)
+  Assumptions : Input F → Environment F → Prop -- the statement to be assumed for completeness
+  Spec : Input F → Output F → Environment F → Prop -- the statement to be proved for soundness. (Might have to include `Assumptions` on the inputs, as a hypothesis.)
```
pirapira: I think it's better to pass just `TableEnvironment` instead of `Environment`.

Wasn't there a software engineering folklore that every interface gets misused.

pirapira:
Usually we first fill the looking-up tables, gather requirements about what entries are needed, and then fill the looked-up tables.

So, instead of adding tables to the environment, we can modify the completeness statement so that the prover can output a list of required lookup entries.

The prover will gather all the required lookup entries, fill in more tables, get more required lookup entries, etc until the queue of required lookup entries become empty.

Clean/Circuit/Lookup.lean
```diff
@@ -20,17 +20,19 @@ structure Table (F : Type) (Row : TypeMap) [ProvableType Row] where
   /--
   `Contains` captures what it means to be in the table.
   -/
-  Contains : Row F → Prop
+  Contains : Array (Row F) → Row F → Prop
```
pirapira: This needs some descriptions.

Can `Contains rows row` be read like "`rows` contains `row`"?

Clean/Circuit/Lookup.lean
```diff
@@ -20,17 +20,19 @@ structure Table (F : Type) (Row : TypeMap) [ProvableType Row] where
   /--
```
pirapira, line 20: Eight lines above I see `Contains _ := False` in a comment.The comment needs to be updated.

Clean/Table/Basic.lean
```diff
@@ -484,15 +485,25 @@ instance [Repr F] : Repr (TableOperation S F) where
 
 export TableOperation (boundary everyRow everyRowExceptLast)
 
+structure TableEnvironments (F : Type) where
+  /-- environment for each constraint, for each row -/
+  witnessEnvs : ℕ → ℕ → WitnessEnvironment F
```
pirapira: That N's are confusing. Do they mean `witnessEnvs (constraint : N) (row : N) : ...` ?

Clean/Table/Basic.lean
```diff
@@ -484,15 +485,25 @@ instance [Repr F] : Repr (TableOperation S F) where
...
+  /-- stable environment representing other tables -/
+  otherTables : TableEnvironment F
```
pirapira: The comment says stable environment. I'm going to pass more arguments that represent the contents of the other tables so that I can use the `otherTables` with unstable contents, but that's fine.

Clean/Table/Basic.lean
mitschabaude: good call!

pirapira:
> I don't want to refactor Table and Lookup as my next task is removing them 

Maybe it's not much worth reading them either. Probably we don't have to see the mid-refactoring state on `main` (nobody else is making big changes now). Can you continue doing this here?

mitschabaude:
> Can you continue doing this here?

Why? This represents a discrete change unrelated to the move toward channels. It solves a problem with code on main (the fact that FemtoCairo has to hard-code the whole memory table, and therefore makes a statement that is unlike what a real zkVM should do).

On the other hand, @pirapira your comment about `Table` being not defined optimally is something that was here before the PR and not related to the problem it addresses, so it shouldn't be mixed together with this work.

mitschabaude:
> Please remove it or modify it (I suggested `Contents : Array (Row F)`.

as I've also argued in DM, I'm strongly against requiring to put the actual table contents (or a placeholder for them) in the `Table` structure, as this misunderstands how the circuit is supposed to be an abstract definition of some polynomials without access to their actual values.

Also note how the very problem that this PR was trying to solve is _removing_ hard-coded table values from the FemtoCairo definitions, by adding slots that get these values at runtime (= the env.data and the `Contains` statement)

pirapira:
> this is a statement about how the table is structured, and without it, the proofs don't work.

I disagree. There can be a strange `table` elements whose addresses are wrong. This predicate `Contains` just ignores these entries, and it doesn't enforce a condition on the structure of the `table`.

mitschabaude:
> I'm not getting it. Before this PR, a `Table` knew the content, as a characterization predicate `Contains : Row F -> Prop`.

yeah, I thought that would be enough info before FemtoCairo. but it isn't. a lookup constraint is really establishing two pieces of information:

* that the looked up entry has the properties that all table entries have (= the characterization)
* that the looked up entry is in the table!

the second one is non-trivial because it relates different looked up values to each other, in a way that a pure characterization of a single row can't. and for the memory argument we needed that

Clean/Circuit/Lookup.lean
```diff
@@ -19,18 +19,21 @@ structure Table (F : Type) (Row : TypeMap) [ProvableType Row] where
   name : String
   /--
   `Contains` captures what it means to be in the table.
+  Read it like "`table` contains `row`".
   -/
-  Contains : Row F → Prop
+  Contains : (table: Array (Row F)) → (row: Row F) → Prop
```
pirapira, line 24: I think we should allow the two `Row F`s to be different types.

In case of logups, the committed values and the looked up values can be different. When `table` contains `(x,y,z)` the `Contains` might accept `(x)` as a row, etc. This is useful for using an 16-bit xor table as a 16-bit range-check.

pirapira (approved):
This is confusing but there's an interpretation that `Table` now captures whatever the lookup constraints impose, which might not be just a membership checking.

There will be more changes on the confusing parts so I think it's good to merge.

PR #309, opened 2025-11-27
Add multiplicity-based multiset operations for circuit execution tracking

pirapira:
## Summary

This PR implements a multiplicity-based approach for proving cairo-like vm execution validity, centered around a graph-theoretic flow-balance theorem.

### Key tool: SourceSinkPath Theorem

The key tool is `SourceSinkPath.exists_path_from_source_to_sink` in `Clean/Utils/SourceSinkPath.lean`:

> **Theorem**: Given a finite directed graph where the net flow is +1 at source `s`, -1 at sink `d`, and 0 at all other vertices, there exists a valid cycle-free path from `s` to `d`.

This theorem enables proving execution validity without explicit step indices or timestamps.

### Application: PicoCairoMultiplicity

The theorem is applied to VM execution tracking:
- Each instruction emits `single(preState, -1) + single(postState, +1)` (edge contributions)
- Initial state +1 and final state -1 (boundary conditions)
- When the multiset balances to zero, the flow-balance theorem guarantees a valid execution path exists

The important statement is [ExecutionExistenceSpec](https://github.com/Verified-zkEVM/clean/pull/309/files#diff-2d75e2fbef3a7cd3116d669ea6fea480032550ae87b3fcb62441afb5f075f424R1975). If constraints are satisfied and the multiset balances out, there exists a `femto` execution.

### Scope

- **Included**: Full soundness proofs for all instruction circuits, bundles, and the execution theorem
- **Out of scope**: Completeness proofs (left as `sorry`)

## Test plan

- [x] `lake build` passes with no errors

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #328, opened 2026-01-23
Channels

mitschabaude:
this PR is an attempt to
* introduce `Channels` as an abstraction similar to `Tables`, that model _all_ cross-table interactions
* figure out a good e2e UX for using channels in various ways, and in particular for multi-table ensembles
* enable proving soundness and completeness at the multi-table level, in a modular way that makes use of local constraints to establish local channel "guarantees"

started on top of #309 but heavily refactoring its contributions

The working example for Channels is [8-bit Fibonacci](https://github.com/Verified-zkEVM/clean/blob/multiset-with-env-tables/Clean/Examples/FibonacciWithChannels.lean) which we implement using 3 different channels that represent 3 prominent ways channels are used:
* for static lookup tables
* for looking up into an "ALU chip", to outsource constraints from one table to another
* for state transitions within one table (here: fibonacci), replacing the classic "next-row" constraint

TODO:

* [x] finish (almost-automatic) soundness proof framwork that works in Fib example
* [x] merge main and fix conflicts
* [x] add separate Assumptions to Subcircuit definition to reclaim FormalCircuit flexibility
* [x] spike: remove properties on Subcircuit that auto-simply to their intended values by rewriting, on particular formal circuits
* [x] make FormalCircuitWithInteractions the default FormalCircuit
  * make sure non-interacting circuits are unaffected (extra props are auto-proved)
  * get rid of sorries in subcircuit composition
* [x] simplify channel interface to not separate assumptions and requirements
* ~~FibonacciWithChannels: get rid of empty verifier assumption in sound channels~~ (will be follow-up PR)
* [x] clean-up pass FibonacciWithChannels - remove unused stuff, improve organization
* [x] move defs and lemmas to their proper locations
* ~~(optional) test framework by replacing one of the existing lookups (e.g. xor) with a channel~~  (will be follow-up PR)
* ~~(optional) test framework on more interesting example like PicoCairo~~  (will be follow-up PR)


pirapira (approved):
Looks good to me (I think most files are changed to accommodate Var changes). I saw everything I expected.

pirapira:
Merged because big

PR #336, opened 2026-02-23
Replace custom LogUp with Plonky3 built-in LogUpGadget

pirapira:
## Summary

- Replace the hand-rolled lookup protocol and custom multi-AIR prover/verifier with upstream Plonky3 crates (`p3-lookup`, `p3-batch-stark`)
- Use `p3-air`'s `Lookup` type with `Kind::Global` / `Direction::Send` / `Direction::Receive` semantics, constructed via `Air::register_lookup()`
- Delegate prover/verifier to `prove_batch` / `verify_batch` from `p3-batch-stark`, using `ProverData::from_airs_and_degrees()` for preprocessing
- Net reduction of ~930 lines (865 added, 1798 removed across 15 files) by leveraging upstream crates instead of custom code
- **`tests/clean_air.rs`**: 4 new integration tests (multi-column, expression, range-check-16, two-table lookups)

### Out of scope

Even after this change, looked-up tables are all preprocessed. The next PR will introduce looking up from the main trace.

### DIFF
--- a/backends/plonky3/src/check_constraints.rs
+++ b/backends/plonky3/src/check_constraints.rs
@@ -19,6 +19,7 @@ use tracing::instrument;
 /// - `air`: The AIR logic to run
 /// - `main`: The trace matrix (rows of witness values)
 /// - `public_values`: Public values provided to the builder
+#[allow(dead_code)]
 #[instrument(name = "check constraints", skip_all)]
 pub(crate) fn check_constraints<F, A>(
     air: &A,


backends/plonky3/src/clean_air.rs
```diff
@@ -149,26 +158,84 @@ impl<F: Field> MainAir<F> {
...
+        let ops_with_assignments = self.clean_ops.lookup_ops_with_assignments();
+
+        // Deduplicate by (table, entry debug repr) and group by table
+        let mut seen: BTreeSet<String> = BTreeSet::new();
+        let mut lookups_by_table: BTreeMap<String, Vec<LookupInput<F>>> = BTreeMap::new();
+
+        for (lookup_op, assignment) in &ops_with_assignments {
+            let key = alloc::format!("{}:{:?}", lookup_op.table, lookup_op.entry);
+            if !seen.insert(key) {
+                continue;
```
pirapira: I prefer just a warning here. There's a possibility that different lookups ending up with the same debug string.

pirapira: We don't do anything special if the same constraint appears twice. So we should remove the `seen` variable and this case.

PR #342, opened 2026-03-18
feat: add CompletenessSpec to GeneralFormalCircuit

pirapira:
Add a CompletenessSpec field to GeneralFormalCircuit that is available during completeness proofs (via UsesLocalWitnesses) but not used during soundness. This allows subcircuit parents to reason about how the output is computed using ProverData, eliminating the need to inline subcircuit logic in completeness proofs.

This is useful when 
(1) a circuit uses outputs from sub-circuits to construct a bigger output using `ProverData`
(2) the outputs from sub-circuits need to satisfy an assertion
(3) the nice property comes from how the output is computed using `ProverData`.
(4) we don't want to inline sub-circuits.

Before this PR, although `Spec` had access to `ProverData`, `Spec` was not useful for describing any nontrivial relationship between the `Output` and `ProverData`. The soundness proof doesn't have any access to witness computation so the soundness could not prove any nontrivial relationship between `Output` and `ProverData`.

After this PR, `CompletenessSpec` can state nontrivial relationship between `Output` and `ProverDaata`. Moreover, `CompletenessSpec` only gets proven in the context of completeness where the specific witness computation is available. This allows subcircuit to inform the supercircuit about the specific choice of `Output` based on `ProverData`.

### DIFF
--- a/Clean/Circuit/Basic.lean
+++ b/Clean/Circuit/Basic.lean
@@ -373,6 +373,17 @@ def GeneralFormalCircuit.Completeness (F : Type) [Field F] (circuit : Elaborated
   -- the constraints hold
   ConstraintsHold.Completeness env (circuit.main input_var |>.operations offset)
 
+@[circuit_norm]
+def GeneralFormalCircuit.CompletenessSpecProof (F : Type) [Field F]
+    (circuit : ElaboratedCircuit F Input Output)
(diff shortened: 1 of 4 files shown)


mitschabaude (approved):
Good to have this for when needed. 

Could there be better names than CompletenessSpec and CompletenessSpecProof? Esp. the latter

PR #351, opened 2026-04-01
feat: add generic IsEqual gadget for any ProvableType (closes #126)

### LINKED ISSUE #126: Generic `IsEqual` gadget
One of the most important core circuits is boolean equality, which returns a field element which is 1 when the two inputs are equal, 0 otherwise.

This should be done in two versions:
* Single-field version that expects two `Expression F`. Uses two quadratic constraints.
* Generic version that works on any `ProvableType`.

Bonus points if the generic version doesn't naively use the single-field version n times, but saves constraints in R1CS-type scenarios.

Bonus points for introducing a `BEq` instance so that we can just use the `==` symbol to compare two variables.

MavenRain:


Clean/Gadgets/IsEqual.lean
```diff
@@ -0,0 +1,89 @@
...
+Compute element-wise differences between two values of a ProvableType.
+Defined separately so it survives circuit normalization as a single vector.
+-/
+def diffs (x y : Var α F) : Vector (Expression F) (size α) :=
+  Vector.mapFinRange (size α) fun i => (toVars x)[↑i] - (toVars y)[↑i]
+
+/--
+Circuit that checks if two values of a ProvableType are equal.
+Returns 1 if the inputs are equal, otherwise returns 0.
+Uses two constraints per field element (via IsZeroField on each component difference).
```
mitschabaude: that's not true, it uses at least 3n constraints since in almost any proof system, a multiplication will translate to a constraint


```suggestion
Uses three constraints per field element (IsZeroField on each component difference + one multiplication).
```

mitschabaude (changes requested):
it's nice to have this but IMO it should reuse `IsZero`! (i.e. just pass the diffs to IsZero)

Clean/Gadgets/IsEqual.lean
```diff
@@ -0,0 +1,86 @@
...
+
+/--
+Circuit that checks if two values of a ProvableType are equal.
+Returns 1 if the inputs are equal, otherwise returns 0.
+Uses three constraints per field element (IsZeroField on each component difference + one multiplication).
+-/
+def main (input : Var α F × Var α F) : Circuit F (Var field F) := do
+  let (x, y) := input
+  let d := diffs x y
+  IsZero.main (fromVars d)
```
mitschabaude: This is wrong, it repeats the constraints of IsZero inline instead of using it as a subcircuit. That forces you to reprove what IsZero already proved, instead of using its spec

mitschabaude:
```suggestion
  IsZero.circuit (fromVars d)
```

pirapira:
Meanwhile the Lean version changed on main. Merging `main` will make the CI accurate.

mitschabaude (approved):

mitschabaude:
thanks @MavenRain, this is perfect!

PR #352, opened 2026-04-02
feat: add FiniteField typeclass abstracting prime/binary field properties

MavenRain:
Toward (#154)

### DIFF
--- a/Clean/Utils/FiniteField.lean
+++ b/Clean/Utils/FiniteField.lean
@@ -56,6 +56,7 @@ instance {p : ℕ} [Fact p.Prime] : FiniteField (F p) where
   val := ZMod.val
   val_lt := by
     intro x
+    unfold F
     rw [ZMod.card]
     exact ZMod.val_lt x
   val_injective := by


Clean/Utils/FiniteField.lean
```diff
@@ -0,0 +1,70 @@
...
+class FiniteField (F : Type) : Type where
+  [toField : Field F]
+  [toFintype : Fintype F]
+  [toDecidableEq : DecidableEq F]
+  /-- Canonical embedding of field elements into natural numbers. -/
+  val : F → ℕ
+  /-- The number of elements in the field. -/
+  fieldSize : ℕ
+  /-- The field has at least two elements. -/
+  fieldSize_pos : fieldSize > 1
```
mitschabaude: you can remove this from the assumptions, it follows from `val_lt` and `val_one`

Clean/Utils/FiniteField.lean
```diff
@@ -0,0 +1,70 @@
...
+/--
+`FiniteField` extends `Field` with a canonical embedding into `ℕ` and finiteness.
+This captures the shared structure of prime fields and binary fields that circuit
+gadgets rely on: the ability to decompose field elements into bits and reason
+about their numeric value.
+-/
+class FiniteField (F : Type) : Type where
+  [toField : Field F]
+  [toFintype : Fintype F]
+  [toDecidableEq : DecidableEq F]
```
mitschabaude: isn't there a `DecidableEq` instance automatically from `Fintype`?

Clean/Utils/FiniteField.lean
```diff
@@ -0,0 +1,70 @@
...
+  /-- Canonical embedding of field elements into natural numbers. -/
+  val : F → ℕ
+  /-- The number of elements in the field. -/
+  fieldSize : ℕ
```
mitschabaude: mathlib already has a concept for the number of elements of a finite type: `Fintype.card`. maybe we should use that, or at least require that `fieldSize = Fintype.card`

Clean/Utils/FiniteField.lean
```diff
@@ -0,0 +1,70 @@
...
+
+/--
+`FiniteField` extends `Field` with a canonical embedding into `ℕ` and finiteness.
+This captures the shared structure of prime fields and binary fields that circuit
+gadgets rely on: the ability to decompose field elements into bits and reason
+about their numeric value.
+-/
+class FiniteField (F : Type) : Type where
+  [toField : Field F]
+  [toFintype : Fintype F]
```
mitschabaude: isn't this the more typical way to do this:
```lean
class FiniteField (F : Type) extends Field F, Fintype F where
```

mitschabaude (approved):

PR #360, opened 2026-04-21
Pass full prover environment where useful

mitschabaude:
Instead of prover data + prover hint separately, this just passes the full `ProverEnvironment` to a circuit's completeness-related statements (`Assumptions`, `CompletenessSpec`).

This increases the flexibility of the general formal circuits API: in particular, it enables passing callbacks of the form
```lean
ProverEnvironment F -> \alpha
```
to general formal circuits, and allow `Assumptions` and `CompletenessSpec` to access its return value.


mitschabaude:
not needed anymore for ergonomic hints, thanks to #363

PR #363, opened 2026-04-22
Prover Hints as Callbacks

mitschabaude:
Make prover hints a first-class circuit input. This unlocks a pattern used in some production circuit libraries that, so far, clean couldn't represent in an idiomatic way.

```lean
/--
  A circuit that witnesses a boolean field element using a prover hint.

  The hint callback tells the prover which boolean value to witness.
  The circuit constrains the output to be boolean (0 or 1).
-/
def witnessBool : GeneralFormalCircuit.WithHint (F p) (Unconstrained Bool) field where
  main (hint : ProverEnvironment (F p) → Bool) := do
    let b ← witness fun env => if hint env then 1 else 0
    assertBool b
    return b

  Assumptions (_ : Unit) _ := True
  Spec (_ : Unit) (output : F p) _ := IsBool output

  ProverAssumptions (hint : Bool) _ _ := True
  ProverSpec (hint : Bool) (b : F p) _ := b = if hint then 1 else 0

structure Input (F : Type) where
  x : F
  y : F
deriving ProvableStruct

/--
  A circuit that computes the AND of two boolean inputs.

  This is a plain `FormalCircuit` (no hint input). It creates the hint
  internally from its inputs and passes it to `witnessBool`.
-/
def booleanAnd : FormalCircuit (F p) Input field where
  main | ⟨x, y⟩ => do
    -- Use witnessBool as a subcircuit with a hint synthesized from the inputs
    let z ← witnessBool fun env => eval env x = 1 ∧ eval env y = 1
    -- Constrain result = x * y (multiplication is AND for booleans)
    z === x * y
    return z

  Assumptions | ⟨x, y⟩ => IsBool x ∧ IsBool y
  Spec | ⟨x, y⟩, z => IsBool z ∧ z.val = x.val &&& y.val
```

### DIFF
--- a/Clean/Circuit/CircuitType.lean
+++ b/Clean/Circuit/CircuitType.lean
@@ -1,10 +1,13 @@
 import Clean.Circuit.Provable
 
+structure Unconstrained (Hint : Type) where
+  value : Hint
+
 /-
   Prover hints: additional data that can be passed to a circuit's witness generation
   but does not affect the circuit's constraints or spec.
 -/
-def ProverHint (Hint : Type) (F : Type) := Environment F → Hint
+def ProverHint (Hint : Type) (F : Type) := ProverEnvironment F → Hint
 
 /--
 `CircuitType` is a joint generalization of `ProvableType` and `ProverHint`.
@@ -18,23 +21,35 @@ class CircuitType (F : Type) (Var : Type) (Value : outParam Type) where
 class ProverType (F : Type) (Var : Type) (Value : outParam Type) where
   evalProver [Field F] : ProverEnvironment F → Var → Value
 
-class Eval (Env : Type) (Var : Type) (Value : outParam Type) where
+class Eval (Env : Type) where
+  Var : Type
+  Value : Type
   eval : Env → Var → Value
 
 export CircuitType (evalVerifier)
 export ProverType (evalProver)
 export Eval (eval)
 
+variable {F : Type} [Field F] {M : TypeMap} [ProvableType M] {Hint : Type}
 variable {Variable : Type} {Value : Type}
-variable {F : Type} {M : TypeMap} [ProvableType M] {Hint : Type}
 
-instance Eval.verifier [Field F] [CircuitType F Variable Value] :
-    Eval (Environment F) Variable Value where
-  eval env v := evalVerifier env v
+instance Eval.verifier [CircuitType F Variable Value] : Eval (Environment F) where
+  Var := Variable
+  Value := Value
+  eval := evalVerifier
 
-instance Eval.prover [Field F] [ProverType F Variable Value] :
-    Eval (ProverEnvironment F) Variable Value where
-  eval env v := evalProver env v
+instance Eval.prover [ProverType F Variable Value] : Eval (ProverEnvironment F) where
+  Var := Variable
+  Value := Value
+  eval := evalProver
+
+@[circuit_norm] lemma eval_eq_evalVerifier [CircuitType F Variable Value]
+ {env : Environment F} (var : Variable) :
+  eval env var = evalVerifier env var := rfl
+
+@[circuit_norm] lemma eval_eq_evalProver [ProverType F Variable Value]
+ {env : ProverEnvironment F} (var : Variable) :
+  eval env var = evalProver env var := rfl
 
 instance {F : Type} : CircuitType F (ProverHint Hint F) Unit where
   evalVerifier _ _ := ()
--- a/Clean/Circuit/CircuitType.lean
+++ b/Clean/Circuit/CircuitType.lean
@@ -25,10 +25,12 @@ class Eval (Env : Type) (Var : Type) (Value : outParam Type) where
 export Eval (eval)
 
 /-- Verifier evaluation is `Eval` specialized to `Environment F`. -/
+@[circuit_norm]
 abbrev CircuitType (F : Type) (Var : Type) (Value : outParam Type) :=
   Eval (Environment F) Var Value
 
 /-- Prover evaluation is `Eval` specialized to `ProverEnvironment F`. -/
+@[circuit_norm]
 abbrev ProverType (F : Type) (Var : Type) (Value : outParam Type) :=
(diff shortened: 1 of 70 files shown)


Clean/Circuit/CircuitType.lean
```diff
@@ -0,0 +1,138 @@
...
+  with missing holes, and each hole contains a polynomial that can refer to fixed
+  positions of the environment.
+  -/
+  Var : TypeMap
+  /-- Verifier value: hint fields are erased to `Unit`. -/
+  Value : TypeMap
+  /-- Prover value: hint fields carry their underlying type. -/
+  ProverValue : TypeMap
+  evalVerifier : ∀ {F : Type} [Field F], Environment F → Var F → Value F
+  evalProver   : ∀ {F : Type} [Field F], ProverEnvironment F → Var F → ProverValue F
```
pirapira, line 72: This decouples the prover and the verifier concerns mostly separate. That's probably the right direction.

I'm wondering whether it's needed to connect the results of `evalVerifier` and `evalProver`.

Let's say a circuit takes `(h, x)` (but `h` is hint), constrains `h^2 = x` and returns `h`. Maybe it's important that `evalVerifier` and `evalProver` do the same thing on `x`?

An example that takes both a hint and a regular input would be a good exercise.

After that, I'm also wondering about completeness proofs involving subcircuits. The completeness of the subcircuit is combined with the soundness of the sibcircuit. This will also require some relation between `evalVerifier` and `evalProver`.

pirapira (dismissed):

Clean/Circuit/CircuitType.lean
mitschabaude: > I'm wondering whether it's needed to connect the results of evalVerifier and evalProver.

good question!

I think the answer is: it seems like we don't need any connection in the generic case.
And in fact, there is no obvious generic connection that covers both ProvableType and Unconstrained (so it's good we don't need one).

> After that, I'm also wondering about completeness proofs involving subcircuits. The completeness of the subcircuit is combined with the soundness of the sibcircuit. This will also require some relation between evalVerifier and evalProver.

interestingly, no connection is needed in the generic cases that are already part of this PR: I proved the "prover spec" part of `(GenericFormalCircuit.WithHint.toSubcircuit ...).completeness`, which involves using completeness to show that the constraints hold, and then soundness to show that the spec holds. We can use completeness and soundness together here because they are bridged by a statement (`ConstraintsHold`) that mentions neither `evalVerifier` nor `evalProver`, but that is formulated in terms of concrete `Expression.eval`.

In the concrete case of ProvableType, the two evals are the same, this is even a simp lemma:
```lean
@[circuit_norm] lemma eval_var_prover_to_verifier (env : ProverEnvironment F) (v : Var M F) :
    eval env v = eval env.toEnvironment v
```

> Let's say a circuit takes (h, x) (but h is hint), constrains h^2 = x and returns h. Maybe it's important that evalVerifier and evalProver do the same thing on x?

here we are in the concrete case, where there is no ambiguity. for everything that can be constrained (= ProvableTypes, but not hints), we do know that evalVerifier and evalProver do the same.

an observation here is that you can't state any circuit operation on generic CircuitTypes. circuit operations operate on raw expressions (`assert`) or vectors of expressions (`witness`, `lookup`). so when proving something about an actual circuit, you are always in the concrete case.

pirapira: > there is no obvious generic connection that covers both ProvableType and Unconstrained

The obvious thing is to add
```
compat: Value F -> ProverValue F -> Prop
evalCompat: compat (evalVerifier env v.toEnvironment) (evalProver env v) 
```
in the signature.

I'll read `(GenericFormalCircuit.WithHint.toSubcircuit ...).completeness` to see whether this will be useful.

mitschabaude: It doesn't buy us anything to add a _generic_ property connecting `evalVerifier` and `evalProver`. there is such a `compat` for any `CircuitType` already (the one that returns `True`)

pirapira: I see. When there's a strange boolean that looks reversed to the prover and the verifier, the prover side `ProverSpec` etc will see the reversed value and that's it.

pirapira:
I think we'll see some generalizations of
```
@[circuit_norm] lemma eval_var_prover_to_verifier (env : ProverEnvironment F) (v : Var M F) :
    eval env v = eval env.toEnvironment v
```
to input & hint mixtures. That can happen in the future in the user space first.

Clean/Circuit/Foundations.lean
```diff
@@ -10,10 +10,53 @@ This ensures that the `FormalCircuit` and `FormalAssertion` definitions are not
...
 
+variable {F : Type} [Field F]
+variable {α β : TypeMap}
+
+namespace GeneralFormalCircuit.WithHint
+variable [CircuitType α] [CircuitType β]
+
+/--
+  Justification for using a modified statement for `ConstraintsHold`
+  in the `FormalCircuit` definition.
```
pirapira:
```suggestion
  in the `GeneralFormalCircuit` definition.
```

Clean/Circuit/Foundations.lean
```diff
@@ -10,10 +10,53 @@ This ensures that the `FormalCircuit` and `FormalAssertion` definitions are not
...
+    ConstraintsHold env (circuit.main b_var |>.operations offset) →
+    -- the spec holds
+    circuit.Spec (eval env b_var) (eval env (circuit.output b_var offset)) env.data := by
+  intro offset env b_var h_assumptions h_holds
+  apply circuit.soundness offset env b_var _ rfl h_assumptions
+  exact Circuit.can_replace_soundness h_holds
+
+/--
+  Justification for using modified statements for `UsesLocalWitnesses`
+  and `ConstraintsHold` in the `FormalCircuit` definition.
```
pirapira:
```suggestion
  and `ConstraintsHold` in the `GeneralFormalCircuit` definition.
```

pirapira (approved):
Looks good to me. Just nits.

PR #365, opened 2026-04-27
Add CircuitType deriving for record inputs

mitschabaude:
follow up to #363 

- Add `deriving CircuitType` support for record-like circuit input structures whose fields implement `CircuitType`.
- Generate companion `Var`, `Value`, and `ProverValue` structures and the corresponding `CircuitType` instance.
- Add focused deriver tests and a mixed provable/unconstrained input example.

### DIFF
--- a/Clean/Utils/Tactics/ProvableStructDeriving.lean
+++ b/Clean/Utils/Tactics/ProvableStructDeriving.lean
@@ -1,4 +1,5 @@
 import Lean
+import Lean.Elab.Deriving.Util
 import Clean.Circuit.Provable
 
 /-!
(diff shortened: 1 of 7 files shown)


pirapira (commented):
There seem to be some possibilities to use the utilities in `Deriving.Util` like [mkContext](https://github.com/leanprover/lean4/blob/432d11541ba76fb474d5817769f8a9b9a5fd5c4c/src/Lean/Elab/Deriving/Util.lean#L103), [mkInductiveApp](https://github.com/leanprover/lean4/blob/432d11541ba76fb474d5817769f8a9b9a5fd5c4c/src/Lean/Elab/Deriving/Util.lean#L34).

This point is not a blocker. Perhaps some maintenance will be taken care of in the upstream if we use the library.

The docstring at the top of ProvableStructDeriving can be updated. Now it only talks about `deriving ProvableStruct`.

Clean/Utils/Tactics/ProvableStructDeriving.lean
pirapira:
```suggestion
      throwError "unsupported field type for ProvableStruct or CircuitType: {fieldType}. Expected either F, M F, Vector F n, or Vector (M F) n where F is the type parameter."
```

This can be called from `mkCircuitViewStruct`.

pirapira (approved):
Looks good to me. One nit.

PR #366, opened 2026-04-28
CircuitType with properties

mitschabaude:
This is an experimental follow-up for `CircuitType` deriving. It extends the derived `CircuitType` path to handle structures with dependent prop fields, that could capture properties of the circuit type:

```lean
structure CircuitBool (F : Type) where
  bool : F
  isBool [Field F] : IsBool bool
deriving CircuitType

example (input : CircuitBool.Var (F p)) : ∀ env : ProverEnvironment (F p),
    IsBool (eval env input.bool) := by
  intro env; exact input.isBool env
```

## Why This Is Draft

The comment above `CircuitBool` captures the main issue: proof-carrying circuit inputs look attractive, but they do not compose well with values generated inside circuits from prover knowledge.

For expressions built from existing variables, the proof field can be transported syntactically. But for witnessed values, the invariant usually follows from honest witness generation, constraints, or a circuit spec. We currently do not surface that reasoning in a way that lets us construct the proof field of the generated `Var`.

For example, it is impossible to implement the simple `HasAssignEq` class for `CircuitBool`.

### DIFF
--- a/Clean/Circuit/Basic.lean
+++ b/Clean/Circuit/Basic.lean
@@ -491,25 +491,28 @@ export Circuit (witnessVar witnessField witnessVars witnessVector assertZero loo
 
 -- general `witness` method
 
-class Witnessable (F : Type) [Field F] (value : outParam TypeMap) (var : TypeMap) [ProvableType value] where
-  witness : (ProverEnvironment F → value F) → Circuit F (var F)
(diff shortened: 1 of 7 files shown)


mitschabaude:
This won't ship without a major redesign, closing

PR #367, opened 2026-04-29
Improve CircuitType proof-start normalization

mitschabaude:
Circuit-deriver follow up that makes `circuit_proof_start` work well with derived circuit types.

## Why

Derived mixed `CircuitType` inputs could leave subcircuit constraints and prover obligations phrased in terms of `eval env input_var` even after `circuit_proof_start`. That forced proofs to rewrite in the less ergonomic direction, from high-level input values back to evaluated vars.

The tactic support should match the old `ProvableStruct` ergonomics where applicable: derived `CircuitType` inputs should participate in the existing eval and equality normalization flow. Verifier values use the generated `ProvableStruct` instance, while prover-value equalities get a targeted `DerivedCircuitType` splitting path, and hand-written `CircuitType`s should still be able to preserve their abstraction boundary.

## Summary

- Add `DerivedCircuitType` as the opt-in marker for `CircuitType`s generated by the deriver.
- Replace generated per-type eval reduction lemmas with generic `DerivedCircuitType.eval_verifier` / `eval_prover` bridges that match the actual `CircuitType.verifierEval` / `proverEval` eval terms.
- Teach `simplify_provable_struct_eval` to use those generic derived eval bridges before reducing the resulting value constructors.
- Restrict the new equality-splitting path to `DerivedCircuitType` value/prover-value equalities, so ordinary non-`ProvableStruct` records remain opaque.
- Add regression coverage for mixed provable/unconstrained `CircuitType` inputs and for non-`ProvableStruct` records not being split.


Clean/Utils/Tactics/CircuitProofStart.lean
```diff
@@ -131,6 +201,10 @@ elab_rules : tactic
 
   -- simplify structs / eval first
   try (evalTactic (← `(tactic| provable_struct_simp))) catch _ => pure ()
+  try CircuitProofStart.exposeGeneratedCircuitTypeHInput catch _ => pure ()
+  try (evalTactic (← `(tactic| simp only [circuit_norm, ProvableType.eval_field,
+    CircuitType.evalVerifier, CircuitType.evalProver] at $(mkIdent `h_input):ident))) catch _ => pure ()
```
mitschabaude: if we need additional lemmas/defs beyond circuit_norm here, I don't see why we wouldn't just make them part of circuit_norm!

Clean/Utils/Tactics/CircuitProofStart.lean
```diff
@@ -5,6 +5,76 @@ import Clean.Utils.Tactics.ProvableStructSimp
...
+not be unfolded unless they provide an explicit lemma with this shape.
+-/
+private def exposeGeneratedCircuitTypeHInput : TacticM Unit := do
+  let lemmas ← generatedEvalLemmasForHInput
+  if lemmas.isEmpty then
+    throwError "no generated CircuitType eval lemma found for h_input"
+  let mut progressed := false
+  for lemmaName in lemmas do
+    try
+      evalTactic (← `(tactic| rw [$(mkIdent lemmaName):ident] at $(mkIdent `h_input):ident))
```
mitschabaude: I'm scared about how expensive rw can be when it fails. `simp only` would be much preferred since it seems to skip over non-applicable lemmas much much faster

Clean/Utils/Tactics/ProvableStructDeriving.lean
```diff
@@ -537,6 +537,83 @@ def mkCircuitValueProvableStructInstance (valueStructName : Name) (paramInfos :
...
+  let valueLemma := mkIdent (← relativeToCurrentNamespace (structName ++ `value_eq))
+  let proverValueLemma := mkIdent (← relativeToCurrentNamespace (structName ++ `proverValue_eq))
+  let evalVerifierLemma := mkIdent (← relativeToCurrentNamespace (structName ++ `eval_verifier))
+  let evalProverLemma := mkIdent (← relativeToCurrentNamespace (structName ++ `eval_prover))
+
+  let cmd ←
+    if binderSyntaxes.isEmpty then
+      `(
+        theorem $varLemma $fBinder :
+            CircuitType.Var $appliedStructType $fIdent = $varType $fIdent := rfl
```
mitschabaude: ack: this lemma must not be a simp lemma, as it rewrites away from the form that our lemmas below assume

Clean/Utils/Tactics/CircuitProofStart.lean
mitschabaude: if we ship this I require that we introduce build time regression benchmarks

Clean/Utils/Tactics/CircuitProofStart.lean
mitschabaude: I checked.. these were already in circuit_norm, so they're redundant

Clean/Utils/Tactics/CircuitProofStart.lean
mitschabaude: I found out that `simp only` just doesn't work because of a clumsy rewrite earlier that changes the form away from our simp lemma..

Clean/Utils/Tactics/ProvableStructDeriving.lean
```diff
@@ -537,6 +537,83 @@ def mkCircuitValueProvableStructInstance (valueStructName : Name) (paramInfos :
 
   elabCommand cmd
 
+/-- Generate reduction lemmas for `CircuitType` view projections and eval lemmas. -/
+def mkCircuitTypeViewReductionLemmas (structName : Name) (paramInfos : Array ParamInfo)
```
pirapira: I was asking Codex (5.5) whether this can be replaced with something taking a typeclass.

---

Codex gpt-5.5:

  I think the generated `Input.value_eq` / `Input.proverValue_eq` / eval lemma approach works, but I wonder if
  this would be cleaner as an explicit marker typeclass for derived record-backed `CircuitType`s.

  Something like:

  ```lean
  class DerivedCircuitTypeViews (M : TypeMap) [CircuitType M]
      (VarView ValueView ProverValueView : outParam TypeMap) where
    var_eq : ∀ F, CircuitType.Var M F = VarView F
    value_eq : ∀ F, CircuitType.Value M F = ValueView F
    proverValue_eq : ∀ F, CircuitType.ProverValue M F = ProverValueView F
```

Then the deriving handler would emit an instance for the generated view records instead of separate named reduction lemmas. The tactic could synthesize this class for M, get the concrete view types through the outParams, and use the class projections plus the generic CircuitType.eval_verifier / CircuitType.eval_prover.

That seems preferable to rediscovering generated facts by name convention (baseName ++ suffix), and it also preserves the intended abstraction boundary: handwritten CircuitTypes remain opaque unless they opt into this marker class.

Relatedly, I think the new split_provable_struct_eq behavior is a bit too broad. Since hasMkInjEqForType accepts any structure with a generated mk.injEq, the tactic now splits ordinary non-ProvableStruct records too. That is broader than the tactic name/docs suggest and could unexpectedly rewrite unrelated record equalities. I’d restrict the new path to generated CircuitType view records, ideally by checking the proposed marker class, while keeping the existing ProvableStruct behavior unchanged.

Comment (mitschabaude) [reply Clean/Utils/Tactics/ProvableStructDeriving.lean]: nice prompt, it ended up being something even simpler:

```lean
/--
Marker class for `CircuitType`s generated by the deriving handler.

Tactics/simp lemmas use this opt-in to expose and split generated record types while leaving
hand-written `CircuitType`s opaque by default.
-/
class DerivedCircuitType (M : TypeMap) extends CircuitType M
```

and we were able to remove _all_ per-type generated lemmas and instances!

> Relatedly, I think the new split_provable_struct_eq behavior is a bit too broad

this is also a good point that we addressed, again by only applying the additional splitting to derived circuit types

PR #237, opened 2025-07-25
[circomlib] Poseidon

mitschabaude:

PR #368, opened 2026-04-30
Poseidon follow-up

mitschabaude:
simplify proofs in #237


Clean.lean
```diff
@@ -15,6 +15,7 @@ import Clean.Circomlib.Mux2
 import Clean.Circomlib.Mux3
 import Clean.Circomlib.Rotate
 import Clean.Circomlib.Sign
+import Clean.Circomlib.Poseidon
```
mitschabaude:
```suggestion
```

PR #369, opened 2026-05-04
Update lean v4.29.0

pirapira:


Clean/Utils/Rotation.lean
```diff
@@ -8,6 +8,7 @@ open Bits (toBits toBits_injective)
 
 -- Theorems about 64-bit rotation
 
+set_option maxHeartbeats 400000 in
```
pirapira, line 11: really needed?

pirapira:
Abandoned

PR #370, opened 2026-05-05
Bump Lean to 4.30

mitschabaude:
Bumps Clean from Lean **4.28** to **4.30** (mathlib v4.30.0), skipping stabilization on 4.29 since core fixed several 4.29 simp regressions only in 4.30 (leanprover/lean4#12924, #12776; 4.29 was called a "construction-zone release" by core devs).

`lake build`, `lake build --wfail` and `lake build CleanTests` are all green; the regression tests (`SimpMatchRegression.lean`, `TestMixedCircuitType.lean`, `TestCircuitProofStart.lean`, `TestElaborateCircuit.lean`) assert that post-`circuit_proof_start` goal shapes and `elaborate_circuit`'s stored metadata are unchanged — **userland circuit proofs work as before**, and a few got simpler.

## Root cause of the fallout

Lean 4.29's `isDefEq` no longer bumps transparency to default when comparing implicit/instance arguments (leanprover/lean4#12179), and simp/dsimp no longer rewrite inside instance arguments (#12195/#12244). Clean's `circuit_norm` automation relied on both: lemmas keyed on one instance spelling used to fire on any defeq spelling. In 4.30, `inferInstanceAs` additionally wraps instances (#12897), destroying syntactic canonicity.

## Framework changes (the important part of the diff)

- **`field := fun F => F`** instead of `id` — `id` is not reducible, so `field F` was opaque exactly where the new rules demand transparency.
- **Instance architecture around the `field` synonym rebuilt**: the ~60 bespoke operator instances on `field F` / `Value field F` / `ProverValue field F` are deleted (with a transparent synonym they resolve canonically — and would otherwise be self-looping). Only `Var field F` (projection-headed, invisible to TC goal normalization) keeps bespoke instances, now `@[reducible]` with canonical bodies via `(inferInstance : _)`.
- **Priorities**: `(priority := high)` on the `fields n`/`fieldPair`/`fieldTriple` `ProvableType` instances (the transparent synonym lets `ProvableVector`/`ProvablePair` capture them, dead-keying all their `circuit_norm` lemmas); `(priority := low)` on the generic `Zero (α F)`/`Inhabited` instances.
- **`@[implicit_reducible]`** on `Circuit` and the `localLength` definitions; `@[reducible]` on the `CircuitType`-producing instances — restores defeq through implicit args/match discriminants where it is genuinely needed.
- **`circuit_proof_start`** now collapses the `field`/`Var`/`Value`/`ProverValue` synonyms in all hypotheses and goals; the whole tactic layer (`circuit_proof_start`, `provable_struct_simp`, `simplify_provable_struct_eval`) runs its simp passes with `+instances`.
- **`elaborate_circuit`** runs its internal metadata normalization with `instances := true` for the same reason — without it, stored `ElaboratedCircuit` metadata degrades to huge unreduced terms wherever a loop body is a bind-chain of subcircuits (Keccak ThetaC → Theta → KeccakRound → Permutation, 33KB `localLength`), which is what previously forced `maxRecDepth` bumps and several of the scoped legacy-defeq options.
- **`ProvableStruct` deriving now also generates a per-struct `@[circuit_norm]` rfl-lemma `Foo.fromComponents_cons : fromComponents (.cons a …) = Foo.mk a …`**, and the tactic layer stops unfolding `ProvableStruct.fromComponents` through its match equations: 4.29 simp leaves those as stuck dependent `HList` matches (minimal repro: `SimpMatchRegression.lean`; structurally fixed in 4.31). The generated lemma avoids the private `match_1` equation theorem and its non-normal implicit types, so it is the more robust mechanism independent of that regression. New `eval_var_eq_eval`/`_prover` bridge lemmas feed the same pass.
- A generic `eval_unit` lemma (any `Eval` into `Unit` is trivial by unit-eta) replaces instance-shape-sensitive hint-erasure lemmas.

## Recurring proof-fix recipes (applied across ~40 gadget/table files)

1. `simp +instances only [circuit_norm, …]` when a keyed lemma stops matching through instance args — the single most common fix; when `ElaboratedCircuit.localLength <sub>.main` is stuck, also unfold the subcircuit's **`.circuit` constant**.
2. `exact`/`show`/defeq application instead of `rw` where LHS and goal differ only up to defeq (vector length indices, `TableContext.empty` spellings, derived `fromComponents_cons` on nested structures).
3. Mathlib drift: `ZMod.val ↑n` now normalizes to `n % p`.
4. Deleting simp steps that the stronger `circuit_proof_start` normalization made redundant.
5. The two heavyweight `FormalTable` literals (Fibonacci8/32) still hit whnf/isDefEq timeouts and keep a **declaration-scoped** `set_option backward.isDefEq.respectTransparency false`, marked `-- TODO(4.30 bump)` (grep the marker — likely removable on 4.31, which fixes the match-auxiliary blockage structurally via leanprover/lean4#13281/#13363). All other early workarounds (BLAKE3/Keccak/AliasCheck legacy defeq, every `maxRecDepth` bump) turned out to be fixable and are gone.

## History

Earlier commits on this branch (pre-merge) were the initial 4.29 exploration, which already landed most of the core adaptations: toolchain bump, core circuit/`Vector`/`ProvableType` fixes, `+instances` across the tactic layer, the generated `fromComponents_cons` lemma, and the simp/match regression test. The branch was then brought up to date with main (809 commits, including witgen-ir) and moved to 4.30, where the remaining per-proof fallout was fixed following the recipes above.

The known blanket escape hatch — pinning `backward.isDefEq.respectTransparency false` project-wide — was deliberately **not** used, since it would freeze us on legacy defeq behavior that upstream changed for performance reasons.


pirapira (approved):
Looks good to me.

pirapira:
Merged because the diff is big.

PR #373, opened 2026-05-20
foldlRange extension

TalDerei:
References https://github.com/tachyon-zcash/ragu/pull/710#pullrequestreview-4317615259 which doesn't currently seem feasible without upstream changes.

Adds a `ConstantOutput`-collapsed variant of `foldlRange.output_eq` in `Clean/Circuit/Loops.lean`.

### Motivation

`Circuit.foldl` already exposes a single-application output via `foldl.output_eq`, because it bundles `ConstantOutput` in its signature. `Circuit.foldlRange` doesn't bundle it, so `foldlRange.output_eq` produces a residual `Fin.foldl m step init` term — even when the body's output is acc-independent (and thus reducible via `Fin.foldl_const`).

This PR adds the analogous lemma:

```lean
lemma foldlRange.output_eq_of_constantOutput [NeZero m]
    (h_const_out : ConstantOutput fun (t : β × Fin m) => body t.1 t.2) :
  (foldlRange m init body constant).output n =
    (body default ⟨m - 1, Nat.pred_lt (NeZero.ne m)⟩).output
      (n + (m - 1) * (body default default).localLength)
```

Callers supply `h_const_out` (typically dischargeable by `intros; rfl` for bodies that allocate fresh wires and don't reuse the accumulator in the returned expression).

### Downstream use case

For example, the Ragu FV pipeline rewriting `Element::enforce_root_of_unity` as a polymorphic-in-`k` reimpl with `Circuit.foldlRange`: the formal-instance `same_output` proof needs the symbolic output to match an extracted flat-trace expression `var ⟨n+3k-1⟩`. With the existing `foldlRange.output_eq` alone, the goal stalls on `Fin.foldl k (fun acc _ => …) init`. The new lemma discharges it in one step.

### Proof

Mirrors `foldl.output_eq`: unfold via `FoldlM.output_eq`, normalize the `finRange` indexing, rewrite each step's body via the supplied `h_const_out`, then `Fin.foldl_const_succ` collapses to the last application.

### Tests

Full `lake build` passes (1771 jobs).

### DIFF
--- a/Clean/Circuit/Loops.lean
+++ b/Clean/Circuit/Loops.lean
@@ -882,6 +882,32 @@ lemma foldlRange.output_eq :
   congr! 6
   rw [constant.localLength_eq (_, _)]
 
+/--
+Variant of `foldlRange.output_eq` for bodies whose output is independent of their
+input pair (`ConstantOutput`). Mirrors `foldl.output_eq`: the output collapses to a
+single application of `body` at the last index, eliminating the residual `Fin.foldl`.
+Users must supply the `ConstantOutput` proof explicitly because `foldlRange`, unlike
+`foldl`, does not bundle it in its definition.
+-/
+lemma foldlRange.output_eq_of_constantOutput [NeZero m]
+    (h_const_out : ConstantOutput fun (t : β × Fin m) => body t.1 t.2) :
+  (foldlRange m init body constant).output n =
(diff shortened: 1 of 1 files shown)


TalDerei:
closed based on OOB clarification that we shouldn't explicitly call simplification lemmas.

PR #375, opened 2026-05-21
Automatic circuit elaboration

mitschabaude:
This is intended to automate the elaboration of "explicit" circuit data: `localLength`, `output`, `channelsWithGuarantees` and `channelsWithRequirements` in `ElaboratedCircuit`.

- The role of `ElaboratedCircuit` is changed to be a field of `FormalCircuit` rather than the parent class. Instead, we have a new parent class called `FormalCircuitBase`. This gives us more flexibility in deriving the `ElaboratedCircuit` instance, and it's more logical that it depends on `main` so it's actually uniquely derivable.
- **Main change**: Add a family of tactics that automatically construct `ElaboratedCircuit`. Builds on the existing (previously unused) `ExplicitCircuit` typeclass infra. This makes it painless to add a new "leaf case" to auto-derivation: just add an `ExplicitCircuit(s)` instance for your case.
- Downstream effect: get rid of manual circuit elaboration in almost all gadgets
- Side effects:
  - Needed to shuffle around some code so that the `Formal*` circuits can already depend on the elab tactic for the default value. Split off `Circuit/Formal.lean` from `Circuit/Basic.lean` and moved a few lemmas to fix dependency fallout.
  - Moved `exposedChannels`, which is an opt-in annotation and shouldn't be auto-derived, from `ElaboratedCircuit` to `FormalCircuitBase`
  - Added `explicit_circuit_no_unfold` attribute to mark "primitive" circuits that should not be unfolded by the derivation tactic. This lets `ExplicitCircuit` and `ElaboratedCircuit` inference be quite aggressive in unfolding names like `MyCircuit.circuit` that would hinder it in deriving concrete values

## TODOs (possibly follow-up)

- [x] rename the tactics
- [x] `===` should be recognized reliably as a base instance (there is one circuit where this doesn't work)
- [x] handle circuits with a `match` statement, this is the main class of circuit where autoelab still doesn't work
- [x] ideally, overriding tactics should not leave `.withData` in place but reduce it and construct the transport proof internally
- [x] fix Poseidon (not in the module graph, so it's easily missed. it has a sorry that's why it can't be in the module graph)

## Should we ship this?

**EDIT: this section is no longer relevant. The PR improves overall heartbeat count and the main regressions are just core modules that grew more code**

This PR is fantastic for making our framework nicer-looking and more intuitive to use.

What's _not_ great is that there are still a handful of modules where autoelab doesn't work. In cases like that, it'll be less clear to users how to fix the problem compared to the current state where every circuit acts as an example.

What I dislike the most is that with the PR, some previously very cheap modules now become dominated in cost by the autoelaboration:
Screenshot from 2026-05-28 09-36-54

We should put in some work to give autoelab better happy & fast paths to avoid this.


pirapira (approved):
I got the basic idea and the diff is big. I'll review in detail later.

PR #383, opened 2026-05-26
[codex] Add Lean build benchmark suite

mitschabaude:
## Summary

Adds a mathlib-style Lean build benchmark suite for tracking build regressions with instruction counts instead of noisy wall-clock timings.

## Changes

- Adds `scripts/bench/measure.py` for JSONL metric capture using `perf` and rusage.
- Adds a build benchmark that runs Lake with `LAKE_OVERRIDE_LEAN` and a Lean wrapper, so per-module measurements come from the actual Lake build graph.
- Records whole-build metrics, per-module instruction counts, per-module line counts, and Lean profile buckets.
- Adds `scripts/bench/report.py` for current and baseline-vs-current benchmark summaries.
- Adds a `Bench` GitHub Actions workflow that enables perf counters, runs the benchmark, renders a summary, and uploads `measurements.jsonl`.

## Validation

- `python3 -m py_compile scripts/bench/measure.py scripts/bench/report.py scripts/bench/build/fake-root/bin/lean`
- `bash -n scripts/bench/run scripts/bench/build/run`
- Synthetic JSONL smoke test for `scripts/bench/report.py`

I did not run the full benchmark locally because this machine has `perf_event_paranoid=4`; the benchmark preflight exits before `lake clean` or `lake build` in that case.

### DIFF
--- a/.github/workflows/bench-command.yml
+++ b/.github/workflows/bench-command.yml
@@ -1,0 +1,49 @@
+name: Bench Command
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  dispatch:
+    if: >
+      github.event.issue.pull_request &&
+      (github.event.comment.body == '/bench' || startsWith(github.event.comment.body, '/bench ')) &&
+      contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association)
+    runs-on: ubuntu-latest
+    permissions:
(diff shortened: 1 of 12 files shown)


mitschabaude:
Closing in favor of #384, which has the same final benchmark infrastructure on a fresh branch to avoid stale required-check state from the earlier workflow iterations.

PR #390, opened 2026-05-28
[codex] Cache heartbeat benchmark baselines

### LINKED ISSUE #389: Optimize benchmark caching for heartbeat-only reports
## Context

The heartbeat benchmark is now the primary/default module metric, so we no longer need to optimize the benchmark around collecting whole-build instruction counts. That changes the caching tradeoffs: heartbeats are collected by re-elaborating Lean frontend commands, while instruction counts currently force us through a conservative clean instrumented build path.

We already have persistent Lean/Lake dependency caching for the runner: `.lake/packages` is keyed by `lean-toolchain` plus `lake-manifest.json`, and the benchmark runs `lake exe cache get`. This issue is about optimizations beyond that existing dependency cache.

## Goal

Reduce `/bench` runtime by reusing safe state where possible, while keeping heartbeat results meaningful and comparable between the baseline and PR commit.

## Possible optimizations

- Build on the existing persistent `.lake/packages` cache and remove instruction-count-driven constraints that force conservative clean builds.
  - Stop requiring `perf` availability if instruction counts are no longer reported.
  - Avoid `lake build --no-cache` when it is no longer needed for instruction-count measurement.
  - Reconsider whether `rm -rf .lake/build` is still necessary for heartbeat-only benchmarking.

- Store heartbeat measurements for exact `main` commits.
  - Run the heartbeat benchmark on every push to `main`.
  - Store the resulting JSONL/report artifact in a way the PR benchmark can retrieve by exact baseline SHA.
  - Reuse stored heartbeat results only when they were produced for the exact same `main` commit.
  - If no stored result exists for the requested baseline SHA, fall back to measuring the baseline in the PR benchmark.

- Keep PR-head measurement fresh.
  - Always measure the PR checkout directly.
  - Do not cache or reuse PR heartbeat measurements across different SHAs/checkouts.
  - Do not reuse heartbeat measurements for unchanged files from another checkout: heartbeat cost can change substantially when dependencies change.

- Consider splitting dependency preparation from measurement.
  - A preparation step can populate toolchain, package, and dependency build caches.
  - The measured step should then focus on Clean modules and heartbeat extraction.

## Acceptance criteria

- `/bench` still compares the requested baseline SHA and PR SHA.
- Reports continue to show heartbeat-based top module regressions/improvements by default.
- Dependency artifacts are not repeatedly fetched on a warm runner.
- Stored heartbeat results are reused only for exact `main` commits.
- PR-head heartbeat results are always measured from the requested checkout.
- Documentation explains which caches are persistent, which measurements may be reused, and which parts are intentionally rebuilt.

## Notes

This is follow-up work after making heartbeats the default metric. It is okay for the current heartbeat version to keep the conservative clean rebuild behavior until this is addressed.

mitschabaude:
## Summary

Implements the first cache optimization pass for heartbeat-based `/bench` runs.

- Adds a `Bench Main` workflow that runs the heartbeat benchmark on every push to `main` and stores exact-SHA baseline JSONL on the self-hosted runner.
- Reuses cached baseline measurements for exact base SHAs during PR benchmark runs, with fallback to measuring and storing the baseline on cache miss.
- Keeps PR-head measurements fresh: the current PR SHA is still always checked out and measured directly.
- Makes build benchmarks heartbeat-only and removes the old instruction/perf path.
- Warms the checkout with `lake exe cache get` and `lake build`, then enumerates built `Clean` modules in a parallel heartbeat pass so reports stay complete even when Lake reuses build artifacts.
- Passes each module's Lake setup file to heartbeat measurement so package/module options match the Lake build.
- Serializes cache-mounted benchmark containers with a host `flock` and prunes old per-PR build caches.
- Uploads a full alphabetically sorted module table artifact for deeper investigation.
- Adds local `scripts/bench/heartbeats` and `scripts/bench/heartbeat-diff` helpers.

Closes #389 once we are happy with this first-pass behavior.

## Notes

This intentionally does not cache PR-head heartbeat results and does not reuse per-file heartbeat measurements across checkouts.

Baseline cache files are stored under `/var/lib/clean-bench/cache/baselines/<sha>.jsonl`.

## Validation

- `bash -n scripts/bench/heartbeat-common.sh scripts/bench/heartbeats scripts/bench/heartbeat-diff scripts/bench/build/run scripts/bench/build/heartbeats-all scripts/bench/runner/run-in-docker.sh`
- `python3 -m py_compile scripts/bench/measure.py scripts/bench/report.py`
- `git diff --check`
- Smoke-rendered existing heartbeat report and full module table reports.
- Manual warm-cache runner validation on `kevin` after review fixes: baseline reused, current build hot, setup files passed to heartbeat workers, JSONL counts `304` baseline / `308` current, report rendered successfully, current output completed in about 6m34s.


mitschabaude:
Note from manual runner validation: this PR currently regresses end-to-end benchmark runtime despite the cache improvements.

What we observed:

- `main` completed a `/bench` run in about 27 minutes.
- This branch's warm-cache run took about 35 minutes.
- The cache pieces are working: baseline JSONL was reused, `lake exe cache get` reported `No files to download`, and the warm current `lake build` took about 1.56s.

Likely cause:

- On `main`, heartbeat measurement is performed via the fake `lean` wrapper during `lake build --no-cache`, so Lake schedules module work in parallel.
- On this branch, we switched to `lake build` followed by `scripts/bench/build/heartbeats-all`.
- `heartbeats-all` currently measures modules sequentially, one Lean process per module.

So the PR removes baseline work, but loses Lake's parallelism for the heartbeat sweep. The caching idea is still sound, but before landing this as an optimization we should parallelize the heartbeat pass, probably with a conservative worker count controlled by something like `BENCH_HEARTBEAT_JOBS`.

mitschabaude:
Update: the heartbeat sweep is now parallelized in `265221fa`.

Manual warm-cache validation on `kevin`, same `main` vs `autoelaborate` SHAs as before:

- baseline JSONL reused
- `lake exe cache get`: no downloads
- current `lake build`: hot, report wall time ~1.73s
- end-to-end runner output: started `14:45:38`, `current.jsonl` written `14:54:59`, so about 9m21s
- report generation succeeded with the same 304/308 JSONL line counts as before

This addresses the earlier regression: the previous version measured heartbeats sequentially after the build and took ~35 minutes warm; this version uses four heartbeat workers by default (`BENCH_HEARTBEAT_JOBS` can override it).

mitschabaude:
## Review

Read the full diff plus surrounding workflow/script state and verified a couple of behaviors empirically. Solid, well-structured PR — the pivot to deterministic per-module heartbeats + cached per-SHA baselines is the right direction, dropping `perf` is a nice security win (no more `--cap-add PERFMON --security-opt seccomp=unconfined`), baseline writes are atomic, and `baseline_cache_path` correctly validates the SHA against path traversal. All local validation (`bash -n`, `py_compile`, `git diff --check`) passes.

One concrete data-loss bug, plus a few architectural/operational things worth deciding before this runs unattended.

### 🔴 1. A single failing module discards *all* heartbeat data (`heartbeats-all`)

`heartbeats-all` runs under `set -euo pipefail`, and GNU `xargs` exits **123** if any child fails. Confirmed the consequence:

```
before xargs
EXIT=123          # collection loop never runs; trap rm's the tmpdir
```

So if **one** module fails its heartbeat re-elaboration, `set -e` aborts *before* the `while … cat` collection loop, and the `trap` then deletes every per-module `.jsonl`. Result: the report shows **zero** modules, not "all but the bad one." The failure path is reachable — when `measure_heartbeats` returns non-zero, `value=""`, then the Python heredoc does `float("")` → `ValueError` → non-zero → xargs 123 (also prints a stray traceback).

Fix: make `measure_module` swallow per-module failure (`if [ -z "$value" ]; then echo "skip $module" >&2; return 0; fi`) and/or `xargs … || true`, so partial results survive and get logged.

### 🟠 2. New shared build caches race with no concurrency guard

`run-in-docker.sh` now bind-mounts persistent build caches:
- `cache/lake-build/<key>` — shared by **main baseline runs *and* PR cache-miss baselines** (keyed only by toolchain+manifest, not SHA)
- `cache/pr-lake-build/<repo>/<pr>/<key>` — per PR

Two jobs mounting the same `lake-build/<key>` into `.lake/build` and running `lake build` concurrently will corrupt it. Neither `bench.yml` nor the new `bench-main.yml` has a `concurrency:` group — only safe today via implicit single-runner serialization, but the shared cache turns concurrent runs from *slow* into *corrupting*. Suggest a `concurrency:` group spanning both workflows (or a host `flock` around the cache-mounted `docker run`) as defense-in-depth.

### 🟠 3. `pr-lake-build` caches grow unbounded

`cleanup()` only removes `RUN_DIR`; everything under `CACHE_DIR` persists by design. `pr-lake-build/<repo>/<pr>/<key>` is a full project build tree per PR with no eviction — GBs per PR forever. The others (`lake-build`, `lake-packages`, `mathlib-cache`) are bounded by manifest keys, but the per-PR caches need a cleanup policy (prune on PR close, or LRU by mtime).

### 🟡 4. `build//wall-clock` and `build//maxrss` are now warm-cache noise

With the persistent `.lake/build` mount and plain `lake build` (no `--no-cache`), these whole-build metrics now measure an *incremental* rebuild of only the changed modules — a 1-file PR shows near-zero wall time. Heartbeats are the real signal (re-measured deterministically regardless of cache), which is the point, but the summary table still presents "Wall time" / "Max RSS" with equal prominence. Consider dropping them from the summary or labeling them as incremental.

### Minor

- **leanOptions not replicated.** The heartbeat helpers run `lake env lean --run … "$path"` with no `--setup`, so the package's `autoImplicit := false` / `relaxedAutoImplicit := false` aren't applied — measured under default `autoImplicit := true`. Benign for *these* options (code valid under the strict options elaborates identically under the looser ones), but a fidelity gap vs. the old in-build measurement: a future option affecting elaboration would silently diverge.
- **HeartbeatCheck re-elaborated per module.** `lake env lean --run` interprets the checker afresh for every module (×N, ×6 workers); the old fake-root path compiled it once. Pre-compiling to an exe would cut constant overhead.
- **`heartbeat-diff` edge cases:** `fetch --all` runs every invocation (fails offline under `set -e`); `current=$(measure_heartbeats …)` masks `set -e` via assignment, so a failed measurement yields `current=""` and a confusing `$((…))` error later. Fine for a dev helper, just noting.
- **PR description outdated:** the "Notes" section says baselines live at `baselines/heartbeats-v2/<sha>.jsonl`, but the code/README use `baselines/<sha>.jsonl` (no version segment — intentional, per the "Simplify baseline cache path" commit). Worth updating the description to match.

**Bottom line:** design is good. I'd block on **#1** (real, easy-to-hit data loss) and get explicit decisions on **#2/#3** before unattended runs. #4 + minors are quick polish.

PR #395, opened 2026-06-02
Casual Sunday SHA256 Optimizations With Claude

rot256:
Who would win?

- A decade of some of the smartest humans trying to optimize (non-deterministic) SHA-256 circuits.
- A big matrix with 4 hours, cvc5 and Sagemath.

I got Claude to optimize (for R1CS) the verified implementation of SHA-256, that @gio54321 created. After a few rounds of `\goal Optimize the SHA-256 compression. Do not change the spec. Consider using cvc5 and Sagemath` the result is this PR. The primary tricks he came up with where:

- Instead of doing `AddMod32` on separate operands, he simply adds *multiple* things unreduced over the field, then witnesses the carry bits and bit decomposition for the result. 

- The other big win was the following two R1CS relations:

     - `(o + 2a + 2b + 7c) * (a + b − 4c + 1) = 6a + 6b − 24c` : XOR3, 3-way XOR in a single constraint.
     - `(o + a + b − 9c + 3) * (a + b + 6c − 4) = −12` : MAJ3, majority gate in a single constraint.
   
   Which he discovered using Gröbner basis.
   
I haven't really looked at the code in detail, but the neat trick is that I don't have to: he did not change the spec and the proof passes, so he (Claude) could not possibly screw this up, so ¯\\_(ツ)_/¯

Here is a table of different implementations that I / "Deep Research" could find:

| Impl | Constraints / block | Year |
|---|---|---|
| **Clean** (Lean-verified) | **17,824** | 2026 |
| **xJsnark** (Kosba, UMD) | 25,538 | 2018 |
| **jsnark** (Kosba) | 25,650 | 2016+ |
| **bellman / bellperson** (Sapling) | 25,840 | 2018+ |
| **bkomuves/hash-circuits** (Faulhorn) | 26,170 | 2023–25 |
| **libsnark** (SCIPR Lab) | 27,904 | 2014–16 |
| **Zcash / Zerocash** | 27,904 | 2014–16 |
| **circomlib** (iden3) | 29,380 | ~2019– |
| **gnark** (Consensys) | ~30,000 | 2020+ |
| **arkworks** | ~40,000 | ~2021+ |

In other words, this formally verified impl is SOTA, and by quite a margin.


rot256:
Btw, should this go in Clean or should we open a separate repo for a "circuit library"?

@mitschabaude @gio54321

mitschabaude:
@rot256 I think it should go in the clean repo, I like the monorepo style

PR #398, opened 2026-06-08
Refactor `channelsWithRequirements` to support pushes/pulls with a boolean `enabled` flag

mitschabaude:
> this PR description is **fully human-written**

The goal of this PR was to support doing conditional pulls and pushes, as is typical in VMs for padding.

The refactor ends up being a surprisingly large change to the library core.
Source of complexity is that, when the multiplicity is an arbitrary `enabled` or `-enabled` element, we no longer know statically that it is 0, -1 or 1. It is the _constraints of the circuit_ that force it to one of those values.

## Motivating background

Previously we could assume that `channelsWithGuarantees` and `channelsWithRequirements` can be derived statically from the circuit operations:
* guarantees are gated behind the opt-in `assumeGuarantees` flag plus $-1$ multiplicity
* requirements are gated behind multiplicity $\notin 0, -1$ (0 was added in this PR)

With `enabled` flags, multiplicities are no longer statically known! The problem with that is that we would need to conservatively include channels in `channelsWithRequirements` that we conditionally _pulled from_, like Add8 in the FibonacciCircuit. That, in turn, would break the "ordered channels" soundness argument.

For guarantees it's still fine because we have the static `assumeGuarantees` flag: circuits already opt into using guarantees explicitly and statically, and if they don't, then the channel won't be in `channelsWithGuarantees`. Opting out statically here is possible because guarantees are a _premise_, and it's fine to not use a premise without affecting validity of your statement. This is different for requirements which are a conclusion.

The most natural way to solve this was to take into account the actual constraints in proving `channelsWithRequirements` correct, so that we can use the boolean constraint for `enabled`. E.g., now we know that the Add8 interaction is always a pull, and so we retain `OrderedChannel` for the Add8 channel.

## What changed

* Ripped out `channelsWithRequirements` from the entire auto-elaboration machinery (`Explicit.lean`, `ElaboratedCircuit`). Instead it lives as a normal, explicit field on `FormalCircuitBase`. This seems fine because channel pushes are usually done very rarely, at the highest circuit layer (they act as "top level circuit output"); so writing them out won't create noise.
* Added a tactic that manages to automatically prove the default `channelsWithRequirements := []` correct on all of our gadgets except 3. I.e., no disruption to existing gadgets
* On the AIR layer, changed many statements that need to reason about channel requirements to have an additional `Constraints` hypothesis. Luckily the overall AIR ensemble arguments still go through, with almost no user interface changes (they ultimately already had constraints as one of their assumptions that could be threaded through).
* Ported the VM soundness argument from assuming -1/1 multiplicities to support the padded version. This is the biggest change in terms of code size since several long theorems had to be adapted.
* Moved Fibonacci to use an `enabled` flag for padding interactions to demonstrate the new layer is functional

### DIFF
--- a/Clean/Circuit/Basic.lean
+++ b/Clean/Circuit/Basic.lean
@@ -1,4 +1,5 @@
 import Clean.Circuit.Operations
+import Clean.Utils.Field
 import Mathlib.Control.Monad.Writer
 
 variable {F : Type} [Field F] {α β : Type} {n : ℕ}
@@ -472,12 +473,6 @@ attribute [circuit_norm] Fin.coe_ofNat_eq_mod
 
 -- simplify `vector[i]` (which occurs in ProvableType definitions) and similar
 attribute [circuit_norm] Fin.val_eq_zero Fin.cast_eq_self Fin.coe_cast Fin.isValue
-
-@[circuit_norm]
-lemma zero_ne_neg_one {F : Type} [Field F] : (0 : F) ≠ -1 := by
-  intro h
-  apply zero_ne_one (α:=F)
-  rw [← neg_zero, h, neg_neg]
 
 -- simplify constraint expressions and +0 indices
 attribute [circuit_norm] neg_mul one_mul add_zero zero_add neg_zero neg_eq_zero one_ne_zero zero_ne_one
--- a/Clean/Circuit/Basic.lean
+++ b/Clean/Circuit/Basic.lean
@@ -141,8 +141,8 @@ def Channel.pull {Message : TypeMap} [ProvableType Message] (channel : Channel F
 
 @[circuit_norm]
 def Channel.pullIf {Message : TypeMap} [ProvableType Message] (channel : Channel F Message)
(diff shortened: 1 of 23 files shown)


Clean/Circuit/Basic.lean
```diff
@@ -461,8 +473,15 @@ attribute [circuit_norm] Fin.coe_ofNat_eq_mod
 -- simplify `vector[i]` (which occurs in ProvableType definitions) and similar
 attribute [circuit_norm] Fin.val_eq_zero Fin.cast_eq_self Fin.coe_cast Fin.isValue
 
+@[circuit_norm]
+lemma zero_ne_neg_one {F : Type} [Field F] : (0 : F) ≠ -1 := by
```
mitschabaude: `Basic` this seems like the wrong place for this lemma, can you move it to `Utils.Field` please?

pirapira (approved):
Looks good to me.

The shallow version sounds like a useful direction.

pirapira:
Merged because it's big.

PR #401, opened 2026-06-09
[codex] Compile circuit witnesses to checked array writes

mitschabaude:
## Summary

Adds a prototype witness compiler in `Clean/Circuit/Witness.lean` that reuses the explicit circuit normalization pipeline to inline witness callbacks and generate Lean witness-array functions.

The generated functions now take an `offset`, a preallocated witness array, and a static local-length bounds proof, then write directly with checked `Array.set` operations and checked array reads instead of building intermediate witness vectors or using `getD`/`set!` in the generated code.

`Clean/Examples/Witness.lean` contains small maintained examples that exercise scalar witnesses, dependent witnesses, and a `Vector.mapRange` witness callback.

## Validation

- `lake build Clean.Circuit Clean.Examples.Witness`
- Full `lake build` passed after merging `origin/main` earlier in this branch.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -131,6 +131,7 @@ theorem soundness : Soundness F elaborated Assumptions Spec := by
 - Assumptions capture preconditions (e.g., value ranges)
 - Follow Mathlib naming conventions
 - Never modify maxHeartbeats
+- Keep core library files sorry-free. Temporary or example-only sorries are acceptable only in example/test files when explicitly intended.
 
 ## Key Files to Understand
 
--- a/Clean.lean
+++ b/Clean.lean
@@ -18,6 +18,7 @@ import Clean.Circomlib.Sign
 import Clean.Examples.AddOperations
 import Clean.Examples.Add32Explicit
 import Clean.Examples.ToJson
+import Clean.Examples.Witness
 import Clean.Examples.HintExample
 import Clean.Examples.FemtoCairo.FemtoCairo
 import Clean.Examples.FemtoCairo.Plonky3Helpers
--- a/Clean/Circuit.lean
+++ b/Clean/Circuit.lean
@@ -2,6 +2,7 @@ import Clean.Circuit.Formal
 import Clean.Circuit.Foundations
 import Clean.Circuit.LookupCircuit
 import Clean.Circuit.Explicit
+import Clean.Circuit.Witness
 import Clean.Circuit.Extensions
 import Clean.Utils.Tactics.ProvableStructDeriving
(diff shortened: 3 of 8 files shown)


mitschabaude:
Current assessment of the witness compiler approach:

The present prototype is useful as a probe and has already shaken out real cases, but I do not think the current architecture will scale to the whole codebase, especially not Keccak-sized gadgets.

The main issue is that the compiler is still too normalization-driven. It compiles witness callbacks by repeatedly normalizing Lean expressions and pattern-matching whatever shape happens to result. That means small changes in `simp`/`whnf` behavior can change sharing, generated code shape, and performance. We already saw this with `Vector.map`: one path fixed `Mux1`, but also duplicated `fieldToBits` vectors in `Num2Bits` until direct mode was narrowed again.

The timeout failures are also consistent with this diagnosis. The expensive reductions are not really circuit semantics; they are Lean container implementation details such as `Vector.toArray`, `Array.toList`, `List.concat`, `Vector.mapRange`, and casts. Broad `whnf`, especially under high transparency, is the wrong tool here. `elaborate_circuit` carefully avoids this style, and the witness compiler should follow that more closely.

What I would change:

1. Stop treating normalization as the compiler. Use targeted `dsimp`/`simp` with the existing `explicit_circuit_norm` mechanisms only to expose known constructors.
2. Introduce a small witness IR and lower callbacks into that IR through an explicit frontend. The IR would have assignments, reads, arithmetic/value expressions, lets, conditionals, loops, and subcircuit calls.
3. The IR still has to come from Lean callbacks, but the frontend should accept a deliberately small Lean subset: `Expression.eval`, `Environment.get`, arithmetic, `let`, `if`, tagged loop/vector constructors, and controlled `ProvableType.toElements`/`fromElements` cases. Unsupported terms should fail with useful errors.
4. Preserve loops and structured vector operations instead of reducing through `Vector`/`Array`/`List` internals.
5. Generate compositional witness functions for formal subcircuits and call them instead of flattening everything by default.
6. Separate code generation from bounds-proof generation. For large circuits, generate looped code and use reusable bound lemmas from a hypothesis like `offset + localLength <= witnesses.size`, instead of asking `get_elem_tactic` to rediscover large index facts inside huge generated terms.

So the current branch is still valuable as an MVP and testbed for small circuits, but I would not try to scale it by adding more ad hoc cases around normalized Lean terms. The next serious milestone should be removing broad `whnf` from `Witness.lean` and replacing the callback compiler with a controlled lowering pass into a stable witness IR.

mitschabaude:
closing bc it's superseded

PR #406, opened 2026-06-12
Rewrite Rotation proof to avoid bv_decide timeout

### LINKED ISSUE #399: Rewrite Rotation proof to avoid bv_decide timeout
During the channel PR validation, `lake build Clean` intermittently failed in `Clean.Utils.Rotation` because the `bv_decide` proof for `h_sat` timed out:

```lean
have h_sat : offset_bv < 64 → offset_bv > 0 →
    (x_bv % 1<<<offset_bv) <<< (64 - offset_bv) = x_bv <<< (64 - offset_bv) := by
  bv_decide
```

Location: `Clean/Utils/Rotation.lean`, inside `rotRight64_eq_bv_rotate`.

This is showing up often enough that the proof should be rewritten to be robust, rather than relying on SAT solver behavior or increasing timeouts.

Expected fix:

- Replace the `bv_decide` step with a proof that is stable under rebuilds and CI load.
- Prefer a small structural lemma about masking low bits before a left shift, specialized to the 64-bit/offset range if that keeps the proof simple.
- Keep the proof safe from solver timeouts; do not address this by changing `maxHeartbeats` or global timeout settings.

Observed behavior:

- Full `lake build Clean` can fail at this proof by solver timeout.
- Retrying `lake build Clean.Utils.Rotation` alone may pass, which suggests solver/runtime instability rather than a deterministic type/proof error.

VanshSahay:
## Summary
- Replaces two `bv_decide` calls in `Clean/Utils/Rotation.lean` with a
  deterministic `Nat` lemma `mod_shiftLeft_eq_shiftLeft_mod_pow`
- The shared lemma serves both `rotRight64_eq_bv_rotate` and
  `rotRight32_eq_bv_rotate`, eliminating ~95 lines of duplicated
  UInt64/UInt32 wrapper code
- Fixes the intermittent SAT solver timeout reported in #399

## Test plan
- [x] `lake build Clean --wfail` passes with no warnings

Closes #399

### DIFF
--- a/Clean/Utils/Rotation.lean
+++ b/Clean/Utils/Rotation.lean
@@ -6,6 +6,29 @@ import Clean.Utils.Bits
 namespace Utils.Rotation
 open Bits (toBits toBits_injective)
 
+/--
+  When shifting left by `N - o` in an N-bit context, the low `o` bits of `x`
+  do not affect the result -- they get shifted past the N-bit boundary.
+-/
+lemma mod_shiftLeft_eq_shiftLeft_mod_pow (x o N : ℕ) (ho_lt : o < N) :
+    ((x % (1 <<< o)) <<< (N - o)) % (2^N) = (x <<< (N - o)) % (2^N) := by
+  simp [Nat.shiftLeft_eq]
+  have h_pow : 2^o * 2^(N - o) = 2^N := by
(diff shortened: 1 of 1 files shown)


mitschabaude (approved):
great, thanks a lot for this!

PR #410, opened 2026-07-03
BabyJubJub spec and point addition circuit

VanshSahay:
## Summary

Adds the BabyJubJub elliptic curve specification and a point addition circuit, following circomlib's BabyAdd template. Part of the circomlib formalization effort ([issue #155](https://github.com/Verified-zkEVM/clean/issues/155)).

### Spec file (`Clean/Specs/BabyJubJub.lean`)

- [x] **Point type** — `Point F` with `x : F`, `y : F`, and `Zero`/`Neg` instances (identity = `(0, 1)`, negation = `(-x, y)`)
- [x] **Curve constants** — `a_nat = 168700`, `d_nat = 168696` (matching circomlib's `babyjub.circom`), subgroup order `l`, cofactor 8
- [x] **Curve equation** — `onCurve a d p` defined as `a * p.x^2 + p.y^2 = 1 + d * p.x^2 * p.y^2`
- [x] **Point arithmetic** — `add` (unified Edwards formula), `double`, `scalarMul` (double-and-add). The unified formula is complete on BabyJubJub since `d` is a non-square, so denominators never vanish for valid curve points.
- [x] **Generator points** — `G` (full curve generator, order `8*l`) and `Base8` (`8*G`, prime-order subgroup generator). Both defined over the BN254 scalar field (`ZMod BN254_PRIME`).

### Circuit file (`Clean/Circomlib/BabyJubJub/PointAdd.lean`)

- [x] **ProvableType Point** — manual instance (`size := 2`, `toElements`/`fromElements`) so `Point` can be used as circuit input/output
- [x] **Inputs struct** — `Inputs F` with `p1 p2 : Point F`, deriving `ProvableStruct` (following the `Mux1.lean` pattern)
- [x] **Circuit implementation** — `main` implements the unified Edwards addition formula:
  - 4 intermediate products (`x1x2`, `y1y2`, `x1y2`, `y1x2`)
  - `tau = d * x1x2 * y1y2`, denominators `denom1 = 1 + tau`, `denom2 = 1 - tau`
  - numerators `num1 = x1y2 + y1x2`, `num2 = y1y2 - a * x1x2`
  - Division via the standard zk circuit pattern: witness the inverse with `.ite (denom =? 0) 0 denom⁻¹`, then constrain `x3 * denom1 = num1` (same pattern as `Comparators.IsZero`)
- [x] **All witness generation in the witgen IR** — no `witnessNative`, so the circuit is fully exportable (`#assert_exportable` will pass once proofs are complete)
- [x] **ElaboratedCircuit instance** — via `elaborate_circuit`, needed for subcircuit composition
- [x] **FormalCircuit bundling** — with `Assumptions` (both input points on the curve) and `Spec` (output equals `Point.add`)

## What's still TODO in this PR

- [ ] **`eval_point` lemma** — `ProvableType.eval` on `Point` needs a `circuit_norm` lemma that reduces `eval env p` to `Point.mk (Expression.eval env p.x) (Expression.eval env p.y)`. This is blocked on the framework: `ProvableType.eval` is a `def` that `simp`/`dsimp`/`delta` can't unfold. Without this lemma, `circuit_proof_start` can't decompose struct inputs into component `Expression.eval`s, which means we can't connect the constraint equations (which mention `Expression.eval env input_var.p1.x`) to the spec (which mentions `input.p1.x`).
- [ ] **Soundness proof** — the structure is clear (destructure 13 constraint equations from `<==` and `===`, chain-substitute through the intermediate value definitions, then for each coordinate: `h_check_x` gives `x3 * denom1 = num1`, the goal is `x3 = num1 / denom1`, close with `field_simp` or case-split on denominator zero). Blocked on `eval_point`.
- [ ] **Completeness proof** — follows the standard pattern for circuits using `<==` (witness + constrain simultaneously, so constraints hold by construction). Blocked on `eval_point`.

## What's out of scope (future PRs)

- CurveCheck circuit (point-on-curve validation)
- Scalar multiplication circuit (double-and-add composing PointAdd + Num2Bits)
- EdDSA verification

### DIFF
--- a/Clean.lean
+++ b/Clean.lean
@@ -1,4 +1,6 @@
 import Clean.Circuit
+import Clean.Specs.BabyJubJub
+import Clean.Circomlib.BabyJubJub.PointAdd
 import Clean.Examples.WitnessExport
 import Clean.Utils.FiniteField
 import Clean.Utils.SourceSinkPath
--- a/Clean/Circomlib/BabyJubJub/PointAdd.lean
+++ b/Clean/Circomlib/BabyJubJub/PointAdd.lean
@@ -1,0 +1,96 @@
+/-
+BabyJubJub Point Addition Circuit
+
+Implements the unified Edwards point addition formula from circomlib's BabyAdd.
+Original source: https://github.com/iden3/circomlib/blob/master/circuits/babyjub.circom
+
+Formula:
+  x₃ = (x₁·y₂ + y₁·x₂) / (1 + d·x₁·x₂·y₁·y₂)
+  y₃ = (y₁·y₂ - a·x₁·x₂) / (1 - d·x₁·x₂·y₁·y₂)
+
+Division is handled via witness + multiplication check (same pattern as Comparators.IsZero).
+-/
+import Clean.Circuit
+import Clean.Specs.BabyJubJub
+import Clean.Utils.Tactics.CircuitProofStart
+
+namespace Circomlib.BabyJubJub
+
+open Specs.BabyJubJub (Point a_nat d_nat)
+
+variable {p : ℕ} [Fact p.Prime]
+
+instance : ProvableType Point where
+  size := 2
+  toElements pt := #v[pt.x, pt.y]
+  fromElements v :=
+    let ⟨.mk [x, y], _⟩ := v
+    ⟨x, y⟩
+  fromElements_toElements x := by
+    cases x; rfl
+
+-- TODO: add proper eval lemmas for custom ProvableType Point.
+-- `ProvableType.eval` needs a circuit_norm lemma that reduces `eval env p` to
+-- `Point.mk (Expression.eval env p.x) (Expression.eval env p.y)` so that
+-- `circuit_proof_start` can decompose struct inputs in soundness/completeness proofs.
+@[circuit_norm]
+theorem eval_point {F : Type} [FiniteField F] (env : Environment F) (p : Point (Expression F)) :
+    eval env p = Point.mk (Expression.eval env p.x) (Expression.eval env p.y) := by
+  sorry
+
+structure Inputs (F : Type) where
+  p1 : Point F
+  p2 : Point F
+deriving ProvableStruct
+
+namespace PointAdd
+
+def main (input : Inputs (Expression (F p))) : Circuit (F p) (Point (Expression (F p))) := do
+  let { p1, p2 } := input
+  let a : Expression (F p) := (a_nat : F p)
+  let d : Expression (F p) := (d_nat : F p)
+  let x1x2 <== p1.x * p2.x
+  let y1y2 <== p1.y * p2.y
+  let x1y2 <== p1.x * p2.y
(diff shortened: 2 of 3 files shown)


VanshSahay:
The soundness and completeness proofs are stuck on one thing: connecting `eval env input_var` (from the Soundness hypothesis) to the component `Expression.eval`s that appear in the constraint equations.

For the Inputs struct (which derives ProvableStruct), `circuit_proof_start` successfully decomposes `h_input` into:

```
{ p1 := eval env input_var.p1, p2 := eval env input_var.p2 } = input
```

But then we need one more step, reducing `eval` on `Point` itself. The lemma we need is:

```
eval env (p : Point (Expression F)) =
  Point.mk (Expression.eval env p.x) (Expression.eval env p.y)
```

This should be true by the `ProvableType Point` instance (it just maps `toElements` and `fromElements` through `Expression.eval`), but neither `simp`, `dsimp`, `unfold`, nor `delta` can unfold `ProvableType.eval`. The definition seems opaque to the simplifier.

What's the intended way to handle `eval` decomposition for custom ProvableType types? Is there an existing pattern I'm missing, or does `ProvableType.eval` need to be made reducible? @mitschabaude

mitschabaude:
> What's the intended way to handle `eval` decomposition for custom ProvableType types?

it's intended that this does _not_ automatically decompose by default, because you only need that decomposition in circuits that talk about the individual coordinates. in most high-level gadgets, you won't do that and the proof actually becomes _simpler_ by staying at the level of full, atomic points.

in circuits that do need to talk about coordinates, like `add`, adding a lemma (that's not in circuit_norm) like you proposed is the right way. we have a simp set called `explicit_provable_type` that should help prove that lemma, and also destructing the points should help. so I would expect something like this to close the proof:

```lean
rcases p with \< x, y \>
simp [explicit_provable_type, circuit_norm]
```

Clean.lean
```diff
@@ -1,4 +1,6 @@
 import Clean.Circuit
+import Clean.Specs.BabyJubJub
+import Clean.Circomlib.BabyJubJub.PointAdd
```
mitschabaude, line 3: the spec import can be transitive

```suggestion
import Clean.Circomlib.BabyJubJub.PointAdd
```

also how about moving this down next to the other circomlib imports

Clean/Circomlib/BabyJubJub/PointAdd.lean
```diff
@@ -0,0 +1,96 @@
...
+  sorry
+
+structure Inputs (F : Type) where
+  p1 : Point F
+  p2 : Point F
+deriving ProvableStruct
+
+namespace PointAdd
+
+def main (input : Inputs (Expression (F p))) : Circuit (F p) (Point (Expression (F p))) := do
```
mitschabaude, line 48: would be nice to add the `.circom` source code for the gadget here, like we've been doing in other circomlib gadgets

PR #412, opened 2026-07-03
WASM WAT compiler for witgen IR

VanshSahay:
## Summary

Compiles Clean's witness-generation IR (WitgenIR from PR #403) to WAT (WebAssembly text format). The generated WAT compiles to .wasm with `wat2wasm` and runs in any WASM runtime.

This is the witness-generation side of the Circom-style backend. Stacked on PR #409.

## What's working

**Expression compilers** that walk the IR and emit WAT for all constructors:
- FExpr: const, add, mul, inv, ite, expr (circuit vars + constants), localVar, ofNat
- NExpr: const, val, localVar, add, mul, div, mod, bitwise ops, ite
- BExpr: true, false, feq, lt, neq, not, and

**Full module compilation** (`compileModule`) walks circuit operations and produces a complete WAT module. Handles multi-witness circuits correctly via variable index tracking — witnesses start at offset `numInputs`, avoiding index collisions with inputs.

**Field arithmetic** helpers are auto-generated for any prime modulus. Currently uses i64 (fits primes up to 2^63 like BabyBear).

**Verified end-to-end** with IsZeroField (3 constraints, 2 witnesses):
- Lean generates WAT → wat2wasm compiles to .wasm → Node.js runs it
- in=0 → [0, 1], in=5 → [5^(-1), 0], in=1 → [1, 0] ✓

## Still TODO

- [x] VExpr.mapRange support (loop-based output, needed for SHA256 etc.)
- [x] Let-step support (shared sub-computations)
- [ ] snarkjs ABI compatibility
- [ ] BN254 field support (multi-precision arithmetic for primes > 2^63)

### DIFF
--- a/Clean.lean
+++ b/Clean.lean
@@ -1,5 +1,6 @@
 import Clean.Circuit
 import Clean.Orchard
+import Clean.Backends.Wasm.Compile
 import Clean.Examples.WitnessExport
 import Clean.Utils.FiniteField
 import Clean.Utils.SourceSinkPath
--- a/Clean/Backends/Wasm/Compile.lean
+++ b/Clean/Backends/Wasm/Compile.lean
@@ -11,8 +11,6 @@ open Witgen (FExpr NExpr BExpr VExpr Step WitgenIR)
 
 variable {F : Type} [FiniteField F]
 
-/-! ## Builder -/
-
 structure Builder where
   lines : List String := []
   indent : ℕ := 0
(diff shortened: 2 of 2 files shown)


VanshSahay:
I'm hitting an issue with variable indexing in the witgen IR. When a circuit has multiple witness operations, later operations reference both the original input and earlier witness outputs using the same variable index (`var 0`). For example in IsZero, `x` (the input) and `z` (the first witness) both show up as `expr(var 0)` in the second witness's IR. The compiler can't tell which is which.

Is there a way to distinguish these in the IR, or is this something that needs a framework change?

PR #415, opened 2026-07-03
WASM WAT compiler + R1CS constraint export for snarkjs

VanshSahay:
# WASM backend: binary WASM + R1CS export for snarkjs

Compiles Clean circuits to snarkjs-compatible `.wasm` witness generators (Circom 2 ABI) and `.r1cs` constraint files. Verified end-to-end: `snarkjs wtns calculate`, `snarkjs r1cs info`, and the full Groth16 pipeline (setup → prove → verify).

## Architecture

Two compilers sharing a common flattening pass in `Compile.lean`:

- **WASM compiler** (`Compile.lean` → `Ast.lean` → `Binary.lean`): fully typed WASM AST with a LEB128 binary encoder (type/function/memory/export/code/data/name sections). Compiles witness-generation IR to WASM with the full snarkjs Circom 2 ABI (`init`, `setInputSignal`, `getWitness`, `witness`, ...). No `.raw` instructions.
- **R1CS export** (`R1CS.lean`): constraint extraction shared with the WASM side, serialized both as pretty-printed JSON (`compileR1CS`) and as binary `.r1cs` in the r1csfile format (`compileR1CSBin`, consumed directly by `groth16 setup`; small primes are padded to snarkjs's field-word width).

## Field arithmetic

- **Single-word (primes ≤ 2³²)**: i64 ops with `i64.mul`/`i64.rem_u`; larger primes with `numWords = 1` are rejected with an error.
- **Multi-word (e.g. BN254)**: full multi-precision arithmetic in Montgomery form (`x·R mod p`, `R = 2^(N·64)`):
  - `$mul64x64`: 64×64→128 with carry detection
  - `$fmul`: N×N schoolbook + CIOS Montgomery reduction (HAC Algorithm 14.36)
  - `$fadd`: limb-wise addition with two-operand carry checks + conditional mod p
  - `$finv`: Fermat square-and-multiply via `$fmul`
  - `ite`, `flt`, `feq`, `bit`, `bitsOf`, `listGet` with multi-word support; intermediate signals, let-steps, and a shared scratch region; outputs-first signal layout

## API

```lean
compileModule  fieldPrime numInputs (inputNames := []) (outputVarIdx := []) ops numWords : Except String ByteArray
compileR1CS    fieldPrime numInputs (inputNames := []) (outputVarIdx := []) ops numWords : Except String String
compileR1CSBin fieldPrime numInputs (inputNames := []) (outputVarIdx := []) ops numWords : Except String ByteArray
```

- `inputNames` (one per input): strict FNV-1a input-key validation like circom — unknown keys are rejected ("Signal not found"). Empty = lenient mode.
- `outputVarIdx`: outputs-first signal layout (`1, outputs…, inputs…, witnesses…`) so `groth16 public.json` contains the circuit's real outputs and inputs.
- Everything returns `Except String`; unsupported constructs fail at compile time with a message — no silent fallbacks.

## Verification

- **33 `CleanTests` checks pass**, including: Poseidon1 compiled to WASM, run through `snarkjs wtns calculate`, output matched against Lean ground truth (`Specs.PoseidonOptimized.poseidon1Opt`) on 3 inputs; every module passes `wasm-validate`; binary `.r1cs` validated by `snarkjs r1cs info`; negative tests (native witnesses, lookups, `numWords = 1` on BN254, unknown input keys) all rejected.
- **Full Groth16 pipeline on Poseidon1** (BN254, 620 signals / 618 constraints): `powersoftau` → `groth16 setup` → `zkey export verificationkey` → `groth16 prove` → `groth16 verify` = **OK**, with `public.json = [poseidon1(0), input]` correct under the outputs-first layout.
- **Independent audit**: a 5-lens audit (codegen arithmetic, binary encoding, R1CS, end-to-end ABI, reviewer hygiene) with adversarial verification of every finding surfaced 14 issues. All fixed, each with a regression test, including: `$fadd` carry-corner bug, shared-scratch sizing, multi-word `feq`/`flt` codegen, `listGet` index and Montgomery-form bugs, R1CS signal numbering sync, strict input names, outputs-first layout, the `witness` ABI signature, and the data section.

## Verified circuits

| Circuit | Kind | Where |
|---|---|---|
| Poseidon1 | full packaged `FormalCircuit` with soundness/completeness proofs (BN254, 620 signals, 618 constraints) | tests + `Clean/Examples/WasmDemo.lean` |
| mulAdd + ~12 small circuits (add, assert, let-steps, flt, bit, bitsOf, envRange, append, listGet, val, feq, nested flt) | hand-written witness IR covering the feature matrix | `Clean/Utils/Test/TestWasmCompile.lean` |

## Key changes in this PR

- Montgomery CIOS reduction (replacing Barrett), two-operand carry detection in `$fadd`
- Strict input-name validation (`inputNames`) and outputs-first signal layout (`outputVarIdx`)
- Binary `.r1cs` export (`compileR1CSBin`) in the r1csfile format
- Data section emitted (constant signal initialized statically); `witness` export with the correct zero-arg ABI signature
- Typed WASM AST with zero `.raw` instructions; dynamic memory sizing; LEB128 binary encoder
- 14 audit fixes with regression tests; `WasmDemo.lean` walkthrough; README rewrite

## Known limitations (documented in the README)

- The compiler is tested against Lean ground truth — the codegen itself is not formally verified.
- `init(sanityCheck)` is accepted but ignored; nested `listGet` is rejected at compile time; `native` witnesses and `dataGet`/`hintGet` are not representable; lookups/interactions are rejected by R1CS export.
- Keccak-sized multi-word circuits can exceed WASM's 50,000-locals-per-function limit (documented).
- Performance has not yet been benchmarked against circom's generated WASM.

Progress on #420.


VanshSahay:
Keccak256 permutation (Keccak-f[1600]) now compiles and runs end to end:

- 31113 signals (200 inputs, 30912 witness cells)
- 657K lines of WAT, 1.7MB WASM binary
- 25401 non-zero witness signals, 186/200 output signals non-zero
- Full snarkjs ABI works: init, setInputSignal, getWitness, getWitnessSize, etc.

Keccak-f[1600] of the all-zero state produces the expected non-zero permutation output

VanshSahay:
On the BN254 Fact issue, the suggestion to use #eval! doesn't quite work here, here's why:

The chain is: compileModule needs FiniteField (F BN254_PRIME), which needs Fact (Nat.Prime BN254_PRIME), which is currently sorry in Circomlib/Poseidon.lean. #eval! only lets you call unsafe definitions, it can't skip past sorry in typeclass proofs

Three ways to unblock this:

1. just use native_decide anyway. It takes 8+ minutes for a 254-bit number but circuit export is a one-time build step, not a hot loop. Could be fine for now but im not in favour of this

2. add a compileModule variant that works with plain ℕ, the arithmetic generators (genFmul, genFinv etc.) already take (p numWords : ℕ) and return Func, they don't need FiniteField at all. The type parameter F is only there for type safety in the expression compilers. A ℕ-based path would work today and avoid the Fact dependency entirely for code generation

3. Wait for CompPoly in Lean 4.30, which has proper primality proofs for large primes and will make native_decide unnecessary like you suggested

Option 2 seems like the lowest friction, it's basically just threading ℕ through the places that currently need F, and the actual arithmetic code doesn't care about the type parameter

mitschabaude:
the bump to Lean 4.30 was merged today, so option 3 after finding a compatible recent comppoly version should be a low lift now. (latest version is already 4.31, but we can pin an earlier one)

no need to wait for anything at this point

Clean/Specs/Poseidon.lean
```diff
@@ -7,12 +7,19 @@ https://github.com/iden3/circomlib/blob/master/circuits/poseidon.circom
...
 namespace Specs.Poseidon
 
 -- BN254 scalar field prime (same as used in circomlib/snarkjs)
 def BN254_PRIME : ℕ := 21888242871839275222246405745257275088548364400416034343698204186575808495617
 
+-- Proven prime via CompPoly's Pratt certificate
+instance : Fact (Nat.Prime BN254_PRIME) := ⟨by
+  have h : BN254_PRIME = BN254.scalarFieldSize := by native_decide
+  rw [h]
+  exact BN254.ScalarField_is_prime⟩
```
mitschabaude: fwiw `native_decide` is mostly frowed upon since it delegates more work to the kernel. here it's completely unnecessary, the two definitions are defeq and the proof can be `rfl`.

try to avoid `native_decide`!

Clean/Circomlib/Poseidon.lean
```diff
@@ -18,8 +18,8 @@ namespace Circomlib.Poseidon
 open Specs.Poseidon (F BN254_PRIME C_t2 M_t2)
 open Specs.PoseidonOptimized (P_t2 S_t2)
 
--- BN254 prime facts (BN254_PRIME is a well-known prime, proofs omitted for performance)
-instance : Fact (Nat.Prime BN254_PRIME) := ⟨by sorry⟩
+-- BN254 scalar field prime, proven via CompPoly Pratt certificate (see Specs.Poseidon)
+instance : Fact (Nat.Prime BN254_PRIME) := inferInstance
```
mitschabaude: you can delete this. if `inferInstance` proves it, that means you don't need that instance (it can already be inferred without it)

Clean/Backends/Wasm/Binary.lean
```diff
@@ -0,0 +1,166 @@
...
+  | .relop .i64 .eqz => arr.push 0x50
+  | .relop _ _ => arr
+  | .localGet idx => putULEB128 (arr.push 0x20) idx
+  | .localSet idx => putULEB128 (arr.push 0x21) idx
+  | .localTee idx => putULEB128 (arr.push 0x22) idx
+  | .call name => putULEB128 (arr.push 0x10) (resolveCall name)
+  | .br _ => putULEB128 (arr.push 0x0C) 0
+  | .brIf _ => putULEB128 (arr.push 0x0D) 0
+  | .block _ _ body => encodeBlock arr resolveCall 0x02 body
+  | .loop _ _ body => encodeBlock arr resolveCall 0x03 body
```
mitschabaude: it seems wrong to model blocks and loops with a result type, but when encoding, ignoring the result type and writing an empty result type instead. decide for one of either

* we only model blocks without result for now
* we correctly encode all result types

Clean/Backends/Wasm/Binary.lean
```diff
@@ -0,0 +1,166 @@
...
+  | .relop .i32 .lt_u => arr.push 0x49
+  | .relop .i64 .lt_u => arr.push 0x65
+  | .relop .i64 .eqz => arr.push 0x50
+  | .relop _ _ => arr
+  | .localGet idx => putULEB128 (arr.push 0x20) idx
+  | .localSet idx => putULEB128 (arr.push 0x21) idx
+  | .localTee idx => putULEB128 (arr.push 0x22) idx
+  | .call name => putULEB128 (arr.push 0x10) (resolveCall name)
+  | .br _ => putULEB128 (arr.push 0x0C) 0
+  | .brIf _ => putULEB128 (arr.push 0x0D) 0
```
mitschabaude: wait, shouldn't the emitted instruction have something to do with the given label you `break` to? here you ignore the label, does that mean you hard-code breaking to the innermost parent scope? if yes then you should not make the blocks get a label at all. or (better), properly implement translation of labels into numbers

Clean/Backends/Wasm/Binary.lean
```diff
@@ -0,0 +1,166 @@
...
+  | .block _ _ body => encodeBlock arr resolveCall 0x02 body
+  | .loop _ _ body => encodeBlock arr resolveCall 0x03 body
+  | .ifElse result thenBody elseBody =>
+    let arr := arr.push 0x04
+    let arr := match result with
+      | [] => arr.push 0x40  -- empty block type
+      | [t] => arr.push (vtOpc t)  -- single result type
+      | _ => arr.push 0x40  -- multi-value: placeholder (needs type section ref)
```
mitschabaude: this is another placeholder with currently incorrect behavior.. again, the best way to handle missing support is to not model multivalue returns in the type for now

Clean/Backends/Wasm/Binary.lean
```diff
@@ -0,0 +1,166 @@
...
+    let arr := thenBody.foldl (fun a i => encodeInstr a resolveCall i) arr
+    let arr := if elseBody.isEmpty then arr else
+      (arr.push 0x05) |> fun a => elseBody.foldl (fun a' i => encodeInstr a' resolveCall i) a
+    arr.push 0x0B
+  | .memLoad .i32 off _ => putULEB128 (putULEB128 (arr.push 0x28) 2) off
+  | .memLoad .i64 off _ => putULEB128 (putULEB128 (arr.push 0x29) 3) off
+  | .memStore .i32 off _ => putULEB128 (putULEB128 (arr.push 0x36) 2) off
+  | .memStore .i64 off _ => putULEB128 (putULEB128 (arr.push 0x37) 3) off
```
mitschabaude: you don't use the `align` value, looks incorrect to me?

Clean/Backends/Wasm/Compile.lean
```diff
@@ -0,0 +1,893 @@
...
+  else List.range nw |>.foldl (fun cb' w => cb'.push (i64.const ((val >>> (w * 64)) % (2^64)))) cb
+
+def pushVar (idx : ℕ) (vm : VarMap) (cb : CodeBuilder) : CodeBuilder :=
+  let nw := vm.numWords
+  let base := vm.lookup idx
+  if nw = 1 then cb.push (local.get base)
+  else List.range nw |>.foldl (fun cb' w => cb'.push (local.get (base + w))) cb
+
+mutual
+partial def compileFExpr (vm : VarMap) : FExpr F → CodeBuilder → CodeBuilder
```
mitschabaude: would be nice to get rid of the `partial`s here so that you prove the recursion terminates

Clean/Backends/Wasm/Compile.lean
```diff
@@ -0,0 +1,893 @@
...
+  | .localVar i, cb => cb.push (local.get (vm.lookup (vm.letBase + i)))
+  | .add a e, cb => let cb := compileNExpr vm a cb; let cb := compileNExpr vm e cb; cb.push i64.add
+  | .mul a e, cb => let cb := compileNExpr vm a cb; let cb := compileNExpr vm e cb; cb.push i64.mul
+  | .div a e, cb => let cb := compileNExpr vm a cb; let cb := compileNExpr vm e cb; cb.push (.binop .i64 .div_u)
+  | .mod a e, cb => let cb := compileNExpr vm a cb; let cb := compileNExpr vm e cb; cb.push i64.rem_u
+  | .land a e, cb => let cb := compileNExpr vm a cb; let cb := compileNExpr vm e cb; cb.push i64.and
+  | .lor a e, cb => let cb := compileNExpr vm a cb; let cb := compileNExpr vm e cb; cb.push i64.or
+  | .lxor a e, cb => let cb := compileNExpr vm a cb; let cb := compileNExpr vm e cb; cb.push (.binop .i64 .xor)
+  | .shiftL a e, cb => let cb := compileNExpr vm a cb; let cb := compileNExpr vm e cb; cb.push i64.shl
+  | .shiftR a e, cb => let cb := compileNExpr vm a cb; let cb := compileNExpr vm e cb; cb.push i64.shr_u
```
mitschabaude: this is wrong in that it treats natural numbers as 64-bit values. instead, they should be expected to be at least as large as any field element, since NExprs are often generated from FExprs via .val.

in some sense, this exposes an issue the witnessIR has not solved well in its design. it has no way of saying "I know this will be a small number". without something like this existing, it should arguably assume naturals are the same size as field elements, otherwise it will be broken in some examples.

Clean/Backends/Wasm/Compile.lean
```diff
@@ -0,0 +1,893 @@
...
+  match a, b with
+  | [], _ => b
+  | _, [] => a
+  | (i1, c1) :: xs, (i2, c2) :: ys =>
+    if i1 < i2 then (i1, c1) :: addLinCombs xs ((i2, c2) :: ys) p
+    else if i1 = i2 then (i1, (c1 + c2) % p) :: addLinCombs xs ys p
+    else (i2, c2) :: addLinCombs ((i1, c1) :: xs) ys p
+
+open Expression (var const add mul) in
+partial def flattenExpr (p : ℕ) (vm : VarMap) : Expression F → FlattenState → (LinComb × FlattenState)
```
mitschabaude: get rid of `partial`

Clean/Backends/Wasm/Compile.lean
```diff
@@ -0,0 +1,893 @@
...
+  | .var i, st => ([(1 + vm.lookup i.index / vm.numWords, 1)], st)  -- R1CS signal = 1 + field element index
+  | .const c, st =>
+    let val := FiniteField.val c % p
```
mitschabaude: why `% p` here, c is already a field element so this is wrong

Clean/Backends/Wasm/Compile.lean
```diff
@@ -0,0 +1,893 @@
...
+  | .feq a e, cb => let cb := compileFExpr vm a cb; let cb := compileFExpr vm e cb; cb.push i64.eq
+  | .lt a e, cb => let cb := compileNExpr vm a cb; let cb := compileNExpr vm e cb; cb.push i64.lt_u
+  | .neq a e, cb => let cb := compileNExpr vm a cb; let cb := compileNExpr vm e cb; cb.push i64.eq
+  | .not x, cb => let cb := compileBExpr vm x cb; cb.push i64.eqz
+  | .and a e, cb => let cb := compileBExpr vm a cb; let cb := compileBExpr vm e cb; cb.push i64.and
+end
+
+/-! ## Expression flattening (shared by WASM and R1CS compilers) -/
+
+abbrev LinComb := List (ℕ × ℕ)  -- sparse (signalIndex × coefficient) pairs
```
mitschabaude: coefficients are field elements, the type erases them to arbitrary natural numbers, which is wrong and makes you add extra `% p` below

Clean/Backends/Wasm/Compile.lean
```diff
@@ -0,0 +1,893 @@
...
+def scaleLinComb (c : ℕ) (lc : LinComb) (p : ℕ) : LinComb :=
+  lc.map fun (i, coeff) => (i, (c * coeff) % p)
+
+def addLinCombs (a b : LinComb) (p : ℕ) : LinComb :=
+  match a, b with
+  | [], _ => b
+  | _, [] => a
+  | (i1, c1) :: xs, (i2, c2) :: ys =>
+    if i1 < i2 then (i1, c1) :: addLinCombs xs ((i2, c2) :: ys) p
+    else if i1 = i2 then (i1, (c1 + c2) % p) :: addLinCombs xs ys p
```
mitschabaude: this should take to field elements `c1` and `c2` and add them using the `Field` instance, not hand-roll field addition inline

Clean/Backends/Wasm/Compile.lean
```diff
@@ -0,0 +1,893 @@
...
+
+structure FlattenState where
+  nextSignal : ℕ := 1
+  constraints : List Constraint := []
+
+def isConstant (lc : LinComb) : Bool :=
+  match lc with | [(0, _)] => true | _ => false
+
+def scaleLinComb (c : ℕ) (lc : LinComb) (p : ℕ) : LinComb :=
+  lc.map fun (i, coeff) => (i, (c * coeff) % p)
```
mitschabaude: this should do multiplication in the `Field` instance, not hand-roll finite field multiplication

Clean/Backends/Wasm/Compile.lean
```diff
@@ -0,0 +1,892 @@
...
+  -- Step 3: q1 = c[3..7] (c >> 192, 5 limbs)
+  let extractQ1 : List Instr :=
+    (List.range (N+1)) >>= fun i => [ local.get (cBase + 3 + i), local.set (q1Base + i) ]
+
+  -- Step 4: q2 = q1 * μ (5×5 schoolbook → 9 limbs, scratch at 12-15)
+  let muMult := genSchoolbook (N+1) q1Base muBase q2Base scratchN5
+
+  -- Step 5: q3 = q2[5..8] (q2 >> 320, 4 limbs), reuse q1Base[0..3]
+  let extractQ3 : List Instr :=
+    (List.range N) >>= fun i => [ local.get (q2Base + 5 + i), local.set (q1Base + i) ]
```
mitschabaude: this has many comments suggesting hard-coded numbers of limbs and constant number such as 5. it makes me wonder if the function actually works for any `N`? if not it should not pretend it does by taking `N` as a param

Clean/Backends/Wasm/Compile.lean
```diff
@@ -0,0 +1,892 @@
...
+    results := List.replicate N .i64
+    locals := (List.range N).map fun i => (s!"$r{i}", .i64)
+    body := init ++ steps ++ finvRets }
+
+/-- Generate multi-word arithmetic as AST Func list. -/
+def genMultiWordArith (p numWords : ℕ) : List Func :=
+  [ genMul64x64, genFmul p numWords, genFadd p numWords, genFsub numWords, genFinv p numWords ]
+
+structure VarMap where
+  env : List (ℕ × ℕ) := []
```
mitschabaude, line 453: what does this N \times N list mean? would be good to document these data structures

Clean/Backends/Wasm/Compile.lean
```diff
@@ -0,0 +1,892 @@
...
+  letBase : ℕ := 0
+  numWords : ℕ := 1
+
+def VarMap.init (numInputs : ℕ) (numWords : ℕ := 1) : VarMap :=
+  { env := List.range numInputs |>.map fun i => (i, i * numWords)
+    nextLocal := numInputs * numWords
+    numWords }
+
+def VarMap.lookup (vm : VarMap) (idx : ℕ) : ℕ :=
+  match vm.env.find? fun (i, _) => i = idx with | some (_, w) => w | none => idx * vm.numWords
```
mitschabaude, line 485: what does this do? why does this fall back to `idx`?

Clean/Backends/Wasm/Compile.lean
```diff
@@ -0,0 +1,892 @@
...
+  | [], _ => b
+  | _, [] => a
+  | (i1, c1) :: xs, (i2, c2) :: ys =>
+    if i1 < i2 then (i1, c1) :: addLinCombs xs ((i2, c2) :: ys) p
+    else if i1 = i2 then (i1, (c1 + c2) % p) :: addLinCombs xs ys p
+    else (i2, c2) :: addLinCombs ((i1, c1) :: xs) ys p
+
+open Expression (var const add mul) in
+partial def flattenExpr (p : ℕ) (vm : VarMap) : Expression F → FlattenState → (LinComb × FlattenState)
+  | .var i, st => ([(1 + vm.lookup i.index / vm.numWords, 1)], st)  -- R1CS signal = 1 + field element index
```
mitschabaude: why can we divide by `numWords` here? is there an implicit invariant that the number is divisible by that? if yes why don't we store the index directly?

Clean/Backends/Wasm/Compile.lean
```diff
@@ -0,0 +1,892 @@
...
+  | .localVar i, cb => pushVar (vm.letBase + i) vm cb
+  | .envGet _, cb => cb.push (i64.const 0)
+  | .listGet _ _, cb => cb.push (i64.const 0)
+  | .dataGet _ _ _ _, cb => cb.push (i64.const 0)
+  | .hintGet _ _ _ _, cb => cb.push (i64.const 0)
+
+partial def compileNExpr (vm : VarMap) : NExpr F → CodeBuilder → CodeBuilder
+  | .const n, cb => cb.push (i64.const n)
+  | .val x, cb => compileFExpr vm x cb
+  | .idx, cb => match vm.loopIdx with | some _ => cb.push (local.get 0) | none => cb.push (i64.const 0)
```
mitschabaude: why does the `loopIdx` get mapped to `local.get 0`?

Clean/Backends/Wasm/Compile.lean
```diff
@@ -0,0 +1,892 @@
...
+      ({ vmOut with loopIdx := none }, vi + n, instrs)
+  | _, .append _ _ => (vm, vi, acc)
+
+/-- process flat operations, accumulating instructions. -/
+def processFlatOps (numInputs : ℕ) : List (FlatOperation F) → VarMap → ℕ → List Instr → VarMap × ℕ × List Instr
+  | [], vm, _, instrs => (vm, numInputs, instrs)
+  | .witness _ (.ir steps vexpr) :: rest, vm, vi, acc =>
+    let vmStep := { vm with letBase := vi }
+    let (vmS, viS, stepInstrs) := compileSteps vmStep vi steps
+    let (vmOut, viOut, outInstrs) := compileVExpr vmS viS stepInstrs vexpr
```
mitschabaude: any reason the `letBase` is threaded into these function explicitly? why don't they use letBase directly?

Clean/Backends/Wasm/Compile.lean
```diff
@@ -0,0 +1,892 @@
...
+    let (vm', locs) := vm.alloc 1 vi
+    (vm', vi + 1, instrs ++ cb.build ++ locs.reverse.map fun idx => local.set idx)
+  ) (vm, vi, acc)
+
+/-- compile a VExpr to instructions. -/
+def compileVExpr (vm : VarMap) (vi : ℕ) (acc : List Instr) : {m : ℕ} → VExpr F m → VarMap × ℕ × List Instr
+  | _, .lit es => compileLit vm vi acc es.toList
+  | _, .mapRange n body =>
+    match body with
+    | .envGet _ => (vm, vi, acc)
```
mitschabaude: why doesn't result an `.envGet` step in any instruction? shouldn't that load the variable from where the previous index was stored, and store it in the new location?

Clean/Backends/Wasm/Compile.lean
```diff
@@ -0,0 +1,892 @@
...
+  | .assert e => [.assert e]
+  | .lookup l => [.lookup l]
+  | .interact i => [.interact i]
+  | .subcircuit s => s.ops.toFlat
+
+/-- Flatten a list of Operations. -/
+def flattenOps (ops : List (Operation F)) : List (FlatOperation F) :=
+  match ops with
+  | [] => []
+  | op :: rest => flattenOp op ++ flattenOps rest
```
mitschabaude: this method already exists, I think it's called `ops.toFlat`, delete the duplicate

Clean/Backends/Wasm/Compile.lean
```diff
@@ -0,0 +1,892 @@
...
+def compileModule (fieldPrime numInputs : ℕ) (ops : List (Operation F)) (numWords : ℕ := 1) : String :=
+  let nw := numWords
+  let vm := VarMap.init numInputs nw
+  let flatOps := flattenOps ops
+  -- vi starts at numInputs so that circuit variable indices (which start at 0 for
+  -- inputs) align with VarMap entries. vm.alloc adds (vi, local) for each witness,
+  -- and pushVar uses the circuit variable index from the witness IR directly.
+  let (finalVm, _, bodyInstrs) := processFlatOps numInputs flatOps vm numInputs []
+  let witnessWords := finalVm.nextLocal - numInputs * nw
+  let witnessCount := witnessWords / nw
```
mitschabaude: it seems really weird to operate on implicit multiples of `nw` everywhere instead of on the factors directly

Clean/Backends/Wasm/Compile.lean
```diff
@@ -0,0 +1,892 @@
...
+  let signalBaseRaw := 4 + n32 * 4
+  let signalBase := ((signalBaseRaw + 7) / 8) * 8
+  let signalBytes := n32 * 4
+  let startSignal := 1 + finalVm.nextLocal / nw
+  -- Local index base for intermediates in getWitness: param $i(0), $tmp(1), $idx(2), $in_*(3..)
+  -- For multi-word, each input has nw limbs; locals are $in_{i}_{w}
+  -- Each intermediate uses nw consecutive locals
+  let intLocalBase := 3 + numInputs * nw
+  let (numInt, intLocals, intCode) :=
+    discoverAndCompileIntermediates fieldPrime vm flatOps startSignal signalBase signalBytes nw intLocalBase
```
mitschabaude: what is happening here? why do we compile `.assert` operations into code, what does it have to do with witness generation?

Clean/Backends/Wasm/Compile.lean
```diff
@@ -0,0 +1,892 @@
...
+    let elseCB := compileFExpr vm e {}
+    -- compileBExpr pushes i64; ifElse expects i32 condition
+    let results := List.replicate nw ValType.i64
+    cb.push (.unop .i32 .wrap_i64) |>.push (.ifElse results thenCB.build elseCB.build)
+  | .ofNat n, cb => compileNExpr vm n cb
+  | .localVar i, cb => pushVar (vm.letBase + i) vm cb
+  | .envGet _, cb => cb.push (i64.const 0)
+  | .listGet _ _, cb => cb.push (i64.const 0)
+  | .dataGet _ _ _ _, cb => cb.push (i64.const 0)
+  | .hintGet _ _ _ _, cb => cb.push (i64.const 0)
```
mitschabaude: this is all wrong, these instructions seem to be not supported atm

in the PR description you claimed
> Handles all FExpr/NExpr/BExpr constructors including ite with multi-word support.

if you push code that is incomplete you should highlight that clearly

Clean/Backends/Wasm/Compile.lean
```diff
@@ -0,0 +1,892 @@
...
+  | .expr (.add a e), cb => let cb := compileFExpr vm (.expr a) cb; let cb := compileFExpr vm (.expr e) cb; cb.push (call "$fadd")
+  | .expr (.mul a e), cb => let cb := compileFExpr vm (.expr a) cb; let cb := compileFExpr vm (.expr e) cb; cb.push (call "$fmul")
+  | .ite c t e, cb =>
+    let nw := vm.numWords
+    let cb := compileBExpr vm c cb
+    let thenCB := compileFExpr vm t {}
+    let elseCB := compileFExpr vm e {}
+    -- compileBExpr pushes i64; ifElse expects i32 condition
+    let results := List.replicate nw ValType.i64
+    cb.push (.unop .i32 .wrap_i64) |>.push (.ifElse results thenCB.build elseCB.build)
```
mitschabaude: `i32` is the native way to represent a boolean in WASM, so you should change `compileBExpr` to target `i32` not `i64`

Clean/Backends/Wasm/Compile.lean
```diff
@@ -0,0 +1,892 @@
...
+  | .subcircuit s => s.ops.toFlat
+
+/-- Flatten a list of Operations. -/
+def flattenOps (ops : List (Operation F)) : List (FlatOperation F) :=
+  match ops with
+  | [] => []
+  | op :: rest => flattenOp op ++ flattenOps rest
+
+/-- Compile to a WASM Module. This is the main entry point. -/
+def compileModule (fieldPrime numInputs : ℕ) (ops : List (Operation F)) (numWords : ℕ := 1) : String :=
```
mitschabaude: it seems very easy to accidentally call this with the `numWords = 1` default for any fieldPrime, which is a footgun

Clean/Backends/Wasm/Compile.lean
```diff
@@ -0,0 +1,892 @@
...
+/-- Generate single-word field arithmetic functions. -/
+def genSingleWordArith (p : ℕ) : List Func :=
+  let pVal : ℕ := p
+  let pm2 : ℕ := p - 2
+  [
+    { name := "$fadd"
+      params := [("", .i64), ("", .i64)]
+      results := [.i64]
+      body := [.localGet 0, .localGet 1, .binop .i64 .add,
+               .const .i64 pVal, .binop .i64 .rem_u] }
```
mitschabaude: would be nicer to use the prettier instruction API you introduced just above this method

Clean/Backends/Wasm/Compile.lean
```diff
@@ -0,0 +1,892 @@
...
+    ,
+    { name := "$fmul"
+      params := [("", .i64), ("", .i64)]
+      results := [.i64]
+      body := [.localGet 0, .localGet 1, .binop .i64 .mul,
+               .const .i64 pVal, .binop .i64 .rem_u] }
```
mitschabaude: this method is incorrect for primes of bit size between 32 and 64. the multiplication of two 64-bit numbers will overflow the uint64 range before being reduced modulo the prime

Clean/Backends/Wasm/Compile.lean
```diff
@@ -0,0 +1,892 @@
...
+  let memNeeded := signalBase + totalSignals * signalBytes
+  let memPages := (memNeeded + 65535) / 65536  -- ceil division
+  let module : Ast.Module := {
+    memoryPages := memPages
+    dataSegments := [(signalBase, signalInit)]
+    funcs := arithFuncs ++ [computeFunc,
+      { computeFunc with name := "$witness", exportName := some "witness" }]
+      ++ abiFuncs
+  }
+  Module.toString module
```
mitschabaude: can you compile to binary instead? it's the much nicer target IMO

Clean/Backends/Wasm/R1CS.lean
```diff
@@ -0,0 +1,72 @@
...
+def processOps (p : ℕ) (vm : VarMap) (ops : List (FlatOperation F)) (st : FlattenState) :
+    List Constraint × ℕ :=
+  match ops with
+  | [] => (st.constraints, st.nextSignal)
+  | .witness _ _ :: rest =>
+    processOps p vm rest st  -- VarMap already handles witness allocation
+  | .assert e :: rest =>
+    let (lc, st1) := flattenExpr p vm e st
+    let constr : Constraint := (lc, [(0, 1)], [])
+    let st2 := { st1 with constraints := constr :: st1.constraints }
```
mitschabaude: the way we generate constraints is pretty inefficient right now. for example, any R1CS-shaped constraint, even one as simple as `x * y === z`, will cause you to generate two constraints

mitschabaude (changes requested):
this is really promising work!!

at a high-level, the following is missing for me:

* there seem to be several operations that are not supported or cases that don't work. that's perfectly fine, but the current implementation fails silently in each such case, which I don't find acceptable: it's actually quite hard to tell from the code _what exactly_ will work. I want the compiler to return an error when processing inputs that are not supported. the acceptance criteria is that the code either fails with a clear reason or produces something correct.
* you should add documentation, for example a README.md inside Backends/Wasm. it should document how to run the compiler and produce the R1CS and WASM files, and how to use them with snarkjs. there should also be references to the snarkjs ABI you're targeting

mitschabaude:
@VanshSahay one feedback I forgot in https://github.com/Verified-zkEVM/clean/pull/415#pullrequestreview-4703847508 is that there should be tests documenting both working and failing input circuits

Clean/Backends/Wasm/Compile.lean
mitschabaude: once that works, can you also remove the redundant WAT compiler? WASM -> WAT is probably not needed or can be done with separate tooling

rot256:
Hi, this PR is extremely cool, thanks a lot!

I'm using it as the example for Clean tutorial (work in progress at https://clean-intro.zksec.workers.dev/) 

However, I/Codex, uncovered some performance issues: https://gist.github.com/rot256/082e5f8311741da2b0be828329225157

TLDR: appending to an immutable list inside `processFlatOps` creates a new list with all the elements copied, this made the WitGen DSL lowering quadratic time.

VanshSahay:
Hey @rot256 , thanks for the report. I've fixed the WASM-side exponential flattening

one finding of yours is still open and it's the same acc ++ chunk pattern, `Circuit.bind` accumulates operations with O(k²) copying, that's this issue. It affects every backend, and the current non-unpacking shape is load-bearing for soundness/completeness proof performance, I've opened an [issue](https://github.com/Verified-zkEVM/clean/issues/456) for it, happy to open a seperate PR for this as well!

kevin-zks:
Thanks for this backend! With Gregor’s approval, I pushed a small corrective follow-up in [316d88ae](https://github.com/Verified-zkEVM/clean/commit/316d88ae82407732cf083377b14377707c607efe), preserving the existing structure and R1CS special cases:

- Correct scalar-product assertion signs and align intermediate numbering with the original expression traversal.
- Preserve dependency order when emitting intermediate computations.
- Keep bit-decomposition outputs in Montgomery form and account for high carries during field reduction.
- Return zero for empty table reads and preserve Lean’s total UInt64 division/remainder semantics.

Added focused positive conformance tests. Local checks passed: `lake build --wfail`, `lake build CleanTests`, `lake build Clean.Examples.WasmDemo`, `wasm-validate`, and snarkjs witness generation plus `wtns check` for the documented outputs-first Poseidon demo (input 5). Formatting checks passed too; all 34 existing WASM test blocks and three new blocks ran without missing-tool skips.

These tests validate the corrected implementation; I’m not claiming every original issue was dynamically reproduced. Coverage is not exhaustive, and I did not rerun the full Groth16 pipeline. Existing unrelated `sorry` warnings in `TestCircuitProofStart.lean` remain.

Clean/Backends/Wasm/README.md
```diff
@@ -0,0 +1,298 @@
+# Clean WASM Backend
+
+A compiler written in Lean that turns [Clean](https://github.com/VanshSahay/clean) circuits into standalone artifacts the [snarkjs](https://github.com/iden3/snarkjs) toolchain can consume directly:
```
mitschabaude: Fix link to clean

Clean/Backends/Circom/README.md
mitschabaude, line 1: Move all of Backends/Wasm to Backends/Circom

Clean/Backends/Circom/TestWasmCompile.lean
mitschabaude, line 1: Move new tests to Backends/Circom

lake-manifest.json
```diff
@@ -1,7 +1,17 @@
...
  "packages":
- [{"url": "https://github.com/leanprover-community/mathlib4",
+ [{"url": "https://github.com/Verified-zkEVM/CompPoly.git",
+   "type": "git",
+   "subDir": null,
+   "scope": "",
+   "rev": "2b10cc36514ec769a6d6b2abb51ce436a4005085",
+   "name": "CompPoly",
+   "manifestFile": "lake-manifest.json",
+   "inputRev": "v4.32.0",
```
mitschabaude: This will need updating to 4.33.1 probably

lakefile.lean
```diff
@@ -14,3 +14,4 @@ lean_lib CleanTests where
   roots := #[`Clean.Test, `Clean.Specs.BLAKE3.ChunkProcessingTests]
 
 require mathlib from git "https://github.com/leanprover-community/mathlib4"@"v4.32.2"
+require CompPoly from git "https://github.com/Verified-zkEVM/CompPoly.git"@"v4.32.0"
```
mitschabaude: Will need updating

mitschabaude (approved):
Approving modulo addressing final comments and adapting to main

mitschabaude:
thanks a lot for this PR @VanshSahay

PR #421, opened 2026-07-05
R1CS constraint export for snarkjs

VanshSahay:
Adds R1CS constraint generation from Clean circuit operations. Generates JSON compatible with snarkjs for Groth16 prove/verify.

## How it works

Walks each `assert expr` and recursively flattens the expression tree to degree-2 R1CS constraints by introducing intermediate signals for non-constant multiplications. Signal 0 is constant 1, signals 1..numInputs are inputs, the rest are witnesses and intermediates.

## Current status

Draft. Generates valid R1CS JSON. Still needs Clean's internal `a + (-1)*b` encoding normalization to match circom-generated R1CS exactly.


VanshSahay:
Merged into #415 — the R1CS constraint export is now part of the unified WASM + R1CS PR.

PR #424, opened 2026-07-06
Bump Lean to 4.31

mitschabaude:
Bumps Clean from Lean **4.30** to **4.31** (mathlib v4.31.0), on top of #370. Motivation: the library we want to import has moved to 4.31, and 4.31 fixes the match-auxiliary blockage behind several of our 4.30 workarounds.

**Status: fully green** — `lake build` (1811 jobs) and `lake build CleanTests` both pass with zero failures.

## What changed in 4.31 (relevant to Clean)

The transparency overhaul continues, but this time mostly in our favor:

- **leanprover/lean4#13281 + #13363**: the match-auxiliary blockage is fixed structurally — exposed match auxiliaries are `[implicit_reducible]`, and `whnfMatcher`'s transparency bump is replaced by an explicit allowlist (`canUnfoldAtMatcher`). This is what the 4.30 PR predicted would let us drop scoped legacy-defeq workarounds. **The flip side**: the allowlist does *not* include user operator instances, so match-based evaluators (`Expression.eval`, witgen-IR `FExpr.eval`/`NExpr.eval`) no longer see through composite operator instances like `Sub (Expression F)` or `ShiftRight (NExpr F)`. **Also**: matcher reduction no longer eta-expands structure variables, which broke the old struct-eval simplification wholesale (see below).
- **#13389 + #13550**: instance declarations are now validated — an `instance` with explicit arguments that TC synthesis can never infer errors at declaration time. All our `from_*` explicit-circuit constructors are deliberately of that shape (head dispatch `apply`s them, leaving the body-explicitness argument as a goal), so they become `@[implicit_reducible]` defs.
- **New `overlappingInstances` linter**: flags instance binders that shadow section variables or provide diamond-conflicting paths (e.g. `[FiniteField F] [Field F]`).
- **#13636**: `simpa using h` closes at *reducible* transparency; proofs relying on the final defeq check unfolding plain defs (`Soundness`, local `dite` motives) need the unfolds made explicit.
- **#13280**: new `backward.isDefEq.respectTransparency.types` (default **false** — opt-in, no forced migration). **New `backward.isDefEq.implicitBump`** (default **true**): implicit args are checked at `.instances` instead of the caller's transparency — a *softening* relative to 4.30.
- `bv_decide` now requires an explicit `import Std.Tactic.BVDecide`; `simps` config syntax changed (`(config := {isSimp := false})` → `(isSimp := false)`).

## Struct-eval normal form: simprocs + a unified `provable_struct_simp`

The biggest casualty of 4.31 was provable-struct evaluation: `circuit_norm` used to *unfold* `ProvableStruct.eval`/`toComponents` and rely on matchers eta-expanding opaque structure variables during simp's definitional reduction. Without matcher eta, that leaves un-keyable stuck terms `fromComponents (eval.go … (match x with …))`, and every downstream keyed lemma (projection lemmas, `mk.injEq`, …) stops firing.

The replacement is a small set of **simprocs** (`Clean/Circuit/StructEvalSimprocs.lean`, mirroring the witgen IR simprocs) that produce a component-preserving, row-level normal form by meta-level rewriting — recognizing constructor literals and structure projections syntactically, which rewrite lemmas cannot do generically:

- **literal decomposition**: `ProvableStruct.eval env ⟨a, b, …⟩ ~~> ⟨eval env a, eval env b, …⟩` (constructor literals only; opaque values deliberately stay folded row-level atoms, consumable by `h_input`-style row facts — this restriction is what makes the rule set confluent);
- **projection lift**: `Expression.eval env s.pc ~~> (ProvableStruct.eval env s).pc` (registered on `Expression.eval`, `ProvableStruct.eval`, and `Eval.eval` heads), including a proof-backed getElem branch `Expression.eval env (s.c[i]) ~~> (ProvableStruct.eval env s).c[i]` for projected vector fields;
- **constructor-equality split**: `(⟨a, …⟩ : S _) = ⟨a', …⟩ ~~> a = a' ∧ …` via the generated `mk.injEq`, keyed on `_ = _` and gated to `ProvableStruct`/`DerivedCircuitType` structures. This replaces per-type injEq collection and makes the split ordering-independent: it fires wherever the constructor shape materializes during any `circuit_norm` pass.

Rewrites are validated by definitional equality at `.all` transparency (structure eta *is* still part of defeq; only matcher reduction lost it) or carried by real lemma proofs where the identification is propositional (`getElem_eval_fields`/`getElem_eval_vector`).

On top of this, **`provable_struct_simp` was rewritten from scratch** as a single, self-contained tactic (`Clean/Utils/Tactics/ProvableStructSimp.lean`), replacing the three previous sub-tactics (`split_provable_struct_eq`, `decompose_provable_struct`, `simplify_provable_struct_eval`) and their delab-based lemma instantiation. It alternates two passes to a fixpoint: destructure struct variables that participate in struct-specific facts (under a folded `eval`, as projection bases, or opposite constructor literals — with field-based naming), then one `simp only` pass with a fixed struct-eval set whose members are also `circuit_norm` members — so the tactic and later `circuit_norm` passes agree on the normal form by construction. The old sub-tactic test files are merged into `TestProvableStructSimp.lean` with the same shape assertions.

A recurring bug family worth knowing about: **`mkAppM`/`mkAppOptM` inside simprocs elaborate at ambient transparency**, which does not see through the reducible `ProvableType`-derived `CircuitType` instance behind `Value M F`-spelled field types. Three separate simproc gates failed *silently* this way (constructor rebuild in the literal proc, `mk.injEq` application in the equality split, and the original registration attempts) — all now wrapped in `withTransparency .default` with comments explaining why.

## Other fixes

1. **Toolchain/mathlib bump** (commit 1).
2. **Circuit core adaptations**: `from_*` constructors as `@[implicit_reducible]` defs (#13389); surface-keyed `eval_neg`/`eval_sub` lemmas; per-instance `Witnessable` rfl-lemmas; witgen-IR per-operator desugar rfl-lemmas; `simpa`-transparency fixes; `LawfulMonad` via `pure_def`/`map_def`; `BVDecide` import.
3. **`cases_match_discr`** keeps the pervasive `fun (x, y) => ...` pattern-match style for `main`s working, and now also **iota-reduces matches on constructor literals** (arising when an enclosing `cases` substitutes an outer variable into a nested match's discriminant) — this fixed `ThetaXor`'s `elaborate_circuit` and made its hand-written demo instance redundant.
4. **Gadget/table proofs**: mostly *simplifications* — Circomlib Mux1/2/3 ported to `circuit_proof_start` (their hand-rolled intro/simpa proofs predate it), several now-obsolete 4.30-era workarounds in `ProcessBlocksInductive` deleted outright, `Conditional`/`Comparators`/Keccak/BLAKE3/SHA256/FemtoCairo all work with framework-level fixes only.
5. **`Fibonacci8`**: the `formalFibTable` soundness proof was removed (not fixed): the struct-eval validation at `.all` transparency zeta-expands the proof's `set env := …windowEnv…` binding back into the full `finalAssignment` computation inside that file's scoped legacy-defeq mode, exceeding heartbeats. `Fibonacci32` covers the same pattern and stays green; the table/circuit definitions remain (used by `ToJson`/`Trace`/plonky3 fixtures).

## Expected follow-ups

- Try removing the scoped `backward.isDefEq.respectTransparency false` sites (Fibonacci32 `FormalTable` literal, plus `Table.Inductive`) — #13281/#13363 was predicted to fix these; in practice the window-env assignment reasoning still needs them. (⚠ these declarations blow up in *memory*, not just heartbeats, when they fail — test carefully.) Restoring a `Fibonacci8` soundness proof hinges on the same cleanup, or on making the struct-eval defeq validation zeta-avoidant.
- Unify the vector/pair eval normal form with the struct one (opaque evals as folded row atoms, literal-gated element decomposition): `circuit_norm` currently carries an element-map spelling for vectors (`eval_var_fields` + `Vector.map_mk`) alongside the row-level struct form, and the getElem bridge lemmas can only live in the tactic set because they loop against the map spelling. Deliberately deferred: conceptually the refactor we want, but a library-wide normal-form migration.
- Consider `set_option simp.rfl.checkTransparency true` (#13317) as a CI lint for `circuit_norm` rfl-lemmas.


codygunton:
@mitschabaude Could you share your perspective on Lean version bumps? It seems to be common practice in the broader Lean ecosystem to constantly bump versions across breaking changes that are frequently nontrivial fix work (maybe not easy, but also not terrible). My understanding of why this is done is essentially that mathlib does this, and we 'want stuff that goes in to mathlib'. But It also means so much friction, including e.g. if you're using Aristotle, which is stuck on version 4.28. I think a big bump every 4-6 months, or maybe every 0 mod 5 minor version or something like this, sounds more reasonable. Do you think I'm overstating the friction, or am I missing out the benefits of the bumps?

mitschabaude:
> @mitschabaude Could you share your perspective on Lean version bumps?

@codygunton yeah I agree. the upgrade from 4.28 to 4.29 specifically was motivated by a library dependency we wanted to add. and then I figured that it was unusually hard, because of massive breakage to `simp`, and was actually made easier by doing 4.30 in the same pass. And 4.31 still has many related changes (and further breakage) so I figured to queue it right after to get over it.

Doing it only every couple of months is much saner, and I hope future version bumps will make it easier again to batch several together.

PR #426, opened 2026-07-10
Wire ComputableWitnesses into FormalCircuitBase

mitschabaude:
## What

Closes the *vacuous completeness* gap: our completeness formulation didn't guarantee that an honest prover environment actually exists. This adds a `computableWitnesses` obligation to `FormalCircuitBase` — every witness generator may only read the environment **below its own offset** (with the hint/data inputs held fixed) — and proves it for every circuit in the library.

## Design

- **Relaxed predicate.** `ProverEnvironment.AgreesBelow n env env'` bakes `env.hint = env'.hint ∧ env.data = env'.data` into the `∀ i < n` agreement, so the obligation holds hint/data fixed. `ComputableWitnesses` is stated once, in prover form, on `FormalCircuitBase`.
- **Subcircuit-local, non-recursive.** `Subcircuit.ComputableWitnesses` is a single node per subcircuit (via `Operations.forAll`); parent circuits discharge child-input premises through the composition lemmas (`toSubcircuit_computableWitnesses_onlyAccessedBelow`, and `_of_offset_eq` variants that keep the subcircuit's type-index offset separate from the computability offset so unification never has to reduce an operations list). ProvableType verifier-eval shadows live in the bundle namespaces; the framework layer stays prover-form.
- **`computable_witnesses` tactic.** Default for the field: most gadgets need **no field at all**, a handful pass hints (`computable_witnesses [main]`-style). A second list `computable_witnesses [...] closing [...]` supplies lemmas to the close routes only (after the goal is split to leaves) — for recursive eval decompositions like `eval_vector_set` that would blow the budget in the normalization positions (this is what closes `BLAKE3G`). `computable_witnesses_close` runs just that per-leaf close stage on the current goal, so manual proofs keep only their structural skeleton plus any genuinely bespoke leg (see `FinalizeChunk`). The module docs in `Clean/Utils/Tactics/ComputableWitnesses.lean` explain the pipeline at a glance. Gadget output-helper definitions can opt into expansion by the tactic via the `@[computable_witnesses_metadata]` label.
- **Handwritten proofs remain only where the content is genuinely mathematical** — e.g. the bit-decomposition gadgets (`Add32` and friends), whose proofs rest on arithmetic lemmas over the witness IR — or at documented tactic limits (each remaining manual proof's comment states the specific gap).

## Status

Full library and `CleanTests` build green under `--wfail`. `Clean/Utils/Test/TestComputableWitnesses.lean` is a regression suite pinning the tactic's key proof shapes.


mitschabaude:
## Porting this to Halo2-Clean

We reviewed how this PR's non-vacuity story transfers to Halo2-Clean (`Clean/Halo2`), where completeness has the same gap: `FormalCircuit.Completeness` quantifies over any `ProverEnvironment` satisfying `ExtendsWitnesses`, without establishing one exists. Conclusion: the whole theorem stack (`ComputableWitnesses` → `proverEnvironment` → `proverEnvironment_usesLocalWitnesses` → executable `witgen`) ports faithfully, with one generalization and two facts we must earn that the Main Clean representation gets for free.

### The formulation

**Prefix → determined set.** Main Clean's `AgreesBelow n` relies on the flat witness vector's linear order. Halo2 cells are `(column, absolute row)` with no intrinsic order — the order is the op-fold order. So the predicate generalizes to agreement on a threaded set `S` of *determined cells*:

```
AgreesOn S env env' := env.hint = env'.hint ∧ ∀ c ∈ S, env.get c = env'.get c
```

`ComputableWitnesses place ops` folds in the shape of `Constraints`/`ExtendsWitnesses`, threading `(regionIndex, S)`:
- seed `S₀` = the instance column (the prover's given publics — no Main Clean analogue);
- `assignAdvice` demands its witgen program is `AgreesOn S`-invariant, then adds its resolved cell to `S`;
- `assignFixed`/`loadTable` demand nothing (values are circuit data, env-independent) and add their cells to `S` — this subsumes this PR's `env.data` carve-out structurally: halo2 lookup tables *are* fixed cells, so "programs may read the table" is ordinary `S`-membership once the load op has passed;
- gates/copies/`constrainInstance`: skip.

**Subcircuit-local, non-recursive** — same design as here: each `FormalRegionCircuit` bundle carries a local obligation ("input cells ∈ S → every child program only accesses S ∪ the child's own earlier writes"), parents extend `S` by the child's write set via a per-bundle `writes_call` bookkeeping lemma (parallel to `regionCount`/`localLength`). The input premise `eval env input = eval env' input` simplifies to plain S-membership, since Halo2-Clean inputs are cells.

### What Main Clean gets for free and we don't

1. **No-clobber.** The key lemma `getElem?_dynamicWitnesses_of_lt` (later fold steps don't disturb written cells) is free from list-append semantics. With a cell map it needs a `WritesConsistent place ops` side condition: advice writes pairwise distinct after `place`-resolution; repeated fixed writes allowed iff equal-valued (these exist in practice — e.g. Sinsemilla's `q_s2` boundary re-pins). Threaded as an invariant through the `proverEnvironment_usesLocalWitnesses` induction.
2. **Write distinctness across regions.** Distinct offsets are distinct by construction in Main Clean. In Halo2, cross-region non-clobber depends on the placement: gadget-level statements need a "`place` gives per-column disjoint region windows" hypothesis — which is exactly what the Rust floor planners guarantee by construction, and what Halo2-Clean's pending `Requirements`-style well-formedness was always going to state. The runner doesn't create this obligation, it makes a pre-existing implicit assumption explicit. At the top level (concrete fixture `place`) it's decidable.

The other net-new piece is a **flat write-trace layer** (ordered `(resolved cell, program)` list) to play the role of `FlatOperation`/`forAll_toFlat_iff`, since Halo2-Clean's semantics recurse over the nested region structure directly. Same region-indexed walk the VK layout machinery (`Fixtures/Layout.lean`) already does verifier-side.

### Payoff theorem

```
ComputableWitnesses place ops ∧ WritesConsistent place ops →
  ∀ hint instances, ∃ env, AgreesOn S₀ seed env ∧ ExtendsWitnesses place env ops i
```

constructed by the runner (HashMap fold, executable — enabling cell-exact differential testing of witgen against a Rust `MockProver` advice-table dump, the witgen counterpart of the VK fixtures), unique on the written cells.

### Discharge story

Halo2-Clean is *better* positioned than Main Clean here: witgen programs are first-order IR (`WitgenIR`), so read sets are syntactically computable and the per-op obligation reduces to `readSet prog ⊆ S` — decide-style tactic territory, since ops, `place`, and addresses are concrete data. The only obstruction is the `.native` escape closures (incl. the `UnconstrainedSwaps` Bool closures), which admit no computable read set. Requiring `computableWitnesses` on every bundle therefore creates exactly the right pressure to finish migrating prover inputs to `ProverHint` + `.hintGet` IR programs and retire natives.

### Rollout

Post-ship (after the ironwood FV freeze): land the predicate + runner + bridge theorem as a leaf module, add the bundle field with a default tactic mirroring this PR, and let circuits opt in as they're touched — natives get hand-written `OnlyAccesses` lemmas until #32-style IR hints replace them.

PR #438, opened 2026-07-29
Bump Lean toolchain and mathlib to v4.32.2

gio54321:
## Human description

This is useful for [zk.golf](https://zk.golf) in light of a recent kernel bug. See [release notes](https://lean-lang.org/doc/reference/latest/releases/v4.32.2/#release-v4___32___2) for details.
We may want to hold merging into main as it could break downstream, and this is a patch release on top of v4.32. However, it's useful to have a branch here to reference this specific revision downstream.

## Agent description

Bumps `lean-toolchain` and mathlib **v4.30.0 → v4.32.2**, plus the resulting fallout across
61 `.lean` files (mostly proof-body adjustments, not detailed here).

`lake update` resolved every transitive dependency to revisions that **match mathlib
v4.32.2's own `lake-manifest.json` exactly** — no floating `main`/`master` drift.

## Verification

All three CI steps run locally and pass:

| Step | Result |
|---|---|
| `lake build --wfail` (core target, what CI gates) | exit 0 — **0 errors, 0 warnings** |
| `lake build CleanTests` | exit 0 |
| `scripts/check-consecutive-empty-lines.py` | clean |

No `sorry`, no `native_decide`, and **no `maxHeartbeats` changed** anywhere.

## Definition changes

**No `Spec`, `Assumptions`, `ProverSpec`, or `ProverAssumptions` was touched.** No circuit's
semantics changed. What did change form:

**New `circuit_norm` lemmas** (additive; nothing removed):
- `Clean/Circuit/Expression.lean` — `eval_neg` / `eval_sub`.
- `Clean/Circuit/WitnessIRSugar.lean` — 18 `rfl` lemmas normalizing witness-IR operator sugar
  to constructors, plus `attribute [circuit_norm] EqCond.eqCond`.
- `Clean/Circuit/Basic.lean` — `attribute [circuit_norm] Witnessable.witness Witnessable.witnessIR`.

**Three `main` definitions rewritten, eta-equivalently** — pattern-matching lambdas no longer
iota-reduce against an opaque `Var` argument, which stalled `elaborate_circuit`. Same
operations, offsets and output:
`Clean/Circomlib/Comparators.lean` (×3), `Clean/Gadgets/Addition8/Addition8.lean`,
`Clean/Gadgets/Keccak/ThetaXor.lean`.

**Instances demoted to `@[reducible] def`** — 4.32 hard-errors on instances with
non-inferrable explicit arguments. These are dispatched by the `explicit_circuit_constructor`
registry, not synthesis. `@[reducible]` preserves prior behaviour (`instance` implied it):
`ExplicitCircuit.from_bind`, `ExplicitCircuit.from_map`, `from_forEach`, `from_map_loop`,
`from_mapFinRange`, `from_foldl`, `from_foldlRange`, `ConstantLength.fromConstantLength'`.
**One new instance added:** `ExplicitCircuit.from_map_tc` (synthesis wrapper; `from_bind`
already had one).

Also `instance → def` for `Add8Channel` / `FibonacciChannel` in
`Clean/Examples/FibonacciWithChannels.lean` — their type is not a class; both were only ever
referenced by name.

**`def` → `theorem`** (new `linter.defProp`; each verified `Prop`-valued and not used as a
`Sort*` eliminator): `less_than_p`, `rotRight64_eq_bv_rotate`, `rotRight32_eq_bv_rotate`,
`KeccakRow.normalized_iff`, `everyRowTwoRowsInduction'`, `two_row_induction`,
`pushVarInput_offset`, `femtoCairoStepSoundness`, `femtoCairoStepCompleteness`. The
`Sort*`-valued `every_row_two_rows_induction` eliminators stay `def`.

**13 redundant instance binders removed** (new `linter.overlappingInstances`) across
`SourceSinkPath.lean` (9), `Provable.lean` (2), `Json.lean`, `FlatEnsemble.lean`.

## Caveats

⚠️ **`Provable.lean` instance diamond.** The two low-priority `Inhabited` instances now
resolve through the section's `[FiniteField F]` instead of a separately supplied `[Field F]`
(that pair *was* the reported diamond). Identical throughout this repo, but it matters if
downstream code instantiates them at a `Field` that is not a `FiniteField`.

⚠️ **Tactic behaviour change — widest blast radius.**
`Clean/Utils/Tactics/DecomposeProvableStruct.lean` checked `ProvableStruct`-ness by `whnf`ing
at `.reducible`. `Var M F` is the `CircuitType.Var` projection, which 4.32 no longer reduces
at that transparency, so locals spelled `Var M F` were **silently dropped** from
decomposition. Widened to `.instances` (not `.default`, to keep it minimal). This affects
every consumer of `provable_struct_simp` / `circuit_proof_start`; the full 105-module
transitive dependent set was rebuilt to confirm. `TestSimplifyProvableStructEval.lean` was
pinning exactly this and is unchanged.

⚠️ **Memory regression fixed — `Clean/Tables/Fibonacci32.lean`.** `fib_assignment` ballooned
to **~14 GB** and OOM-killed the whole build. `explicit_provable_type` in the `dsimp only` set
unfolds the `ProvableStruct` layout into raw `Vector.cast`s, leaving a goal not type-correct
at `instances` transparency, which the closing `rfl` then reconciled at default transparency.
Dropping it and restoring the `+instances` pass gives **3.1 GB / ~6 s**. An in-file comment
records why it must not be re-added. Relevant if CI runners are memory-limited.

**Behaviour-neutral:** `Clean/Utils/Bitwise.lean` gains `import Std.Tactic.BVDecide`
(`bv_decide` now requires it explicitly).

## Not addressed (pre-existing)

`Clean/Circomlib/Poseidon.lean` has a `sorry`'d primality `Fact`; `CompConstant.lean` has two
`native_decide` calls (both trivially replaceable with `norm_num`). The 10 `sorry` warnings in
`TestCircuitProofStart.lean` are intentional and live in `CleanTests`, which `--wfail` doesn't cover.


mitschabaude:
this should be rebased to / combined with #424 which went to some lengths to keep proof behavior unchanged

PR #442, opened 2026-08-04
Witgen IR: replace the Nat sort with u64 sort

### LINKED ISSUE #429: Replace WitgenIR's `NExpr` with bounded-range types
WitgenIR is currently designed to cover operations on three different primitive types: Field elements, natural numbers and Booleans or `FExpr`, `NExpr`, `BExpr`.

When we first had to extract this to real code in #414, it became obvious that `Nat` is a pretty annoying type to put in real code. Arbitrary-size natural numbers are hard to implement and expensive to run, and aren't what any implementations actually wants to do. The existence of operations that blow up size beyond the field size inherited from `NExpr.val`, like `NExpr.{mul, add}`, is especially dubious.

Would be better to be able to convert these operations to UInt64 or similar, but then, how do we keep track of the range in Lean?

gio54321:
## Human description

Closes #429.

Replaces `Nat` computation in the IR with U64. This is useful for exporting and for arguing about the concrete and asymptotic runtime of the witness generation.

Conversion between field elements and u64 is done by lifting to `Nat` and taking the modular reduction. Added instructions for computing `lt` over field elements (because casting to `Nat` is not there anymore), and for getting the `n`-th bit of a field element. Proof ergonomics should be fine.

## Agent description

`WitgenIR`'s integer sort was arbitrary-precision `ℕ`, which is not what a witness-generation backend can run. `NExpr` becomes `U64Expr`, evaluating to `UInt64` with Rust-`u64` wrapping:

- `U64Expr.val x` is `FiniteField.val x` truncated to 64 bits (`UInt64.ofNat`)
- `FExpr.ofU64 n` casts back through `FiniteField.fromNat n.toNat`
- `U64Expr.const n` truncates its `ℕ` payload the same way

`Step.letN` / `M.evalNat` / `UnconstrainedNat` follow as `letU` / `evalU64` / `UnconstrainedU64`. JSON export follows too (`ofU64`, `u64Eq`, `u64Lt`, `"sort": "u64"`).

### Operations that can't live in 64 bits

Two stay field-sorted, so no gadget has to assume a 254-bit element fits in a `u64`:

- `FExpr.bitOf x i` (`x.bit i`) and `VExpr.bitsOf` (`x.bits n`) — `CompConstant` decomposes 135 bits of a 254-bit prime, where `(x.val >>> i) % 2` returns `0` for every `i ≥ 64`. They evaluate to exactly the old terms, so `Bits`, `Num2Bits`, `Num2BitsNeg` and `BinSub` needed **no proof changes**.
- `VExpr.envRange offset` — replaces `FExpr.envGet` at a computed index; `witnessAny` and `Table.getRow` read `env.get (offset + i)` with an abstract `offset`, and `env.get ((offset + i) % 2^64)` is not provable there.

Net: `FExpr` −`envGet` +`bitOf`; `VExpr` +`envRange` +`bitsOf`. `BExpr` also gains `flt`, a field-sorted less-than over `FiniteField.val`s; `<?` is now an overloaded `LtCond` class mirroring `EqCond`, so `x <? y` on field operands is exact at any width while `x.val <? y.val` compares truncated `u64`s.

### Proofs

`circuit_norm` gained the `UInt64.toNat_*` push lemmas, so goals stay in `ℕ` but pick up `% 2^64` (values) and `% 64` (shift amounts). Gadgets work well inside those bounds, but the bound lives in their own assumptions — so neither an unconditional rewrite nor a conditional simp lemma can remove the wrap. New `circuit_norm` simproc `Witgen.u64Wrap` asks `omega`, with the local hypotheses as facts, whether the operand is already below the modulus, and rewrites only then. Other moduli are untouched.

That fixed five modules outright; for most of the rest the repair was exposing the byte/bit bound *before* the normalizing `simp`. Two needed more:

- **`Addition32Full`** — the bound chains through the carries and can't be hoisted, so `add8_completeness` takes its byte/bool facts as leading arguments and unwraps internally from the `< 512` bound it already derives.
- **`SHA256.Add32`** — `bitsVal_eval` restated as `= UInt64.ofNat (evalBitsNat env a)`, unconditionally. Generally: state lemmas about u64 quantities as `= UInt64.ofNat <ℕ expr>`, not about `.toNat`, since `omega` treats `UInt64.toNat x` as an unbounded atom.

### One API change worth review

`FemtoCairo` indexes its program and memory tables at a computed row, so completeness needs them to fit in `2^64` rows. Rather than `Fact (p < 2^64)` — which would restrict soundness too — the bounds went into prover-side assumptions only: `programSize ≤ 2^64` (`fetchInstruction.ProverAssumptions`, `femtoCairoStepAssumptions`, `InitialStateAssumptions`) and `mem.size ≤ 2^64` (`MemoryCompletenessAssumption`). Soundness stays fully general; no `Spec` was weakened anywhere in this PR.

### Tests

- `TestU64Wrap.lean` — the simproc erases wraps when bounded, leaves other moduli alone, does **not** fire unbounded; plus `<?` sort dispatch and the contrast case where `x.val <? y.val` compares `% 2^64` values.
- `Examples/WitnessExport.lean` — golden `#witgen_json` pinning the u64 nodes and `bitsOf`.

### DIFF
--- a/Clean/Circuit/WitnessIR.lean
+++ b/Clean/Circuit/WitnessIR.lean
@@ -239,6 +239,55 @@ attribute [circuit_norm]
 theorem UInt64.eq_iff_toNat_eq {a b : UInt64} : a = b ↔ a.toNat = b.toNat :=
   UInt64.toNat_inj.symm
 
+section U64Wrap
+open Lean Meta Simp
+
+/-- Closed `ℕ` value of an expression, seeing through `b ^ k` (which `Meta.evalNat` does
+not reduce at the `Monoid.toPow` instance that `2 ^ 64` elaborates to). -/
+private partial def natLit? (e : Expr) : MetaM (Option ℕ) := do
+  if let some n ← evalNat e |>.run then return some n
+  if e.isAppOfArity ``HPow.hPow 6 then
+    let args := e.getAppArgs
+    let some b ← natLit? args[4]! | return none
+    let some k ← natLit? args[5]! | return none
+    if k ≤ 64 && b ≤ 64 then return some (b ^ k)
+  return none
+
+/--
+Erase a `u64` wrap that provably does not wrap.
+
+Pushing `UInt64.toNat` through a `UExpr` evaluation leaves two kinds of truncation
+behind: `n % 2^64` (from `UExpr.val`, `UExpr.const` and the arithmetic operations) and
+`n % 64` (the shift-amount mask of `<<<` / `>>>`). Gadgets always work well inside those
+bounds — bytes, 32-bit limbs, bit indices — but the bound lives in the gadget's own
+assumptions (`x.val < 256`, `i < 32`, …), so no unconditional rewrite can remove the
+wrap, and a conditional simp lemma cannot discharge it either.
+
+This simproc closes exactly that gap: on `n % 2^64` / `n % 64` it asks `omega` (with the
+local hypotheses as facts) whether `n` is already below the modulus, and rewrites to `n`
+when it is. Other moduli are left alone, so ordinary `% 256`-style specification
+arithmetic is untouched.
+-/
+private def u64WrapSimproc (e : Expr) : SimpM Simp.Step := do
+  unless e.isAppOfArity ``HMod.hMod 6 do return .continue
+  let args := e.getAppArgs
+  let n := args[4]!
+  let m := args[5]!
+  unless (← whnfR (← inferType n)).isConstOf ``Nat do return .continue
+  let some mVal ← natLit? m | return .continue
+  unless mVal == 18446744073709551616 || mVal == 64 do return .continue
+  let g ← mkFreshExprMVar (← mkAppM ``LT.lt #[n, m])
+  try
+    let some g' ← g.mvarId!.falseOrByContra | return .continue
+    g'.withContext do Lean.Elab.Tactic.Omega.omega (← getLocalHyps).toList g'
+  catch _ =>
+    return .continue
+  return .done { expr := n, proof? := ← mkAppM ``Nat.mod_eq_of_lt #[g] }
+
+simproc u64Wrap ((_ : ℕ) % _) := u64WrapSimproc
+attribute [circuit_norm] u64Wrap
+end U64Wrap
+
 variable {M : TypeMap} [ProvableType M]
 
 /-- Evaluation for higher-level provable types. -/
--- a/Clean/Circuit/WitnessIR.lean
+++ b/Clean/Circuit/WitnessIR.lean
@@ -65,8 +65,6 @@ inductive FExpr (F : Type) where
   equals `Nat.cast` on prime fields, but interprets binary digits as coefficients
   on binary fields, where `Nat.cast` would collapse via the characteristic). -/
   | ofU64 (n : U64Expr F)
-  /-- Bit `i` of `FiniteField.val x`, as the field element `0` or `1`. -/
-  | bitOf (x : FExpr F) (i : ℕ)
   | ite (c : BExpr F) (t e : FExpr F)
   /-- Read an expression list at a computed index, 0 if out of range -/
   | listGet (xs : List (FExpr F)) (i : U64Expr F)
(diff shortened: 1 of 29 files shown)


Clean/Circuit/WitnessIR.lean
```diff
@@ -52,59 +54,64 @@ inductive FExpr (F : Type) where
   /-- Embedded circuit expression; this is how callbacks read input vars and earlier
   witnesses (`env x`). -/
   | expr (e : Expression F)
-  /-- Read the environment at a computed index (e.g. consecutive vars in a mapRange). -/
-  | envGet (i : NExpr F)
```
mitschabaude, line 56: we shouldn't delete this I think? we can change it to `envGet (i : U64Expr F)`

Clean/Circuit/WitnessIR.lean
```diff
@@ -52,59 +54,64 @@ inductive FExpr (F : Type) where
...
-  | hintGet (key : String) (n : ℕ) (row : NExpr F) (col : Fin n)
+  | hintGet (key : String) (n : ℕ) (row : U64Expr F) (col : Fin n)
 
-/-- Nat-sorted witness expressions. -/
-inductive NExpr (F : Type) where
+/-- u64-sorted (bounded-nat) witness expressions. All operations wrap modulo `2^64`,
+matching Rust's `u64`. -/
+inductive U64Expr (F : Type) where
+  /-- A `ℕ` literal, truncated to 64 bits. -/
   | const (n : ℕ)
```
mitschabaude: this should just become a `UInt64` constant

Clean/Circuit/WitnessIR.lean
```diff
@@ -52,59 +54,64 @@ inductive FExpr (F : Type) where
...
   /-- Field inverse, with `0⁻¹ = 0` (the `IsZeroField` witness). -/
   | inv (x : FExpr F)
-  /-- Cast from the Nat sort via `FiniteField.fromNat` (the inverse of `val`;
+  /-- Cast from the u64 sort via `FiniteField.fromNat` (the inverse of `val`;
   equals `Nat.cast` on prime fields, but interprets binary digits as coefficients
   on binary fields, where `Nat.cast` would collapse via the characteristic). -/
-  | ofNat (n : NExpr F)
+  | ofU64 (n : U64Expr F)
+  /-- Bit `i` of `FiniteField.val x`, as the field element `0` or `1`. -/
+  | bitOf (x : FExpr F) (i : ℕ)
```
mitschabaude: adding this works for me 👍🏻 but wouldn't it be nicer to return a `BExpr`?

Clean/Circuit/WitnessIR.lean
```diff
@@ -228,6 +304,11 @@ constraint system — porting its computation to the IR must keep it a hint. -/
 inductive VExpr (F : Type) : ℕ → Type where
   | lit {n : ℕ} (es : Vector (FExpr F) n) : VExpr F n
   | mapRange (n : ℕ) (body : FExpr F) : VExpr F n
+  /-- `n` consecutive environment cells starting at `offset` — the "witness a fresh
+  variable per cell" former (`witnessAny`, table rows). -/
+  | envRange {n : ℕ} (offset : ℕ) : VExpr F n
+  /-- The `n` low bits of `FiniteField.val x`, each as the field element `0` or `1`. -/
+  | bitsOf {n : ℕ} (x : FExpr F) : VExpr F n
```
mitschabaude, line 310: why is this needed, seems like a trivial combination of `mapRange` and `bitOf`?

Clean/Circuit/WitnessIR.lean
mitschabaude: oh I see, it typically takes an arbitrary input `offset : Nat` and adds something to that. so this is probably good to have in a follow-up which moves `Environment.get` to be `UInt64`-indexed instead of `Nat`-indexed.

mitschabaude (commented):
awesome, just a few comments

Clean/Circuit/WitnessIR.lean
gio54321: > moves Environment.get to be UInt64-indexed instead of Nat-indexed

Do we actually want to do that? U64 is just a detail of the IR, no need to expose it to the soundness theorem

Clean/Circuit/WitnessIR.lean
gio54321: The problem is that `bitOf` takes a `Nat` currently, and `mapRange` index is a U64.
We could make `bitOf` accept a U64 expr as index, at the cost of slightly worse proof ergonomics.
It seems the better choice anyway

mitschabaude (approved):
nice

PR #450, opened 2026-08-13
feat: Modify `AgreesBelow` to include data and hint agreement.

dtumad:
## The problem

`ProverEnvironment.AgreesBelow` (`Clean/Circuit/Basic.lean:424`) is today

```lean
def ProverEnvironment.AgreesBelow (n : ℕ) (env env' : ProverEnvironment F) :=
  ∀ i < n, env.get i = env'.get i
```

but a `ProverEnvironment` has **three** channels a witness generator can read: cells (`get`, via `Expression.eval`), committed prover data (`data`, via `FExpr.dataGet`) and prover hints (`hint`, via `FExpr.hintGet`). Only the first is constrained.

Since `Operations.ComputableWitnesses` is exactly `AgreesBelow n env env' → compute.eval env = compute.eval env'`, the obligation for any witness program containing a `dataGet` or `hintGet` node asks you to show that two environments with *entirely unrelated* committed data compute the same witness. That is not merely hard — it is **false**, and this PR proves it (`Examples.DataWitness.not_computable_from_cells_alone`). So no circuit using the IR's documented nondeterministic escape hatch (`Examples.FemtoCairo`'s memory) can be shown to generate witnesses honestly, and `Circuit.witgen_usesLocalWitnesses` is unavailable to it.

## The change

```lean
def ProverEnvironment.AgreesBelow (n : ℕ) (env env' : ProverEnvironment F) :=
  (∀ i < n, env.get i = env'.get i) ∧ env.data = env'.data ∧ env.hint = env'.hint
```

plus `AgreesBelow.get_eq` / `.data_eq` / `.hint_eq` accessors and `agreesBelow_rfl`, so downstream scripts do not reach for `.2.1`. The body is a plain `∧` rather than a structure, so `⟨h, rfl, rfl⟩` and `simp_all` keep working.

**Nothing downstream weakens.** `AgreesBelow` sits in hypothesis position at every occurrence except one, so strengthening it *weakens* every obligation — existing instances stay true.

## The example

`Clean/Examples/DataWitness.lean` is the smallest circuit reading committed prover data — a memory read modelled on `Examples.FemtoCairo` — and records both directions:

- `readMemory_computableWitnesses` discharges a genuine `Circuit.ComputableWitnesses`. Its proof is two rewrites, one per environment channel the generator reads: `AgreesBelow.data_eq` for the committed table, and the input hypothesis for the address expression.
- `not_computable_from_cells_alone` **proves** the converse: two environments agreeing on every cell but committing different memory compute different witnesses. The claim above that the old obligation was false is a theorem in the tree, not an argument in a PR description.

### DIFF
--- a/Clean.lean
+++ b/Clean.lean
@@ -20,6 +20,7 @@ import Clean.Examples.AddOperations
 import Clean.Examples.Add32Explicit
 import Clean.Examples.ToJson
 import Clean.Examples.HintExample
+import Clean.Examples.DataWitness
 import Clean.Examples.FemtoCairo.FemtoCairo
 import Clean.Examples.FemtoCairo.Plonky3Helpers
 import Clean.Examples.FemtoCairo.Plonky3TestData
--- a/Clean/Circuit/Basic.lean
+++ b/Clean/Circuit/Basic.lean
@@ -421,8 +421,37 @@ def FlatOperation.dynamicWitnesses (ops : List (FlatOperation F)) (hint : Prover
 def FlatOperation.proverEnvironment (ops : List (FlatOperation F)) (hint : ProverHint F) (init : List F) :=
   ProverEnvironment.fromList (FlatOperation.dynamicWitnesses ops hint init) hint
 
+/--
+Two prover environments are indistinguishable to a witness generator running at offset `n`.
+
+A `ProverEnvironment` has three channels a witness generator can read: cells (`get`, via
+`Expression.eval`), committed prover data (`data`, via `FExpr.dataGet`) and prover hints
+(`hint`, via `FExpr.hintGet`). All three appear here, but not in the same way: cells are
+determined incrementally by witness generation, so only those *below* the current offset are
+constrained, whereas `data` and `hint` are ambient inputs fixed before generation begins, so
+they are constrained in full.
+-/
 def ProverEnvironment.AgreesBelow (n : ℕ) (env env' : ProverEnvironment F) :=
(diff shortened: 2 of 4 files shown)


dtumad:
@mitschabaude I'm not sure if this is the right approach or if I am maybe doing things wrong in `sp1-lean`, but it does seem to be an issue for some cases there.

mitschabaude:
@dtumad yes this is needed and https://github.com/Verified-zkEVM/clean/pull/426 which should land soon also has this change

mitschabaude (approved):

dtumad:
This is approved and mergeable; happy to rebase whenever it is convenient.

Two notes to make #426's life easier whichever way this goes:

- **Conjunct order differs.** #426 defines `AgreesBelow` as `(∀ i < n, …) ∧ hint ∧ data`; this PR
  has `… ∧ data ∧ hint`. Whichever lands second is a two-conjunct swap. This PR also adds the
  `.get_eq` / `.data_eq` / `.hint_eq` accessors and `agreesBelow_rfl`, and leaves `AgreesBelow`
  opaque rather than unfolding it under `circuit_norm`/`grind` — happy to drop either if #426's
  shape is preferred.
- **`Clean/Examples/DataWitness.lean` is only here.** It carries `not_computable_from_cells_alone`,
  which *proves* the old obligation was false for any generator reading `hintGet`/`dataGet` — not
  merely awkward. That example seems worth keeping in the tree independently of which PR carries
  the definitional change.

PR #457, opened 2026-09-01
Upgrade Lean and Mathlib to v4.33.1 and add release tagging

dtumad:
## Summary

- Upgrade Lean and Mathlib from v4.32.2 to v4.33.1, the newest stable release.
- Use the canonical scoped Lake dependency syntax and refresh the transitive manifest.
- Adapt explicit circuit metadata and proofs to Lean 4.33 while preserving semantic subcircuit boundaries.
- Simplify several proof paths, including Addition32Full, FemtoCairo, BLAKE3 ApplyRounds, and SHA-256, without raising resource limits.
- Add ArkLib's release-tag workflow so changes to `lean-toolchain` on `main` or `master` create the corresponding Lean release tag.

## Proof-regression guardrails

- No new sorry or admit declarations.
- No maxHeartbeats or maxRecDepth changes.
- Child gadgets remain bundled subcircuits; new abstraction boundaries use FormalCircuit or ElaboratedCircuit metadata.
- Existing test-only sorry fixtures are unchanged.

## Validation

- `lake build --wfail`
- `lake build CleanTests`
- `python3 scripts/check-consecutive-empty-lines.py`
- `cargo test --release -- --nocapture test_lean_circuit_end_to_end`
- `cargo test test_tampered_multiplicity`
- `cargo test --release test_tampered_multiplicity`
- `cargo test --release -- --nocapture test_femtocairo`
- `git diff origin/main --check`


alexanderlhicks:
Was about to open a PR for this and luckily this saved me some work. I added the release tag workflow so that a corresponding tag will be created when this gets merged.

mitschabaude (changes requested):
the tactic layer, especially `elaborate_circuit`, appears to regress significantly so this shouldn't be merged as is.

can you give more insight into what were the relevant changes in 4.33 and why we now need to annotate everything with `implicit_reducible`?

alexanderlhicks:
I haven't inspected Devon's changes so I'll go take a look.

alexanderlhicks:
ah it seems to stem from 4.33: https://lean-lang.org/doc/reference/latest/releases/v4.33.0/#The-Lean-Language-Reference--Release-Notes--Lean-4___33___0-_LPAR_2026-08-10_RPAR_--Highlights--Breaking-Changes

alexanderlhicks:
Good catch. The relevant upstream change is not specific to the 4.33.1 patch; it landed in 4.33.0 and is inherited by 4.33.1:

- [`lean4#13895`](https://github.com/leanprover/lean4/pull/13895) enables `backward.isDefEq.respectTransparency.types` by default.
- The [4.33.0 migration notes](https://lean-lang.org/doc/reference/latest/releases/v4.33.0/#breaking-changes) describe this under **Breaking Changes → Transparency**.
- [`lean4#14466`](https://github.com/leanprover/lean4/issues/14466) is an open issue with the same characteristic symptom: `simp` can leave a goal that is not type-correct at `.implicit` transparency.

Before 4.33, when a tactic assigned a metavariable at reducible/instances/implicit transparency, Lean compared the metavariable's type with the assigned term's type at `.default` transparency. In 4.33 it uses `.implicit`, so ordinary semireducible definitions are no longer unfolded during that check. `@[implicit_reducible]` makes a selected definition visible at that level.

Clean hits this especially hard because `ExplicitCircuit`, `ExplicitCircuits`, and `ElaboratedCircuit` are indexed by the circuit function itself. `infer_explicit_circuits` unfolds a named `main` to expose its `bind`/`pure` structure, so the inferred evidence can be indexed by the unfolded lambda. `elaborate_circuit` and later simplification then reuse that evidence where the expected index is the named `main`. That comparison succeeded at `.default` before 4.33; at `.implicit` it fails unless `main` unfolds. This is why the workaround spread to circuit `main`/`circuit` definitions.

I agree that the current breadth is not a good endpoint: this PR adds 127 `implicit_reducible` annotations. Making large circuit bodies globally unfoldable during implicit-argument matching is exactly the kind of performance risk the new transparency regime is intended to control.

Some annotations are independently justified because the definition genuinely computes a dependent index—for example `combinedSize'`, and the typed-to-raw conversions whose arity appears in a `Vector` type. The blanket annotations on gadget bodies are different: they are compensating for how the tactic changes the circuit index.

As a control, I tried unmodified 4.32 source on 4.33.1 with Lean's documented compatibility option (`backward.isDefEq.respectTransparency.types false`). After making the separate structural `combinedSize'` fix, `Channel`, `Conditional`, and `Equality` compile without their new annotations. That confirms that much of the annotation churn is tied to the transparency change, not a semantic change in those gadgets. It is not a proposed global fix: other 4.33 changes still require migration, and globally restoring the old behavior gives up the scalability benefit.

My recommendation is to hold this PR, reduce the named-`main`/unfolded-lambda mismatch in `TestElaborateCircuit`, and fix it in the inference/elaboration layer (or use the old option only inside the narrow tactic operation if that proves sufficient). Then we should remove the blanket annotations, retain only definitions that genuinely compute dependent indices, and benchmark `elaborate_circuit` before merging.

alexanderlhicks:
Addressed the tactic-layer regression in `aaa811f6`.

The fix is now centralized in `Clean/Circuit/Explicit.lean`:

- `infer_explicit_circuit(s)`, `elaborate_circuit`, and `elaborate_circuit_with` use the pre-4.33 metavariable-type transparency behavior only within the tactic invocation.
- Unfolding a named circuit wrapper now transports inferred metadata back across a definitional equality, so the generated `ExplicitCircuit(s)` proof stays indexed by the original named circuit.
- A focused regression test consumes such an instance after the scoped compatibility option has ended.

This removes 123 of the migration's 127 new `@[implicit_reducible]` annotations. The four retained additions are limited to definitions that occur in genuinely dependent indices:

- `ProvableType.combinedSize'`
- `Channel.toRaw`
- `ChannelInteraction.toRaw`
- `MemoryAccessList.isTimestampSorted`

The hand-written Keccak elaborated instance gets the same option locally, without making its circuit definitions globally reducible.

Local validation is green:

- `lake --wfail build`
- `lake build CleanTests`
- focused `TestElaborateCircuit` and `Add32Explicit` builds
- consecutive-empty-line check
- all four Plonky3 CI commands (end-to-end, debug/release tampered multiplicity, FemtoCairo)

@mitschabaude this should address the requested changes and restore the intended circuit opacity.

Clean/Utils/Test/TestMixedCircuitType.lean
```diff
@@ -65,7 +65,7 @@ def parent : GeneralFormalCircuit F field field where
     circuit_proof_start [circuit]
     -- The subcircuit spec should be stated in terms of the parent input, not
     -- the inline mixed child input passed to the subcircuit.
-    guard_hyp h_holds : input * env.get i₀ = 1
+    change input * env.get i₀ = 1 at h_holds
```
mitschabaude: this is a regression of the `circuit_proof_start` tactic behavior, exactly what this test was supposed to catch (it no longer would with that change)

Clean/Gadgets/Keccak/ThetaC.lean
```diff
@@ -24,8 +24,22 @@ def Spec (state : KeccakState (F p)) (out : KeccakRow (F p)) :=
...
+  output_eq := by
+    intro state i₀
+    simp only [main, Xor64.circuit, Xor64.elaborated, circuit_norm]
+  subcircuitsConsistent := by
+    intro state i₀
+    simp only [main, Xor64.circuit, Xor64.elaborated, circuit_norm]
+    omega
+  channelsLawful := by
+    intro state i₀
+    simp only [main, Xor64.circuit, Xor64.elaborated, circuit_norm]
```
mitschabaude: I want to fix regressions like these instead of merging the PR with a weaker `elaborate_circuit` tactic.

Clean/Examples/FibonacciWithChannels.lean
```diff
@@ -253,12 +254,14 @@ def fibonacciEnsemble := SoundEnsemble.empty (F p) fieldTriple
     (by simp [circuit_norm, pushBytes]) (by simp [circuit_norm, pushBytes])
   |>.addFinishedChannel BytesChannel.toRaw
   |>.addTable ⟨ add8 ⟩
-    (by simp +instances [circuit_norm, add8]) (by simp [circuit_norm, add8])
+    (by simp +instances [circuit_norm, add8])
+    (by simp [circuit_norm, add8, SoundEnsemble.addTable])
```
mitschabaude: can't we put `SoundEnsemble.addTable` in circuit_norm?

Clean/Circomlib/Bitify2.lean
```diff
@@ -115,16 +115,16 @@ def main (input : Vector (Expression (F p)) 254) := do
...
 def circuit : GeneralFormalCircuit (F p) (fields 254) field where
   main
-  -- elaborated := by elaborate_circuit_with {
-  --   localLength _ := 265
-  --   output _ i := varFromOffset field (i + 264)
-  -- } using (by simp +arith only [circuit_norm])
+  elaborated := by elaborate_circuit_with {
+    localLength _ := 265
+    output _ i := varFromOffset field (i + 264)
+  } using (by simp +arith only [circuit_norm])
```
mitschabaude: change necessary?

Clean/Gadgets/Keccak/KeccakRound.lean
```diff
@@ -23,8 +23,6 @@ def Spec (rc : UInt64) (state : KeccakState (F p)) (out_state : KeccakState (F p
   out_state.Normalized
   ∧ out_state.value = keccakRound state.value rc
 
--- overridden so that `Permutation` can keep its nice proof which uses
--- `Vector.mapRange` instead of `Vector.mapFinRange`.
 @[reducible]
```
mitschabaude, line 28: spurious comment removal

Clean/Gadgets/Or/Or8.lean
```diff
@@ -4,7 +4,7 @@ import Clean.Gadgets.Xor.ByteXorTable
 variable {p : ℕ} [Fact p.Prime] [p_large_enough: Fact (p > 512)]
 
 namespace Gadgets.Or.Or8
-open Xor (ByteXorTable)
+open _root_.Gadgets.Xor (ByteXorTable)
```
mitschabaude:
```suggestion
open Gadgets.Xor (ByteXorTable)
```

Clean/Utils/Tactics/CircuitProofStart.lean
```diff
@@ -157,6 +157,32 @@ elab_rules : tactic
...
+        | none => false
+      else
+        e.getAppFn.isConstOf ``ElaboratedCircuit.output &&
+          (args[args.size - 4]?.map (·.isLambda)).getD false).isSome
+  if hHoldsHasFormalOutput then
+    try (evalTactic (← `(tactic| simp only [FormalCircuitBase.output,
+      ElaboratedCircuit.output]
+      at $(mkIdent `h_holds):ident))) catch _ => pure ()
+    try (evalTactic (← `(tactic| simp only [circuit_norm]
+      at $(mkIdent `h_holds):ident))) catch _ => pure ()
```
mitschabaude: this is ugly, hard to understand

Clean/Utils/Tactics/CircuitProofStart.lean
```diff
@@ -157,6 +170,17 @@ elab_rules : tactic
...
+  -- A child unfolded above may leave behind the output projection of its inline body.
+  if (← getGoals).isEmpty then return
+  let hasInlineOutput ← withMainContext do
+    let some hHolds := (← getLCtx).findFromUserName? `h_holds | pure false
+    pure ((← instantiateMVars hHolds.type).find? isInlineCircuitOutput).isSome
+  if hasInlineOutput then
+    try (evalTactic (← `(tactic| simp only [FormalCircuitBase.output,
+      ElaboratedCircuit.output] at $(mkIdent `h_holds):ident))) catch _ => pure ()
+    try (evalTactic (← `(tactic| simp only [circuit_norm]
+      at $(mkIdent `h_holds):ident))) catch _ => pure ()
```
mitschabaude: maybe we should just add `FormalCircuitBase.output` to `circuit_norm` and remove all this?

mitschabaude (approved):

mitschabaude (approved):

Issue #396, opened 2026-06-05
Circomlib/Poseidon.lean is outside the module graph, so CI never checks it


pirapira:
> [!NOTE]
> This issue was written by Claude Code, at the request of @pirapira (filed while reviewing #375).

## Problem

`Clean/Circomlib/Poseidon.lean` is not imported by any module (nothing in the repo imports `Clean.Circomlib.Poseidon`), so `lake build` / CI never elaborates it. Any refactor can silently break the file.

This already bit us in #375: the PR description notes Poseidon was "not in the module graph, so it's easily missed" and had to be fixed separately as a checklist item.

## Why it is out of the graph

The file contains one `sorry` — the BN254 primality fact at `Clean/Circomlib/Poseidon.lean:22`:

```lean
instance : Fact (Nat.Prime BN254_PRIME) := ⟨by sorry⟩
```

`Clean/Circomlib/Poseidon.md` documents that closing it needs a Pratt/Lucas primality certificate. Keeping the sorry out of the main graph is intentional; the side effect is zero CI coverage for ~480 lines of verified-circuit code.

## Suggested fixes (either or both)

1. **Close the sorry** with a primality certificate (e.g. `Mathlib.NumberTheory.LucasLehmer`-style certificate or `Nat.Prime` via `pratt`-certificate tooling), then import the file from the module graph like every other Circomlib port.
2. **Short term:** add an explicit CI step that builds the module despite the sorry, e.g. `lake build Clean.Circomlib.Poseidon`, so regressions surface as build failures (the sorry only produces a warning, not an error).

Option 2 is cheap and immediately prevents silent breakage; option 1 is the proper end state.

mitschabaude:
my favored version of option 2 would be to just add Poseidon to the existing CleanTests target which allows sorries

proper fix is downstream of https://github.com/Verified-zkEVM/clean/pull/370

PR #460, opened 2026-09-10
feat(module-system): migrate Clean to Lean's module system

### LINKED ISSUE #461: deriving ProvableStruct diverges for nested ProvableStruct fields under the module system
Under Lean's module system, `deriving ProvableStruct` does not terminate for a structure whose field is itself a derived `ProvableStruct`. This is the one thing blocking #460 (the module-system migration); everything else in that PR is finished.

## Repro

On the `dhsorens/module-system` branch of #460, save this anywhere and run `lake env lean` on it:

```lean
import Clean.Circuit.Provable
import Clean.Utils.Tactics.ProvableStructDeriving

namespace Repro
structure Inner (F : Type) where
  a : F
  b : F
deriving ProvableStruct

structure Outer (F : Type) where
  x : Inner F
  y : F
deriving ProvableStruct
end Repro
```

```
error: (deterministic) timeout at `whnf`, maximum number of heartbeats (200000) has been reached
error: failed to synthesize instance of type class ProvableStruct Outer
error: Application type mismatch: The argument
  ProvableTypeList.cons x (ProvableTypeList.cons y ProvableTypeList.nil)
has type
  ProvableStruct.ProvableTypeList F [{type := Inner, provableType := inferInstance},
                                     {type := field, provableType := inferInstance}]
but is expected to have type
  ProvableStruct.ProvableTypeList F (components Outer)
```

`Inner` alone derives fine. So does any structure whose fields are `F`, `Vector F n`, or a type variable — it is specifically a field whose `ProvableType` instance arrives via `ProvableStruct.toProvableType`.

## What is happening

`components Outer` is a projection of the instance currently being constructed, so it cannot reduce. The unifier falls back to whnf and evaluates `componentsToElements` instead. `set_option diagnostics true` at the failing site:

```
[reduction] List.rec ↦ 178129
[reduction] Eq.rec ↦ 119510
[reduction] ProvableStruct.ProvableTypeList.rec ↦ 86542
[reduction] ProvableStruct.componentsToElements ↦ 86534
[reduction] Array.map ↦ 49920
[reduction] Array.foldlM.loop ↦ 9910
[reduction] Vector.flatten ↦ 4992
```

For a struct with a 64-element `ProvableVector` field this is hopeless, but as the repro shows it already fails on two scalar fields.

## Ruled out

I chased each of these to the bottom while porting, so nobody needs to repeat them:

- **Not exposure.** The repro fails with the *consuming* file left classic, and a classic importer loads private bodies and unexposed definitions alike.
- **Not the missing `Array` bodies.** Removing `import all Init.Data.Array.Basic` from `Clean/Circuit/Provable.lean` (and stubbing the three proofs that need it) does not change the failure.
- **Not the deriving handler being a `meta` module.** Reverting `Clean/Utils/Tactics/ProvableStructDeriving.lean` to classic does not change the failure.
- **Not a heartbeat budget.** 2,000,000 heartbeats also times out; this is a blowup, not a slow path.
- **Not `backward.isDefEq.respectTransparency`.** Setting it `false` around the structure, as `Clean/Tables/Fibonacci8.lean` already does for the table elaboration, does not help either.

## Suspected fix

In `ProvableStructDeriving.lean`, emit `toComponents`/`fromComponents` against the *literal* component list rather than against `components Outer`, so the unifier never has to reduce a projection of the instance under construction. Concretely: give the generated `fromComponents` an explicit type ascription naming the literal list, or build the instance with `@ProvableStruct.mk _ [..literal..] ..` instead of structure-instance notation.

I have not attempted this — it is a design call for whoever owns that file, and it is 772 lines of metaprogramming I did not want to change blind in a migration PR.

## Impact

Two files on `main` hit it once the core is modulized:

- `Clean/Gadgets/BLAKE3/FinalizeChunk.lean` (`lake build`)
- `Clean/Utils/Test/TestProvableStructSimp.lean` (`lake build CleanTests`)

Plus `Clean/Tables/Fibonacci8.lean`, which times out in a related way during table elaboration. Those three and their reverse cone (`Clean.lean`, `Clean/Examples/ToJson.lean`) are left classic in #460, which is also why that PR does not yet turn on Lake's `requiresModuleSystem` ratchet.

dhsorens:
Migrates Clean to Lean's module system: `module`, `public import`, `@[expose] public section`. **All 165 files Lake builds** are converted, `lake build` and `lake build CleanTests` are green, and the Lake ratchet is on.

Closes #461.

Produced with **upstream Lean's own rewriter**, [`script/Modulize.lean`](https://github.com/leanprover/lean4/blob/master/script/Modulize.lean) from `leanprover/lean4` at v4.33.1:

```bash
lake env lean --run /path/to/lean4/script/Modulize.lean <files>        # ordinary modules
lake env lean --run /path/to/lean4/script/Modulize.lean --meta <files> # metaprogram modules
```

It is not vendored here — after this, new files are simply written module-style, and `requiresModuleSystem := true` on `lean_lib Clean` makes Lake say so if one is not.

Semantics are preserved deliberately: every file is blanket-exposed and every import is `public`, so no downstream proof loses an unfolding it had. Tightening that (demoting `public import`, removing the blanket `@[expose]`, marking internals `private`) is where the payoff grows, and it is per-file work that can land independently.

## ⚠️ The one change that is not header churn

`Clean/Circuit/Provable.lean` — **please review this independently of the other 166 files.** It changes how `ProvableType.fromStruct` reduces, on `main` as well as here.

`deriving ProvableStruct` did not terminate for a structure whose field is itself a derived `ProvableStruct` (#461). The cause is not the module system as such:

`ProvableTypeList` is indexed by `List WithProvableType`, and `WithProvableType` carries its `ProvableType` instance **as data** (`provableType : ProvableType type`). Elaborating the derived `fromComponents`'s pattern match therefore has to decide definitional equality of two `WithProvableType` records, which compares the instances themselves. When one of them is `ProvableType.fromStruct`, defeq unfolds the bridge instance through `combinedSize'` — which is `@[implicit_reducible]`, so it unfolds even at instance transparency — and on into `componentsToElements`/`componentsFromElements`, `Vector.take`/`drop`/`cast`, and finally the `Array` primitives. It does not come back. A hand-written `ProvableType` instance stops the descent immediately because `size` is a literal; `fromStruct` does not.

The fix splits the two element functions out of the instance and seals them:

```lean
namespace ProvableStruct
def structToElements   {α : TypeMap} [ProvableStruct α] {F : Type} (x : α F) : Vector F (combinedSize α) := …
def structFromElements {α : TypeMap} [ProvableStruct α] {F : Type} (v : Vector F (combinedSize α)) : α F := …

@[circuit_norm] lemma structToElements_eq   … := rfl
@[circuit_norm] lemma structFromElements_eq … := rfl

attribute [irreducible] structToElements structFromElements
end ProvableStruct
```

Two things about the shape:

- **Only the element functions are sealed, not the instance.** `attribute [irreducible] ProvableType.fromStruct` also fixes the derive, but it costs `size` its transparency and breaks ~70 sites (`OfNat (Fin (size RowType)) k`, `.input`/`.aux` dot-notation). Sealing one level down leaves `size = combinedSize α` free to compute, which is what the table and cell machinery needs.
- **The two `rfl` lemmas are what `circuit_norm` reduces through** now, so they have to be proved before the `attribute` line — hence `attribute [irreducible]` after the fact rather than `@[irreducible]` on the definitions.

Ruled out along the way, so nobody repeats them: type ascriptions in the deriving handler; a named `components` definition; passing `provableType` explicitly; `reducible` on `fromStruct`; `irreducible`/semireducible on `combinedSize'`; `irreducible` on `componentsToElements`/`componentsFromElements`; a 2,000,000-heartbeat budget; and consumer-side `backward.proofsInPublic` / `backward.isDefEq.respectTransparency`.

An earlier revision of this PR blamed the deriving handler for emitting `fromComponents` against `components Outer` rather than a literal list. That was wrong — I tested it, and #461 carries the correction.

`Clean/Utils/Test/TestProvableStructSimp.lean` already declares `NestedInputs` nesting `TestInputs`, and it is in `CleanTests`: that is the regression test for #461, so there is no new test file here.

## What the port needed beyond headers

| | |
| --- | --- |
| Metaprogram modules (`public meta section`) | 10, plus a `meta` region in `Clean/Circuit/Basic.lean` and two in `Clean/Circuit/Explicit.lean` |
| `meta import` lines for `native_decide` / `bv_decide` / `#eval` | 48 |
| Core builtin simprocs re-exported as local `meta` simprocs | 12 |
| `import all` for an unexposed core body | **0** — 42 at first, all retired during review |
| `with_unfolding_all` | **7**, down from 12 on `main` |
| Anonymous `‹_›` binders named in exported statements | 10 |
| Proofs adjusted | ~20 files — mostly citing the two new `circuit_norm` lemmas, or routing around an unexposed core body |

Notes on the ones that were not obvious:

- **A simp set only accepts `meta` declarations, and core's `builtin_(d)simproc`s are not marked `meta`.** `attribute [circuit_norm] Fin.reduceCastSucc` is therefore rejected outright. Wrapping the attribute command in `meta section` does not help, nor does `public meta import` of the defining module. The working spelling is to re-export each one as a local `meta` simproc with the same pattern (`Clean/Circuit/Basic.lean`, `Clean/Circuit/Explicit.lean`). Where a proof then listed `reduceIte` explicitly alongside `circuit_norm`, the alias already covers it and the explicit argument became an unused-simp-argument warning, so it is dropped.
- **`native_decide` needs `public meta import` of every module whose compiled code it runs** — including Clean's own (`Clean.Utils.Primes`, `Clean.Utils.Bitwise`, `Clean.Utils.Vector`). The non-meta `public import` has to stay alongside: `meta import` alone hides the module from ordinary code.
- **`import all` turned out to be avoidable, and the reason it looked structural is worth recording.** `circuit_norm`'s struct simprocs validate their rewrites with `isDefEq` under `withTransparency .all`, and `.all` unfolds anything — *except* a definition whose body was never exported. `Array.ofFn` and `Array.map` are not `@[expose]`d by core, so a module that reduces through them silently loses the rewrite and fails several lines later with an unrelated-looking message. The first fix was `import all Init.Data.Array.Basic` per file: it cannot be exported once and for all, because `public import all` is rejected outright and `import all` affects only the importing module's own private scope — it does not propagate to modules that import it. That put 42 of these lines across 36 files. **The better fix, and the one in the branch now, is to remove the need**: proving through lemmas such as `Vector.getElem_ofFn` instead of reducing to the primitive takes `import all` to zero and `with_unfolding_all` from 12 on `main` down to 7. Thanks to @mitschabaude for pushing back on the first version rather than accepting it.
- **`‹_›` in an exported statement leaves a metavariable** the delayed `by` never resolves (`tactic execution is stuck`). Naming the binder fixes it; `Clean/Air/Vm.lean` had ten.
- **`Clean/Gadgets/Boolean.lean` keeps its private constructor.** `Boolean.mk` is private so that only `witness` can build one; exposing `witness`'s body would leak it. `@[no_expose]` on `witness` preserves the encapsulation instead of making the constructor public.

## The ratchet

`lean_lib Clean` now sets `requiresModuleSystem := true`. Verified against a deliberately classic probe file: Lake reports `missing 'module' header as required by the 'requiresModuleSystem' option`, and the `module` root then refuses the import outright. CI's first job runs `--wfail`, so this is a hard gate.

`CleanTests` needs no `allowNonModules` — both its roots are modules. The Lean fixtures under `backends/plonky3/tests/fixtures/` belong to no Lake target and run through `lake env lean --run`, so they are unaffected and stay classic; `cargo test --release test_lean_circuit_end_to_end` passes.

## Verification

```
lake build --wfail        0 errors, 0 warnings
lake build CleanTests     0 errors, 10 `sorry` warnings — the same 10 as main
python3 scripts/check-consecutive-empty-lines.py
cargo test --release -- --nocapture test_lean_circuit_end_to_end   (backends/plonky3)
```

The #461 repro — `Inner` with two field elements, `Outer` nesting two `Inner`s, both `deriving ProvableStruct`, plus `example : size (M := Outer) = 4 := by simp [circuit_norm]` — elaborates in under four seconds.

## Related

The 12 files under `Clean/` that no `lean_lib` root reaches were left alone — they do not compile on `main` either, and the ratchet does not see them. #396 covers one of them (`Circomlib/Poseidon.lean`); the full list is `Circomlib/Poseidon`, `Circuit/ReduceDataExperiment`, `Circuit/WitnessGeneration`, `Examples/PicoCairoMultiplicity`, `Examples/Trace`, `Gadgets/SHA256/And32`, `Specs/Poseidon`, `Specs/PoseidonConstants`, `Specs/PoseidonOptimized`, `Tables/Addition8`, `Types/Byte`, `Utils/Test/SimpMatchRegression`.


dhsorens:
## Progress: `lake build` is green; #461 is fixed, but `CleanTests` is not yet

Pushed `23053567`, kept as its own commit on top of the migration so it can be read separately.

### #461 is fixed, and not the way I guessed

The minimal repro from #461 now elaborates clean, and **`lake build` is green — 0 errors, 0 warnings** (it was 8 errors in `Clean/Gadgets/BLAKE3/FinalizeChunk.lean`).

I said in #461 that the fix probably belonged in `ProvableStructDeriving.lean`, emitting `fromComponents` against the literal component list. **That is wrong** — I tested it and it does not help. Also ruled out, each by direct experiment:

| Attempted fix | Result |
| --- | --- |
| Explicit type ascriptions on the generated `toComponents`/`fromComponents` | no effect |
| Hoisting the component list into a named `def` so the matcher index is a constant | no effect |
| Supplying `provableType` explicitly instead of via the `by infer_instance` autoparam | no effect |
| `attribute [reducible] ProvableType.fromStruct` | no effect |
| `attribute [irreducible] ProvableStruct.combinedSize'`, or dropping its `implicit_reducible` | breaks `Provable.lean` itself |
| `attribute [irreducible] ProvableStruct.componentsToElements componentsFromElements` | no effect |

### What actually fixes it

`ProvableTypeList` is indexed by `List WithProvableType`, and `WithProvableType` carries its `ProvableType` instance **as data**. Elaborating a derived `fromComponents` pattern match therefore has to decide definitional equality of those instances, and when one comes from `ProvableType.fromStruct` the comparison descends into the element functions and does not come back.

Sealing `fromStruct` outright does fix it — but it costs `size` its reducibility, and that broke ~70 sites: `OfNat (Fin (size RowType)) k` no longer synthesizes, and dot-notation like `.input` / `.aux` stops elaborating across the table and cell code. So instead this commit splits the two data fields out of the instance:

```lean
def ProvableType.structToElements   {α} [ProvableStruct α] {F} (x : α F) : Vector F (combinedSize α) := …
def ProvableType.structFromElements {α} [ProvableStruct α] {F} (v : Vector F (combinedSize α)) : α F := …

attribute [irreducible] ProvableType.structToElements ProvableType.structFromElements

instance ProvableType.fromStruct … : ProvableType α where
  size := combinedSize α          -- still computes
  toElements   x := ProvableType.structToElements x
  fromElements v := ProvableType.structFromElements v
```

That stops the descent where a hand-written `ProvableType` instance already stopped it, while `size` stays transparent. Two `@[circuit_norm]` lemmas give simp the unfoldings back; three proofs that relied on the projections reducing now cite them, and the two `and_intros <;> rfl` in `Clean/Tables/Fibonacci32.lean` became `and_intros <;> with_unfolding_all rfl`.

Worth a reviewer's eye: this changes behaviour on `main` too, not just under the module system. It is the one semantic change in the PR.

### Still red: `lake build CleanTests`

15 errors, all in `Clean/Utils/Test/TestProvableStructSimp.lean`. With the element functions sealed, `structEvalLiteralProc`'s `isDefEq` validation stops succeeding, so `ProvableStruct.eval env ⟨a, b, c⟩` no longer decomposes into a struct literal and the `provable_struct_simp` tests fail. Confirmed via `trace.Meta.Tactic.simp.rewrite`:

```
[Meta.Tactic.simp.rewrite] structEvalLiteral: defeq validation failed ProvableStruct.eval env.toEnvironment …
```

It is structural, not a budget problem — it fails in 1.6s at `maxHeartbeats 1000000`. What is odd is that the simproc already validates under `withTransparency .all`, which ought to see through the seal, so either `.all` is not doing what I expect here or the defeq genuinely needs something else. That is the next thread.

### Remaining work

1. The `structEvalLiteralProc` validation above.
2. `FinalizeChunk.lean` can now be migrated — the deriving timeout is gone — but then shows three unrelated proof failures (`split`, `simp_all`, unsolved goals). It is still classic in this push.
3. The other three files left classic: `Clean.lean`, `Clean/Examples/ToJson.lean`, `Clean/Tables/Fibonacci8.lean`.
4. Strip the `import all Init.Data.Array.Basic` workarounds — the seal should make most of them unnecessary — and re-add only what the build still needs.
5. Turn on `requiresModuleSystem` once all 165 files build.

Staying in draft until at least (1) and (2) are done.

dhsorens:
Resolved and out of draft. Since the last update:

**The `CleanTests` blocker was a one-line import.** `structEvalLiteralProc` validates its rewrite with `isDefEq` under `withTransparency .all`, and I had been reading that failure as a transparency problem caused by the new seal. It is not: `canUnfoldDefault` returns `true` unconditionally at `.all`, so `@[irreducible]` is invisible there. What `.all` cannot do is unfold a body that was never exported — and `Array.map`/`Array.ofFn` are not `@[expose]`d by core. `Clean/Utils/Test/TestProvableStructSimp.lean` was simply the one nested-`ProvableStruct` file on the branch without `import all Init.Data.Array.Basic`, because it had never got past the derive to need it. Adding the line cleared all 15 errors.

**The last four files are migrated.** `Clean/Tables/Fibonacci8.lean` and `Clean/Gadgets/BLAKE3/FinalizeChunk.lean` needed the same import and nothing else — the seal had already fixed their real problem; `Clean/Examples/ToJson.lean` and `Clean.lean` were mechanical. All 165 files Lake builds now carry a `module` header.

**The `import all` set is minimised.** All 51 lines came out, and I added back only what each successive build demanded: 42 across 36 files. They are not scaffolding — see the note in the description for why they are structural, and what would retire them.

**The ratchet is on.** `requiresModuleSystem := true` on `lean_lib Clean`, verified by planting a classic probe file and watching Lake reject it.

Green: `lake build --wfail` (0 warnings), `lake build CleanTests` (10 `sorry`s, the same 10 as `main`), the empty-line lint, and `cargo test --release test_lean_circuit_end_to_end` in `backends/plonky3`.

The one thing worth a careful reviewer is `Clean/Circuit/Provable.lean` — sealing `structToElements`/`structFromElements` changes reduction behaviour on `main` too, not just under the module system. Everything else is headers.

Clean/Circuit/Provable.lean
```diff
@@ -381,17 +388,53 @@ theorem toElements_fromElements {F} : (cs : List WithProvableType) → (xs : Vec
...
+/- `ProvableTypeList` is indexed by `List WithProvableType`, and `WithProvableType` carries its
+`ProvableType` instance as *data*. Elaborating a `fromComponents` pattern match therefore has to
+decide definitional equality of those instances, and when one of them comes from `fromStruct`
+that descends into the element functions and does not come back: `deriving ProvableStruct`
+diverges for a struct nesting another struct.
+
+Sealing the two element functions stops the descent there while leaving `size` — which is just
+`combinedSize α` — free to compute, as the table and cell machinery needs. The two lemmas above
+are what `circuit_norm` reduces through, so they must be proved before this line. -/
+attribute [irreducible] ProvableType.structToElements ProvableType.structFromElements
```
mitschabaude: all of this should move into the `ProvableStruct` namespace above IMO

Clean/Gadgets/Rotation32/Theorems.lean
```diff
@@ -32,7 +40,7 @@ def rotRight32_u32 : U32 ℕ → ℕ → U32 ℕ
 
 -- these two are definitionally equal
 lemma rotRight32_bytes_u32_eq (o : ℕ) (x : U32 ℕ) :
-  rotRight32_bytes x.toLimbs o = (rotRight32_u32 x o).toLimbs := rfl
+  rotRight32_bytes x.toLimbs o = (rotRight32_u32 x o).toLimbs := by with_unfolding_all rfl
```
mitschabaude: i'd like to understand why this is needed

Clean/Gadgets/Rotation64/Theorems.lean
```diff
@@ -34,7 +41,7 @@ def rotRight64_u64 : U64 ℕ → ℕ → U64 ℕ
 
 -- these two are definitionally equal
 lemma rotRight64_bytes_u64_eq (o : ℕ) (x : U64 ℕ) :
-  rotRight64_bytes x.toLimbs o = (rotRight64_u64 x o).toLimbs := rfl
+  rotRight64_bytes x.toLimbs o = (rotRight64_u64 x o).toLimbs := by with_unfolding_all rfl
```
mitschabaude: why needed?

Clean/Gadgets/Xor/Xor32.lean
```diff
@@ -1,8 +1,16 @@
...
+
+public import Clean.Circuit
+public import Clean.Utils.Primes
+public import Clean.Utils.Field
+public import Clean.Types.U32
+public import Clean.Gadgets.Xor.ByteXorTable
+
+-- `circuit_norm`'s struct simprocs validate by `isDefEq` through `Array.ofFn`/`mapM`,
+-- whose bodies core does not expose.
+import all Init.Data.Array.Basic
```
mitschabaude: repeating this stuff in gadgets files is noisy, can't we export it once and for all from `Clean.Circuit`?

mitschabaude (commented):
very nice, will dig into a few minor questions but looks broadly ready to merge

Clean/Circuit/Provable.lean
dhsorens: Done in 0f65521 — moved inside `namespace ProvableStruct`, which removes all five `open ProvableStruct in` lines.

The names go with them: `ProvableType.structToElements` → `ProvableStruct.structToElements`, and likewise for `structFromElements` and the two `circuit_norm` lemmas. Only one reference existed outside this file (`WitnessIR.lean:577`), so the rename was cheap; say the word if you'd rather keep the `ProvableType.` prefix via `_root_.` and I'll switch it back.

The ordering constraint the comment block describes is preserved — defs, then the two lemmas, then `attribute [irreducible]`, all before `end ProvableStruct`.

Clean/Gadgets/Rotation32/Theorems.lean
dhsorens: Good question — I'd only half-understood it myself, so I went and pinned it down. Short answer: **`Vector.ofFn`**. I've added a note at both sites in 0f65521 rather than changing the proof, but here's the evidence.

Reverting to plain `rfl` gives:

```
Not a definitional equality: the left-hand side
  rotRight32_bytes x.toLimbs o
is not definitionally equal to the right-hand side
  (rotRight32_u32 x o).toLimbs

Note: This theorem is exported from the current module. This requires that all
definitions that need to be unfolded to prove this theorem must be exposed.
```

So something in the chain is sealed. It is **not** the obvious candidates: this file is under `@[expose] public section`, so `rotRight32_bytes` and `rotRight32_u32` are exposed, and `U32.toLimbs` is exposed too (`Types/U32.lean` is also `@[expose] public section`).

What's left is `rotRight32_bytes`'s `.ofFn`, and that does turn out to be the whole of it. Rewriting the definition as an explicit vector:

```lean
def rotRight32_bytes (xs : Vector ℕ 4) (o : ℕ) : Vector ℕ 4 :=
  #v[xs[0] / 2^o + (xs[1] % 2^o) * 2^(8-o), …]
```

makes plain `rfl` work again — builds clean. So it's core's `Array.ofFn` under `Vector.ofFn`, whose body core doesn't expose. The file's existing header comment already mentions `Array.ofFn`/`mapM` but attributes it to `circuit_norm`'s simprocs, so there was nothing local to the lemma explaining it — which I think is exactly why it read as unmotivated.

**The choice is real, though, and it's yours:** dropping `ofFn` for `#v[...]` removes the `with_unfolding_all` at both sites. I kept `ofFn` because the index-generic form looks like what the downstream rotation proofs want, and one `with_unfolding_all` on a `rfl` lemma seemed the cheaper price. Happy to flip it if you'd rather have no unfolding escape hatches here.

Same cause and same note on `Rotation64/Theorems.lean`.

Clean/Gadgets/Rotation64/Theorems.lean
dhsorens: Same cause as the `Rotation32` one — `rotRight64_bytes` is `.ofFn` over 8 limbs, and core does not expose `Array.ofFn`'s body. Both local defs and `U64.toLimbs` are already exposed, so `ofFn` is the whole of it. Note added at this site too in 0f65521; see the longer reply on the `Rotation32` thread for the evidence and for the `#v[...]` alternative.

Clean/Gadgets/Xor/Xor32.lean
dhsorens: Agreed, and it turns out to be even easier than exporting them: **they were already redundant.** `Clean.Circuit` re-exports `Clean.Utils.Primes` and `Clean.Utils.Field` transitively, so nothing needed adding to `Clean/Circuit.lean` — the lines just needed deleting.

Removed in 0f65521 from every file that also `public import`s `Clean.Circuit`: 14 files, 19 lines. Full build clean at 1858 jobs, no warnings.

I limited it to those two modules and to files that already pull in `Clean.Circuit`, so nothing is relying on an import that only happens to arrive by a longer path. There may well be more of this — `Clean.Types.U32` and `Clean.Gadgets.Xor.ByteXorTable` are still listed here, for instance — but those are not re-exported by `Clean.Circuit`, so trimming them would need a different judgement about what the aggregator ought to carry. Happy to widen it if you want `Clean.Circuit` to be the single entry point for gadget files.

mitschabaude: I was talking about
```lean
-- `circuit_norm`'s struct simprocs validate by `isDefEq` through `Array.ofFn`/`mapM`,
-- whose bodies core does not expose.
import all Init.Data.Array.Basic
```

Clean/Gadgets/Rotation64/Theorems.lean
mitschabaude: I don't want a big comment about how something doesn't work because core doesn't expose Array.ofFn's body. Just make it work instead, change the proof to use a lemma. `rfl` and especially `with_unfolding_all rfl` is an anti-pattern. Remove the comment

dhsorens: Fair — and your `a5eafba` landed while I was doing the same thing, so I've dropped mine in favour of it. Yours is better: `interval_cases i <;> simp [...]` removes the `rfl` as well, where I still had one per index.

For the record, the thing that made this worth doing rather than documenting: once the proof goes through `Vector.getElem_ofFn`, **neither rotation file needs `import all Init.Data.Array.Basic` any more**. So fixing the proof was also the fix for the import noise on the `Xor32` thread. Your commit takes that all the way — `import all` is now at **zero** across `Clean/`, and `with_unfolding_all` is down from 18 to 7.

(One small thing I'd got stuck on and you didn't: `interval_cases` needed `public import Mathlib.Tactic.IntervalCases`, which your commit adds.)

Clean/Gadgets/Xor/Xor32.lean
dhsorens: Sorry — I answered the wrong imports. You meant the `import all Init.Data.Array.Basic` block, which was in 32 files.

I tried to export it from `Clean.Circuit` as you suggested and it is **not possible**, which is worth recording:

```
error: cannot use `all` with `public import`; consider using separate
`public import Init.Data.Array.Basic` and `import all Init.Data.Array.Basic`
directives in order to import public data into the public scope and private
data into the private scope.
```

Taking that suggestion and putting both directives in `Clean/Circuit.lean` does not work either — `Xor32` still fails without its own copy. `import all` only affects the importing module's private scope; it does not propagate to modules that import it. So there is no "once and for all" version of it, and the only way to remove the noise is to remove the *need* for it file by file.

Which is what your `a5eafba` does, and it is now at **zero `import all` across `Clean/`**. I had started down the same path from the rotation lemmas — proving them through `Vector.getElem_ofFn` drops the import from those two files — but yours covers the lot, so I dropped mine.

On my earlier reply here: the `Clean.Utils.Primes` / `Clean.Utils.Field` removals in 0f65521 were me answering a question you had not asked. They are genuinely redundant (`Clean.Circuit` re-exports both transitively) and the tree builds without them, so I have left them in — but say the word and I will revert that part.

mitschabaude: fixed below

dhsorens: sorry for my bot's verbose responses lol

dhsorens:
On the benchmark: net **+0.58%** (5,506k vs 5,475k heartbeats), and the shape of it is explainable.

Six of the seven biggest regressions — `BLAKE3.ApplyRounds` (+11.7%), `BLAKE3.FinalStateUpdate` (+8.3%), `BLAKE3.Round` (+22.0%), `SHA256.SHA256Round` (+11.9%), `FemtoCairo` (+5.3%), `BLAKE3.BLAKE3G` (+7.4%) — have **zero non-header lines** in this diff. No proof in them changed. The only thing that reaches them is the seal in `Provable.lean`: `circuit_norm` now takes a rewrite step through `structToElements_eq`/`structFromElements_eq` where it used to unfold definitionally, and these are the modules that lean hardest on struct evaluation. `Keccak.Chi` (+50%, but only +10k in absolute terms) is the one exception — it lost its `rw [Vector.finRange]` route in a5eafba8.

The same mechanism runs the other way in the improvements column, and there it is worth more: `Tables.Fibonacci8` is **-73.6%** (107k → 28k) on a header-only diff, which I read as the sealed element functions no longer being unfolded during table elaboration. That single module is more than twice the size of the total regression.

So the trade is: a few percent on struct-heavy gadget files, a large win on table elaboration, roughly a wash overall. I don't think it's worth tuning — but if the per-module regressions matter more than the total, the lever is which of the two lemmas carries `@[circuit_norm]`, and I'm happy to look.

mitschabaude (approved):

mitschabaude:
thanks @dhsorens this is great! merging

